PR #101, opened 2020-04-07
Turn examples folder into brevitas_examples package

volcacius:
Models should be exposed with a factory function (1 per cfg) with a flag for downloading the pretrained model. Evaluation scripts should be exposed as entry points in the setup.


brevitas_examples/speech_to_text/quartznet_val.py
```diff
@@ -79,27 +77,12 @@ def main():
...
-    else:
-        # Map model to be loaded to specified single gpu.
         loc = 'cuda:{}'.format(args.gpu)
-        checkpoint_enc = torch.hub.load_state_dict_from_url(pretrained_encoder_url, map_location=loc)
-        checkpoint_dec = torch.hub.load_state_dict_from_url(pretrained_decoder_url, map_location=loc)
-
-    model.restore_checkpoints(checkpoint_enc, checkpoint_dec)
-    model.to(loc)
+    else:
+        loc = 'cpu'
```
volcacius, line 86: Missing `model.to(loc)`

volcacius:
Left to do:

- Move requirements to top level folder
- Import models in the __init__.py of each example subfolder
- Add step in test develop install to test dev installation and import of the models
- Update READMEs

PR #121, opened 2020-04-24
Fix for FP in RestrictValue Class

Giuseppe5:


brevitas/core/function_wrapper.py
```diff
@@ -49,7 +49,7 @@ def __init__(self) -> None:
 
     @torch.jit.script_method
     def forward(self, x: torch.Tensor):
-        return x
+        return identity(x)
```
volcacius, line 52: I don't think this is required, but for sake of structure we can keep it.

Issue #119, opened 2020-04-22
Update CI to run against Pytorch 1.5


volcacius:
Per title.

Giuseppe5:
Support for Pytorch 1.5.0 has been added. MacOS is giving us some issues not directly related to brevitas. 
As of now, for MacOS it is better to use previous pytorch versions.

PR #122, opened 2020-04-24
Add support for pytorch 1.5.0

Giuseppe5:
As per #119 
Initial commit, more to follow.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,8 +51,15 @@ from setuptools.command.install import install
 from setuptools import Extension
 import glob
 import sys
+from distutils.util import convert_path
 
-MIN_TORCH_JITTABLE_VERSION = "1.3.0"
+
+config= {}
+exec(open(convert_path('brevitas/config.py')).read(), config)
+MIN_TORCH_JITTABLE_VERSION = config['MIN_TORCH_JITTABLE_VERSION']
+MAX_TORCH_JITTABLE_VERSION = config['MAX_TORCH_JITTABLE_VERSION']
+
+
 PROJECT_ROOT = os.path.dirname(os.path.abspath(__file__))
 REQUIREMENTS_DIR = os.path.join(PROJECT_ROOT, 'requirements')
 
@@ -60,7 +67,8 @@ REQUIREMENTS_DIR = os.path.join(PROJECT_ROOT, 'requirements')
 def apply_template(dest_build, version):
     template_path = os.path.join('brevitas', 'function', 'ops_ste.py.template')
     generated_path = os.path.join(dest_build, 'ops_ste.py')
-    if version.parse(torch.__version__) >= version.parse(MIN_TORCH_JITTABLE_VERSION):
+    torch_version = version.parse(torch.__version__)
+    if version.parse(MIN_TORCH_JITTABLE_VERSION) <= torch_version <= version.parse(MAX_TORCH_JITTABLE_VERSION):
         d = dict(
             function_suffix='',
             function_prefix='torch.ops.brevitas.',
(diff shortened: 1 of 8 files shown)


volcacius:
Currently conda dry runs for retrieving cache hashes are failing, because the session is not defined in Nox for 1.5.0. You need to add "1.5.0" to PYTORCH_VERSIONS in gen_github_actions.py, that's the single source of truth that Nox adheres too.

volcacius:
Both Windows and Ubuntu are failing in ways I currently don't understand. Can you try disabling extension building with ninja? Add a use_ninja=False flag to the options here https://github.com/Xilinx/brevitas/blob/24d02753a06d151b41e91ea69c6d52769d529178/setup.py#L180

Giuseppe5:
That is one solution, another one proposed in the changelog is the following:
_If a previously succeeding python setup.py install now fails, try setting the MAX_JOBS environment variable._

I will try both and see which one is better (faster, easier to configure, and everything).

Giuseppe5:
I'll leave it here for memo. I need to refactor the torch.autograd functions that are used when the JIT is disabled, since there was a change in the API. This means also a change in the setup.py file for the codegen. Hopefully the new APIs will work also with older version (we were using deprecated APIs, even though no warning was ever printed about that).
If that's not the case, then we'll need support for both version according to pytorch version.

Giuseppe5:
Now everything works as expected, and the errors on MacOS are unrelated to brevitas-support for 1.5

setup.py
```diff
@@ -53,14 +53,16 @@
 import sys
 
 MIN_TORCH_JITTABLE_VERSION = "1.3.0"
+MAX_TORCH_JITTABLE_VERSION = "1.5.0"
```
volcacius, line 56: Move these two constants into brevitas/__init__.py where they are also used, so that we have a single source of truth. __init__.py is not importing anything new compared to setup.py so it should be fine

volcacius: Also, for coherence it should be MAX_TORCH_JITTABLE_VERSION = 1.4.0, since the MIN one is inclusive, and then adjust comparison operators in init and setup accordingly

volcacius:
Also please rebase into two commits, one "Setup: add support for Pytorch 1.5.0" with setup.py and __init__.py, the other "CI/CD: run CI against Pytorch 1.5.0" with everything else

setup.py
Giuseppe5: In  **init.py** there's the import of packaging that breaks everything. Considering the workaround we have in setup.py for having it installed automatically, I would prefer duplicating the two constants for now and look for a cleaner solution in a second moment

volcacius: Just move the import inside the run() methods and then pass the value to the template function.

volcacius: Actually we can just define a bool is_ste_jittable in \_\_init_\_.py and then import it into setup.py from within the run methods.

Giuseppe5: Importing within the run methods causes all the code in __init__ to be executed, included the search for the "_C" file. But at setup-time, the _C module has yet to be created, so it fails.

volcacius: That's a good point, let's bool the bool in brevitas.config and import it into both init and setup. We need to setup a separate pull to move docstrings out of config first though, and into brevitas' init, or we will have another setup time import error.

volcacius:
Please rebase after #136 .

Giuseppe5:
No fix seems to be working for MacOS, pytorch1.5 and python 3.6.
[This (open) issue](https://github.com/pytorch/vision/issues/2163) seems related, even though it happens on ubuntu (in our case, only with MacOS)

PR #145, opened 2020-05-12
Add end2end test on bnn_pynq topologies w/ mnist

Giuseppe5:
Need more commits to setup download from release files

### DIFF
--- a/test/brevitas_examples/common.py
+++ b/test/brevitas_examples/common.py
@@ -1,0 +1,80 @@
+from torchvision.datasets import MNIST
+import pytest
+import os
+import brevitas.nn as qnn
+import torch
+from functools import partial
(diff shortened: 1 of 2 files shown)


test/brevitas_examples/common.py
```diff
@@ -0,0 +1,80 @@
...
+def mnist_datapath_fixture():
+    return mnist_datapath()
+
+def set_and_evaluate_hooks(model, data_loader, dataset ):
+    hooks = []
+    for name, module in model.named_modules():
+        a = register_hook(module, name)
+        if a is not None:
+            hooks.append(a)
+    print("Ciao")
```
Giuseppe5, line 42: Debug line. To be removed

PR #165, opened 2020-06-08
Export QuantTensor

maltanar:
FINN/ONNX export for any layers that use QuantTensor named 3-tuples fail at the moment. Based on our previous discussions I had a first stab at converting all I/O to non-QuantTensors (single tensors) without having to rewrite lots of Brevitas modules.

The core idea is to override the unpack_input and pack_output functions in QuantLayer to do the following:

When export mode is not enabled: cache the incoming/outgoing scale factor and bitwidth values as a member variable in QuantLayer
When export mode is enabled: return cached values for scale factor and bitwidth (if cached, otherwise None) from unpack_input, and always return a single tensor (no QuantTensor 3-tuple) from pack_output
Note that for this scheme to work, all Brevitas modules that want to accept/return QuantTensors must use the pack_input and unpack_output functions, if they manually instantiate a QuantTensor the hook will not catch this and the export will fail. From what I can see, only QuantAvgPool2d actually does this (it calls pack_quant_tensor directly, which is easy to change to pack_output) but I may have missed something, which is also fixed in this PR.

### DIFF
--- a/brevitas/nn/quant_avg_pool.py
+++ b/brevitas/nn/quant_avg_pool.py
@@ -49,6 +49,7 @@ from brevitas.core.quant import QuantType
 from brevitas.function.ops_ste import ceil_ste
 from brevitas.function.ops import max_uint
 from brevitas.nn.quant_layer import QuantLayer
+from brevitas.onnx.onnx_custom_ops import QuantAvgPool2dPlaceholderFunction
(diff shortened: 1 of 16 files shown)


maltanar:
Since this is to be merged into finn_onx_export_dev and we've discussed the details beforehand I'm merging this to enable working on other features for the export.

PR #209, opened 2020-10-01
Setup: move torchvision installation from conda to pip

volcacius:
Signed-off-by: Alessandro Pappalardo <volcacius@users.noreply.github.com>


volcacius:
Not a good idea, closing.

PR #311, opened 2021-05-06
Clamp Half-precision tensors

SpontaneousDuck:
The current implementation of `tensor_clamp` does not work if the input tensor is not the same type as `min_val` and `max_val`. One case where this would happen is if PyTorch's AMP is being used to train at half-precision. With this change, half-precision training is enabled and allowed a 2x training speedup in my use-case. 

This fix was accomplished by casting `min_val` and `max_val` to match the type of input `x` using [`torch.Tensor.type_as()`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.type_as). Per PyTorch's docs, this results in a no-op if the types already match so there should be no performance hit from this change on standard-precision training.

Thanks for all the great work guys!


volcacius:
Thanks a mil for the pull request and the positive feedback.

PR #323, opened 2021-05-21
Fix two typos in README

rgb000000:
There are two typos about torch example in README.md


volcacius:
Thanks for noticing, I merged the fix.

PR #344, opened 2021-06-21
Create const fixed point qunatizer

MichalMachura:
Function that create fixed point quantizer class depends on user parameters like bit width, sign or integer part bits.

### DIFF
--- a/src/brevitas/quant/const_fixed_point.py
+++ b/src/brevitas/quant/const_fixed_point.py
@@ -1,0 +1,132 @@
+import brevitas
+import torch
+from brevitas.inject import ExtendedInjector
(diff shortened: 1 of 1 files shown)


volcacius:
Hi, sorry I never gave feedback to this. We definitely could use const fixed point quantizers to easily generate say tensors in Qm.n format, but I wouldn't go for the style of implementation you are suggesting within Brevitas itself, I think this kind of class-within-function layering is best suited to stay out of tree. Thanks for the contribution though, I appreciate it.

MichalMachura:
Thank you for your answer.
After some... reflection, I agree with you - that's type of implementation is better for custom flow...

PR #347, opened 2021-07-13
Fix resuming in brevitas_bnn_pynq_train by adding strict arg

jmduarte:
Fix resuming from a checkpoint by adding a `strict` arg, which is expected here: https://github.com/Xilinx/brevitas/blob/dev/src/brevitas_examples/bnn_pynq/trainer.py#L156

Otherwise, the current code doesn't work, for example:
```bash
BREVITAS_JIT=1 brevitas_bnn_pynq_train --network CNV_1W1A --experiments /path/to/experiments --epochs 1
BREVITAS_JIT=1 brevitas_bnn_pynq_train --network CNV_1W1A --resume /path/to/checkpoint.tar --evaluate
```


volcacius:
Thanks for spotting it, it's merged in dev now.

PR #408, opened 2022-07-12
Fix (onnx): enable_onnx_checker only for correct torch version

### LINKED ISSUE #402: TRAINED BREVITAS MODEL CAN NOT BE EXPORTED TO ONNX
I have QNN model trained in brevitas. My input shape for a data is (1,3,1024). As I do predictions for my model, I get an accuracy about %80 and on the other hand I am also able to plot a confusion matrix. So; I am sure about my model works. As I want to export this model to onnx in order to use it with FINN I get the error at below.

Please give me any ideas about that.
Sincerely

image

andrei-stoian-zama:
The `enable_onnx_checker` feature was deprecated and removed starting with torch 1.10

This may close #402

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ If you adopt Brevitas in your work, please cite it as:
 
 
 ## History
+
+- *2021/12/13* - Release version 0.7.1, fix a bunch of issues. Added TVMCon 2021 tutorial notebook.
 - *2021/11/03* - Re-release version 0.7.0 (build 1) on PyPI to fix a packaging issue.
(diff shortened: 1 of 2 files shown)


BowenKwan:
I have recently tried to work on the brevitas, it still gives the 

```TypeError: export() got an unexpected keyword argument 'enable_onnx_checker' ```

My pytorch is of version 1.12 and brevitas of version 0.7.1. Is there anything I can do to solve the problem?

volcacius:
It hasn't been released yet but it's merged in the dev branch.

PR #416, opened 2022-08-05
FlexML quantized AvgPool update

volcacius:
Switch over to a Vitis-AI like implementation and export flow for QuantAvgPool2d for FlexML graph quantization. Keep the conversion of AvgPool to dephwise Conv as an option but simplify the flow by having a single scalar weight in the conv layer and by directly doing a AvgPool->QuantConv without passing through float Conv.

@cmcgirr-amd


cmcgirr-amd:
Just an note (don't know if you're aware), I tried using a newer torch version than 1.10.0 and the export fails due to the onnx export API having changed keyword arguments

src/brevitas/nn/target/flexml.py
```diff
@@ -0,0 +1,98 @@
...
+            kernel_size,
+            stride = None,
+            padding = 0,
+            ceil_mode = False,
+            div_quant = Int16QuantAvgPoolDivQuant,
+            return_quant_tensor=True) -> None:
+        nn.AvgPool2d.__init__(
+            self,
+            kernel_size=kernel_size,
+            stride=stride,
```
cmcgirr-amd, line 59: I noticed that the strides are set to `[7, 7]` we need `[1, 1]`. Not sure where this is coming from.

volcacius: The logic for converting an adaptive avg pool to an avg pool assuming the output size divides the input size is (source https://stackoverflow.com/a/63603993/16744139):
stride = input_size // output_size
kernel_size = input_size - (output_size - 1) * stride
padding = 0

cmcgirr-amd: Ah I see, then in this case when kernel=stride=[H, W] would setting stride=[1, 1] be equivalent? I am just asking because I notice Vitis AI Quantizer setting the stride to [1, 1] for this case.

volcacius: I have added a flag with default True to set the stride to 1 whenever the average pool is global (i.e. across all features).

PR #436, opened 2022-10-24
Qcdq-qop-test bug fixes

Giuseppe5:
Follow up on #435 

Now that return_quant_tensor works as intended, few bug fixes were added:
 - Add Clip when narrow_range=True
 - Convert 1D Tensor with one elements to scalar values
 - Compare the output of ONNX Runtime with QCDQ export mode to Floating Point output of Brevitas


volcacius (changes requested):
Can you replace .view(-1) with .flatten() everywhere? Looks cleaner.

Giuseppe5:
@volcacius Changes made, including fixing a typo on a yet untested Module

volcacius (approved):

PR #444, opened 2022-11-07
Fix for failing tests in Dev

Giuseppe5:


volcacius (changes requested):
Avoid removing running GHA against master and dev only, which has been addressed in a separate PR.

volcacius (approved):

PR #445, opened 2022-11-08
Deprecate (export): remove PyXIR and XIR export flow and tests

Giuseppe5:


volcacius:
The imports at `brevitas.onnx` also need to be removed.

volcacius:
The dependencies should also be removed from setup.py and requirements.

volcacius (approved):

PR #448, opened 2022-11-11
Fix for tests in dev

Giuseppe5:


src/brevitas/export/manager.py
```diff
@@ -269,7 +269,10 @@ def jit_inference_trace(cls, module: Module, args: Union[Tensor, QuantTensor, Tu
             module = module.eval()
             module.apply(cls.set_export_handler)
             # do a forward pass with the input to e.g. store input/output shapes
-            cls._cache_inp_out(module, *args)
+            if isinstance(args, Tuple):
```
volcacius, line 272: This doesn't look right, args could be a QuantTensor, which is a NamedTuple, and you don't wanna unpack that

src/brevitas/export/onnx/finn/manager.py
```diff
@@ -121,6 +121,11 @@ def export(
                 or args is not None and isinstance(args, QuantTensor))
                 and bool(input_t) != bool(args)):
             args = args and input_t
+            
+            if args is None:
```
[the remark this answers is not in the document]
Giuseppe5: I don't think it works as you intended. If args is none and input_t is a Tensor, the output of this operations seems to be None

src/brevitas/export/onnx/manager.py
```diff
@@ -71,7 +71,7 @@ def export_onnx(
 
         if onnx is None or opt is None:
             raise ModuleNotFoundError("Installation of onnx and onnxoptimizer is required.")
-        if not (((input_shape is not None) + (input_t is not None) + (args is not None)) == 1):
+        if not (((input_shape is not None) + (input_t is not None) + (args is not None)) >= 1):
```
volcacius, line 74: Keep equality to 1 or things could get weird with passing multiple input definitions

src/brevitas/export/onnx/finn/manager.py
```diff
@@ -121,9 +121,7 @@ def export(
                 or args is not None and isinstance(args, QuantTensor))
                 and bool(input_t) != bool(args)):
             
-            if args is None: # Args needs to be defined, but input_t has to be set to None
-                args = input_t
-                input_t = None
+            args = args or input_t
```
volcacius, line 124: Set input_t to None afterwards

src/brevitas/export/onnx/finn/manager.py
```diff
@@ -121,10 +121,9 @@ def export(
                 or args is not None and isinstance(args, QuantTensor))
                 and bool(input_t) != bool(args)):
             
-            if args is None: # Args needs to be defined, but input_t has to be set to None
-                args = input_t
-                input_t = None
-                
+            args = args or input_t # If either one is QuantTensor, args will be QuantTensor
```
volcacius, line 124: This isn't specific to QuantTensors though isn't it? It could be any input arguments

volcacius (approved):

PR #459, opened 2022-11-23
Fix (flexml): PTQ calibration fix for training/eval mode

Giuseppe5:
Fix for correct handling of training state of layers and proxies during calibration.

BiasCorrection requires more adaptations to match this interface.

```python
self.model = quantize_flexml(self.model)
self.model.eval()
with torch.no_grad():
    with do_calibration(self.model):
        for i, (images, _) in enumerate(self.dataloader_calibration):
            out = self.model(images)
```

### DIFF
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -19,7 +19,8 @@ __all__ = [
     'ClipFloatWeights',
     'DisableEnableQuantization',
     'DisableQuantInference',
-    'BiasCorrection'
+    'BiasCorrection',
+    'calibration_mode'
 ]
 
 _PARAM_PROXIES = (
@@ -43,6 +44,21 @@ _LAYERS_TO_CLIP = (
 def finalize_collect_stats(module):
     if hasattr(module, 'collect_stats_steps') and hasattr(module, 'counter'):
         module.counter = module.collect_stats_steps
+
+
+class calibration_mode(object):
+    def __init__(self, model, enabled=True):
+        self.model = model
+        self.current_state = model.training
+        self.disable_quant_inference = DisableQuantInference()
+        self.enabled=enabled
+
+    def __enter__(self):
+        if self.enabled:
+            self.disable_quant_inference.apply(self.model, is_training=True, quantization_enabled=False)
+
+    def __exit__(self, type, value, traceback):
+        self.disable_quant_inference.apply(self.model, is_training=self.current_state, quantization_enabled=True)
 
 
 class ClipFloatWeights(Transform):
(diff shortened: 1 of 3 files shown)


src/brevitas/graph/calibrate.py
```diff
@@ -43,6 +43,20 @@ def finalize_collect_stats(module):
         module.counter = module.collect_stats_steps
 
 
+class do_calibration(object):
```
volcacius, line 46: Rename to calibration_mode

src/brevitas/graph/calibrate.py
```diff
@@ -43,6 +43,20 @@ def finalize_collect_stats(module):
...
+    def __init__(self, model, enable=True):
```
volcacius, line 47: Rename to enabled

src/brevitas/graph/calibrate.py
```diff
@@ -43,6 +43,20 @@ def finalize_collect_stats(module):
...
+        self.model = model
+        self.current_state = model.training
+        self.disable_quant_inference = DisableQuantInference()
+        self.enable=enable
+
+    def __enter__(self):
```
volcacius, line 53: This way when enabled=False it stills puts the model in training mode. It should be more something like: 
__enter__:
if enabled:
    self.disable_quant_inference.apply(self.model, is_training=True, quantization_enabled=False)

__exit__:
    self.disable_quant_inference.apply(self.model, is_training=current_state, quantization_enabled=True)

src/brevitas/graph/calibrate.py
```diff
@@ -76,42 +90,49 @@ def disable_act_quant_hook(self, module, inp, output):
...
 class DisableQuantInference(DisableEnableQuantization):
 
-    def apply(self, model, *model_args, **model_kwargs):
-        self.disable_act_quantization(model)
-        self.disable_param_quantization(model)
-        output = model(*model_args, **model_kwargs)
-        self.enable_act_quantization(model)
-        self.enable_param_quantization(model)
-        return output
+    def apply(self, model, is_training, enable, *model_args, **model_kwargs):
```
volcacius, line 129: rename enable to quantization_enabled

volcacius (changes requested):

volcacius:
The jupyter tvm tutorial needs to be updated too

tests/brevitas/graph/test_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+    
+    return scale
+
+@given(inp=float_tensor_random_size_st())
+def test_scale_factors_ptq_calibration_po2(inp):
+    class TestModel(nn.Module):
+
+        def __init__(self):
+            super(TestModel, self).__init__()
+            self.act = qnn.QuantIdentity(Int8ActPerTensorFixedPoint)
```
volcacius, line 31: act_quant=Int8ActPerTensorFixedPoint, let's not assume it's the first arg

src/brevitas/graph/calibrate.py
```diff
@@ -45,6 +46,21 @@ def finalize_collect_stats(module):
         module.counter = module.collect_stats_steps
 
 
+class calibration_mode(object):
+    def __init__(self, model, enabled=True):
+        self.model = model
+        self.current_state = model.training
```
volcacius, line 52: this should be called something like previous_training_state

src/brevitas/graph/calibrate.py
```diff
@@ -85,42 +101,49 @@ def disable_act_quant_hook(self, module, inp, output):
...
 
-    def enable_param_quantization(self, model):
+    def enable_param_quantization(self, model, is_training):
         for module in model.modules():
             if isinstance(module, _PARAM_PROXIES):
                 module.disable_quant = False
+                module.train(is_training)
 
 
 class DisableQuantInference(DisableEnableQuantization):
```
volcacius, line 138: We don't need a DisableQuantInference class anymore as inference is performed externally now. The apply method can be moved to DisableEnableQuantization and BiasCorrection can inherit that and then override it.

tests/brevitas/graph/test_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+    assert torch.allclose(expected_scale, scale)
+
+
+def test_calibration_training_state():
+    
+    class TestModel(nn.Module):
+
+        def __init__(self):
+            super(TestModel, self).__init__()
+            self.act = qnn.QuantIdentity(Int8ActPerTensorFixedPoint)
```
volcacius, line 55: act_quant=

tests/brevitas/graph/test_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+        
+    model = TestModel()
+    model.eval()
+    with torch.no_grad():
+        with calibration_mode(model):
+            assert model.act.act_quant.training == True
+            assert model.training == False
+    
+    assert model.act.act_quant.training == False
+    assert model.training == False
```
volcacius, line 68: Final newline

volcacius (changes requested):

src/brevitas/graph/calibrate.py
Giuseppe5: Ok. Maybe in #460 we could move away from the "apply" method and simply use "__call__".

volcacius: let's do it in a separate PR over all graph transformations, it's more work

PR #460, opened 2022-11-24
Update interface for bias correction

Giuseppe5:
~~Tests are missing.~~
~~Update tutorial is missing~~

```python
with bias_correction_mode(self.model):
    for i, (images, _) in enumerate(self.dataloader_calibration):
        print(f"BC iteration {i}")
        model(images) 
```

### DIFF
--- a/notebooks/Brevitas_TVMCon2021.ipynb
+++ b/notebooks/Brevitas_TVMCon2021.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Brevitas requires Python 3.6+ and PyTorch 1.5.0+ and can be installed from PyPI with `pip install brevitas`. This notebook was originally released for Brevitas 0.7.1.\n",
+    "Brevitas requires Python 3.7+ and PyTorch 1.5.0+ and can be installed from PyPI with `pip install brevitas`. This notebook was originally released for Brevitas 0.7.1.\n",
     "It has since been updated to run with Brevitas 0.8 and PyTorch 1.9.0 and is going to be available at `https://github.com/Xilinx/brevitas/tree/master/notebooks`."
    ]
   },
--- a/notebooks/Brevitas_TVMCon2021.ipynb
+++ b/notebooks/Brevitas_TVMCon2021.ipynb
@@ -11,7 +11,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Brevitas requires Python 3.6+ and PyTorch 1.5.0+ and can be installed from PyPI with `pip install brevitas`. This notebook runs with Brevitas 0.7.1 and PyTorch 1.9.0 and is going to be available at `https://github.com/Xilinx/brevitas/tree/master/notebooks`."
+    "Brevitas requires Python 3.6+ and PyTorch 1.5.0+ and can be installed from PyPI with `pip install brevitas`. This notebook was originally released for Brevitas 0.7.1.\n",
+    "It has since been updated to run with Brevitas 0.8 and PyTorch 1.9.0 and is going to be available at `https://github.com/Xilinx/brevitas/tree/master/notebooks`."
    ]
   },
   {
--- a/notebooks/Brevitas_TVMCon2021.ipynb
+++ b/notebooks/Brevitas_TVMCon2021.ipynb
@@ -2240,7 +2240,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from brevitas.graph.calibrate import BiasCorrection\n",
+    "from brevitas.graph.calibrate import bias_correction_mode\n",
     "from brevitas.graph.calibrate import calibration_mode\n",
     "\n",
     "def calibrate_model(calibration_loader, quant_model, args):\n",
@@ -2251,13 +2251,12 @@
     "            for i, (images, _) in enumerate(calibration_loader):\n",
     "                print(f'Calibration iteration {i}')\n",
     "                quant_model(images)\n",
-    "        # Put the model in inference mode to use those statistics\n",
-    "        quant_model.eval()\n",
-    "        bc = BiasCorrection(iterations=len(calibration_loader))\n",
-    "        for i, (images, _) in enumerate(calibration_loader):\n",
-    "            print(f'Bias correction iteration {i}')\n",
-    "            # Apply bias correction\n",
-    "            quant_model = bc.apply(quant_model, images)\n",
+    "        \n",
+    "        # Apply bias correction\n",
+    "        with bias_correction_mode(quant_model):\n",
+    "            for i, (images, _) in enumerate(calibration_loader):\n",
+    "                print(f'Bias correction iteration {i}')\n",
+    "                quant_model(images)\n",
     "    return quant_model"
    ]
   }
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -45,7 +45,7 @@ def finalize_collect_stats(module):
         module.counter = module.collect_stats_steps
 
 
-class calibration_mode(object):
+class calibration_mode:
     def __init__(self, model, enabled=True):
         self.model = model
         self.previous_training_state = model.training
@@ -61,23 +61,20 @@ class calibration_mode(object):
         self.disable_quant_inference.apply(self.model, is_training=self.previous_training_state, quantization_enabled=True)
 
 
-class bias_correction_mode():
+class bias_correction_mode:
     def __init__(self, model, enabled=True):
         self.model = model
-        self.call_pointer = self.model.forward
-        self.current_state = model.training
-        self.bias_correction = BiasCorrection()
-        self.enabled=enabled
-        self.hook = []
+        self.bias_correction = _BiasCorrection()
+        self.enabled = enabled
+        self.hooks = []
 
     def __enter__(self):
         if self.enabled:
-            self.hook.append(self.model.register_forward_pre_hook(self.bias_correction.global_forward_pre_hook))
-            self.hook.append(self.model.register_forward_hook(self.bias_correction.global_forward_hook))
+            self.bias_correction.register_hook_to_wbiol(self.model, self.hooks)
 
     def __exit__(self, type, value, traceback):
         self.bias_correction.apply_correction(self.model)
-        for hook in self.hook:
+        for hook in self.hooks:
             hook.remove()
 
 
@@ -163,14 +160,14 @@ class DisableEnableQuantization(Transform):
             self.enable_param_quantization(model, is_training)
 
 
-class BiasCorrection(DisableEnableQuantization):
+class _BiasCorrection(DisableEnableQuantization):
 
     LAYERS = (QuantWBIOL,)
 
     def __init__(self, layers=LAYERS):
-        super(BiasCorrection, self).__init__()
+        super(_BiasCorrection, self).__init__()
         self.layers = layers
-        self.iterations = 0
+        self.iterations = {}
         self.correction_map = {}
         self.float_mean_map = {}
         self.collect_float_mean_hooks = []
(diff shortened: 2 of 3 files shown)


src/brevitas/utils/torch_utils.py
```diff
@@ -54,4 +55,13 @@ def forward(self, *input):
         out = self.output(modules[0], input)
         for mod in modules[1:]:
             out = self.output(mod, out)
-        return out
\ No newline at end of file
+        return out
+
+
+def torch_partial_deepcopy(model):
+    ### Performs a deepcopy of a torch.nn.Module, except for all the parameters that are instead passed by reference
```
volcacius, line 62: We might as well turn this into a proper docstring

volcacius (changes requested):

src/brevitas/graph/calibrate.py
```diff
@@ -175,41 +218,57 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
+            self.collect_float_mean(model_fp)
+            self.correct_bias(model)
+        else:
+            self.apply_correction(model)
+            self.enable_act_quantization(model_fp, is_training=False)
+            self.enable_param_quantization(model_fp, is_training=False)
+            self.float_hooks_cleanup()
+            self.quant_hooks_cleanup()
+    
+    def call_wrapper(self, *args, model, forward, **kwargs):
```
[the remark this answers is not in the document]
Giuseppe5: No because of how _partial_ works, otherwise when calling the module for the forward pass, it would pass di input to _model_.

src/brevitas/graph/calibrate.py
```diff
@@ -57,9 +57,28 @@ def __enter__(self):
             self.disable_quant_inference.apply(self.model, is_training=True, quantization_enabled=False)
 
     def __exit__(self, type, value, traceback):
+        self.model.apply(finalize_collect_stats)
```
volcacius, line 60: This one and related should be a separate PR, doesn't have anything to do with bias correction. The tutorial should also be updated.

src/brevitas/graph/calibrate.py
```diff
@@ -194,41 +214,44 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
             error = self.float_mean_map[name] - quant_mean
             self.update_correction(name, error)
-            if self.curr_iter == self.iterations - 1:
-                self.apply_correction(name, parent_module)
             del self.float_mean_map[name]
             inp_broadcast_shape = compute_channel_view_shape(inp, channel_dim=self.channel_dim(inp, parent_module))
             return inp + error.reshape(inp_broadcast_shape)
 
-    def collect_float_mean(self, model, *args, **kwargs):
+    def collect_float_mean(self, model):
```
volcacius, line 221: this should be renamed to register_collect_mean_hook

src/brevitas/utils/torch_utils.py
```diff
@@ -54,4 +55,15 @@ def forward(self, *input):
         out = self.output(modules[0], input)
         for mod in modules[1:]:
             out = self.output(mod, out)
-        return out
\ No newline at end of file
+        return out
+
+
+def torch_partial_deepcopy(model):
```
volcacius, line 61: Put this in a separate PR

src/brevitas/graph/calibrate.py
```diff
@@ -57,9 +57,28 @@ def __enter__(self):
...
+    def __init__(self, model, enabled=True):
+        self.model = model
+        self.call_pointer = self.model.forward
+        self.current_state = model.training
+        self.bias_correction = BiasCorrection()
+        self.enabled=enabled
+
+    def __enter__(self):
+        if self.enabled:
+            call_wrapper = partial(self.bias_correction.call_wrapper, model=self.model, wrapped_call=self.model.forward)
```
volcacius, line 74: why wrapped_call is self.model.forward and not self.model.__call__?

notebooks/Brevitas_TVMCon2021.ipynb
```diff
@@ -2252,21 +2252,20 @@
...
+    "                quant_model(images)\n",
     "    quant_model.apply(finalize_collect_stats)\n",
     "    return quant_model"
    ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3.7.15 ('torch_1.10')",
+   "display_name": "Python 3.6.9 64-bit",
```
volcacius, line 2268: The beginning of the tutorial should be updated to say that the tutorial was originally released for brevitas 0.7.1, but has since been updated to brevitas. 0.8.

volcacius (changes requested):

src/brevitas/graph/calibrate.py
Giuseppe5: 1. My understanding is that \_\_call\_\_ is responsible to handle all the hooks. If we would override that, we wouldn't be able to catch all the hooks that we use for collecting FP/Quant values for then applying bias-correction.
2. It seems that torch doesn't even allow a user to change \_\_call\_\_ function, at least from my experiments it seems that just replacing forward with \_\_call\_\_ is not enough to obtain the desired behaviour.

src/brevitas/utils/torch_utils.py
Giuseppe5: Done, will remove from here

src/brevitas/graph/calibrate.py
```diff
@@ -61,6 +61,26 @@ def __exit__(self, type, value, traceback):
         self.disable_quant_inference.apply(self.model, is_training=self.previous_training_state, quantization_enabled=True)
 
 
+class bias_correction_mode():
```
volcacius, line 64: calibration mode inherits from object, this one doesn't. Empty () are also not proper syntax

src/brevitas/graph/calibrate.py
```diff
@@ -61,6 +61,26 @@ def __exit__(self, type, value, traceback):
...
+    def __init__(self, model, enabled=True):
+        self.model = model
+        self.call_pointer = self.model.forward
+        self.current_state = model.training
+        self.bias_correction = BiasCorrection()
+        self.enabled=enabled
+        self.hook = []
```
volcacius, line 71: hooks

src/brevitas/graph/calibrate.py
```diff
@@ -61,6 +61,26 @@ def __exit__(self, type, value, traceback):
         self.disable_quant_inference.apply(self.model, is_training=self.previous_training_state, quantization_enabled=True)
 
 
+class bias_correction_mode():
+    def __init__(self, model, enabled=True):
+        self.model = model
+        self.call_pointer = self.model.forward
```
volcacius, line 67: this seems to be unused now

src/brevitas/graph/calibrate.py
```diff
@@ -61,6 +61,26 @@ def __exit__(self, type, value, traceback):
...
+        self.current_state = model.training
+        self.bias_correction = BiasCorrection()
+        self.enabled=enabled
```
volcacius, line 70: spacing

src/brevitas/graph/calibrate.py
```diff
@@ -195,41 +216,46 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
         self.enable_param_quantization(model, is_training=False)
-        self.correct_bias(model, *model_args, **model_kwargs)
-        self.curr_iter += 1
-        return model
+        self.register_correct_bias_hook(model)
+
+    def global_forward_hook(self, model, inp, output):
+        # Clean-up after forward, to be ready for next input
+        self.correct_bias_hooks_cleanup()
+        self.iterations+=1
```
volcacius, line 261: spacing

src/brevitas/graph/calibrate.py
```diff
@@ -195,41 +216,46 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
         self.correct_bias_hooks = []
-        assert not self.float_mean_map
 
-    def apply(self, model, *model_args, **model_kwargs):
+    def global_forward_pre_hook(self, model, inp):
         self.disable_act_quantization(model, is_training=False)
         self.disable_param_quantization(model, is_training=False)
-        self.collect_float_mean(model, *model_args, **model_kwargs)
+        self.register_collect_float_mean_hook(model)
+        inp, = inp
```
volcacius, line 251: I don't understand this, where is this inp tuple coming from? Why not *model_args, **model_kwargs?

src/brevitas/graph/calibrate.py
```diff
@@ -195,41 +216,46 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
+        model.forward(inp) # Required to avoid infinite recursion
```
volcacius, line 252: this way we are skipping any potential hooks that the user might have defined on the model. Can we instead wrap the model in a module and register the hook on that, and call the forward of the wrapper?

src/brevitas/graph/calibrate.py
```diff
@@ -195,41 +216,46 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
+        self.collect_float_mean_hooks = []
+
+    def correct_bias_hooks_cleanup(self):
         for hook in self.correct_bias_hooks:
             hook.remove()
         self.correct_bias_hooks = []
-        assert not self.float_mean_map
 
-    def apply(self, model, *model_args, **model_kwargs):
+    def global_forward_pre_hook(self, model, inp):
```
volcacius, line 247: It's not obvious to me how this works anymore. Is the implementation tied to bias_correction_mode? We should add a few more comments and potentially rename the class to _BiasCorrection if it's only for internal use

volcacius (changes requested):

src/brevitas/graph/calibrate.py
Giuseppe5: We don't skip any hook because outside the user is still calling `model(inp)`, that executes forward_pre_hooks and forward_hooks.

We are just using a forward_pre_hook to execute a hidden forward without quantization. 
Then we re-enable quantization, the execution continues with the "original forward", then the forward_hooks would be called (and we have one to remove our hooks for the next iteration).

If there are other hooks, they would be called just as ours are. In this way, we are also guaranteeing that if the user set a forward_hook, they would see the original quantized model output.

Giuseppe5: I will investigate more to be sure that everything works as expected, maybe add some tests if you have particular scenarios that we need to preserve.

src/brevitas/graph/calibrate.py
Giuseppe5: Model and Inp are the arguments passed to a forward_pre_hook. Only *args are supported and **kwargs are not passed in forward_pre_hooks

Giuseppe5:
New iteration.
I tried not to disrupt anything while doing bias correction.
In bias_correction_mode, the model structure remains the same (which would not happen if we would wrap the model in another module).

Similarly, if we were to use a wrapper and register a single hook there, when the user would register hooks on quantized layers (intended as layers defined in brevitas.nn), these forwards would be called multiple times and the hooks would behave very differently in bias_correction_mode comapred to "normal" execution.

In this way, hooks behave almost the same way in bias_correction_mode and in normal exeuction, with two small differences:
- The input of QuantLayers (WBIOL) is the not the quantized output of the previous layers, but its FP variant.
- A forward-hook on a QuantLayer won't be able to modify its output (since we already do it with our own hook).

src/brevitas/graph/calibrate.py
```diff
@@ -195,41 +213,61 @@ def correct_bias_hook(self, module, inp, name, parent_module):
...
-        return model
+
+    def register_hook_to_wbiol(self, model, hooks):
+        """
+        Forward hooks are registered to the WBIOL layers.
+        In this way, if more hooks are registered to these layers, they will be only called once per module call.
+        We perform two more forwards for each WBIOL, but since we call forward and not __call__, eventual hooks would be skipped.
+        This is a desired behaviour, since these extra forwards are only necessary to compute the bias correction factor.
+
+        If we register a single hook for the entire model, and hooks are added to WBIOL layers,
```
volcacius, line 247: if we registered ... and hooks were added ... which would be an unexpected

Issue #450, opened 2022-11-11
Unexpected output shape value for BatchNorm1dToQuantScaleBias


green-cabbage:
Hey, I have been trying to implement brevitas' version of nn.BatchNorm1d(), and I receive this error:
`ValueError: Using a target size (torch.Size([64, 1])) that is different to the input size (torch.Size([1, 30, 64, 1])) is deprecated. Please ensure they have the same size.`

I have backtracked the error to the point where I believe this is the cause. When I run this:

```
input_dim = 30
batch_dim = 64
x = torch.rand(64,30)
quant_batchnorm = qnn.BatchNorm1dToQuantScaleBias(
                  num_features=input_dim,
                  return_quant_tensor=True
)  
print(quant_batchnorm(x).shape)
```

the code prints: `torch.Size([1, 30, 64, 30])` when I would expect it to return something like `torch.Size([64,30]).` Do you guys know why the quant layer returns in that specific shape of `[1, input_dim, batch_dim, input_dim]` ? Is this behavior intentional? If so why? 
Thanks in advance

green-cabbage:
Issue solved:

There was two small changes I made which made it work. 

First, the default parameter `runtime_shape=(1, -1, 1)` for `class BatchNorm1dToQuantScaleBias` should actually be `runtime_shape=(1, -1)` . At least that's the shape that worked for me, and makes sense in the context of `torch.nn.BatchNorm1d`

Second, I was initially getting `runtime_shape=(1, -1, 1, 1)` because that is `class ScaleBias`'s default value, and `class BatchNorm1dToQuantScaleBias` wasn't passing down `runtime_shape`parameter properly.
I solved this by changing from `ScaleBias.__init__(self, num_features, bias)` to `ScaleBias.__init__(self, num_features, bias, runtime_shape=runtime_shape)` when initializing `class QuantScaleBias`

If you would like, I can make a pull request with these changes.

MikeHatchi:
Hi @green-cabbage many thanks, I've applied the solution.
I have not noticed normalization actually. In my case, for eg.: among the input data, I had 1.32.... and after BN1dScaleBias 1.39 au lieu de instead of 0.something.

Hi @Giuseppe5 , I was wondering why BN1dScaleBias operation is a missing topic as BN1d is a very largely used in non-quantize context? Is it because as quantization is about low precision, then it is not supposed to use BN logic?

gfwangjg:
> Issue solved:
> 
[...]

PR #465, opened 2022-11-28
fixed wrong runtime shape inference for BatchNorm1dToQuantScaleBias

green-cabbage:
Fixed issue where the output shape of `class BatchNorm1dToQuantScaleBias` was giving an unexpected shape of `[1, input_dim, batch_dim, input_dim]` instead of `[batch_dim, input_dim]`. 

The issue came from the fact that the class ultimately had a default value of `runtime_shape = (1, -1, 1, 1)`  when it should be `runtime_shape = (1, -1)` .

We had `runtime_shape = (1, -1, 1, 1)`  because `class BatchNorm1dToQuantScaleBias` wasn't properly passing down the `runtime_shape` parameter to `class ScaleBias` . That issue has been also fixed.

Original issue from https://github.com/Xilinx/brevitas/issues/450

### DIFF
--- a/src/brevitas/nn/quant_bn.py
+++ b/src/brevitas/nn/quant_bn.py
@@ -97,7 +97,7 @@ class BatchNorm1dToQuantScaleBias(_BatchNormToQuantScaleBias):
         super(BatchNorm1dToQuantScaleBias, self).__init__(
             num_features,
             bias=True,
-            runtime_shape=(1, -1),
+            runtime_shape=(1, -1, 1),
             weight_quant=weight_quant,
             bias_quant=bias_quant,
             input_quant=input_quant,
--- a/src/brevitas/nn/quant_bn.py
+++ b/src/brevitas/nn/quant_bn.py
@@ -97,7 +97,7 @@ class BatchNorm1dToQuantScaleBias(_BatchNormToQuantScaleBias):
         super(BatchNorm1dToQuantScaleBias, self).__init__(
             num_features,
             bias=True,
-            runtime_shape=(1, -1, 1),
+            runtime_shape=(1, -1),
             weight_quant=weight_quant,
             bias_quant=bias_quant,
             input_quant=input_quant,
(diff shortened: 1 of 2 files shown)


src/brevitas/nn/quant_bn.py
```diff
@@ -97,7 +97,7 @@ def __init__(
         super(BatchNorm1dToQuantScaleBias, self).__init__(
             num_features,
             bias=True,
-            runtime_shape=(1, -1, 1),
+            runtime_shape=(1, -1),
```
Giuseppe5, line 100: I think this has to stay as it was before, i.e. (1, -1, 1).
The reason is that BatchNorm1d is supposed to handle both 2D and 3D inputs. In case you need specifically to use a 2D input, you could always specify the runtime shape when instantiating the class.

Giuseppe5 (changes requested):
Hi,
Thanks for following up on your issue and opening this PR.
We are in the process of better defining how to contribute to Brevitas, and recently we added some guidelines regarding [how to add new tests](https://github.com/Xilinx/brevitas/blob/dev/tests/README.md).

Would you be able to add some tests regarding this PR?
For example, checking the output shapes for different types of input (either 2D or 3D).

Let me know if there is something not clear.

Giuseppe

src/brevitas/nn/quant_bn.py
green-cabbage: Sure thing. I haven't really made any serious test cases before, so this will take some time (since I would have to read all the prerequisites), but I could definitely make some drafts.

green-cabbage: I have also made the changes that you requested in my dev branch, but how do I update the pull request? Do I start another pull request? Thank you

Giuseppe5: Every change, including commits, that you do in your dev branch will be automatically added to this PR, so when you have the tests, just add them to the same dev branch

PR #466, opened 2022-11-29
Tests: test against newer pytorch versions up to 1.13

Giuseppe5:


volcacius (changes requested):
Let's test only against the latest minor release for a particular version, e.g. 1.10.1 but not 1.10.0 .

PR #469, opened 2022-12-02
fix: update pytorch syntax for two wrappers

andrei-stoian-zama:
Recent torch version use `destination` as the kwarg name, instead of `dest`. Using `dest` is deprecated and creates a warning

Note that there are two commits:
- the oldest one is for wrappers that I tested (since they are called from Concrete-ML)
- the newer one is for wrappers for which I changed the syntax but could not test 

Do you have a way to test the second the "newest" commit changes ?

### DIFF
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -9,9 +9,10 @@ from brevitas.proxy.runtime_quant import ActQuantProxyFromInjector
 from brevitas.proxy.runtime_quant import TruncQuantProxyFromInjector
 from brevitas.proxy.runtime_quant import ClampQuantProxyFromInjector
 from brevitas.nn.quant_layer import QuantWeightBiasInputOutputLayer as QuantWBIOL
-from brevitas.nn import QuantLinear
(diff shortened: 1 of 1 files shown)


volcacius:
Hi Andrei,

Thanks for the fix. We should add some tests to retrieve the state_dict of each component to check that it doesn't trigger issues with older versions. Right now we still lack some general infrastructure to easily allocate classes of components through pytest fixtures, but let's not try to solve that now and go for something basic. If you could add a test that allocates the respective modules, retrieve the state_dict and verifies that the expected values are there for, that would be great:
- Hadamard under tests/brevitas/nn/test_hadamard.py (not there, needs to be created)
- RNN and LSTM under tests/brevitas/nn/test_recurrent.py
-  View wrapper under tests/brevitas/core/test_stats_view_wrapper.py (needs to be created). For arguments like view_shape_impl we can just put an Identity module since we want to only test the behaviour of the wrapper itself.
- Standalone scaling under tests/brevitas/core/test_standalone_scaling.py similarly for the stats_impl let's have something simple that simply does a max, as long as the returned shape is coherent with the shape of the buffer it's okay.

PR #471, opened 2022-12-02
Fix/pytorch syntax in wrappers

andrei-stoian-zama:
New pytorch versions need `destination` kwarg or they issue a warning

Please note that I could only test the changes in the `ba8c12d` commit, so the other changes in `2331ed3` are not tested

### DIFF
--- a/tests/brevitas/core/test_standalone_view_wrapper.py
+++ b/tests/brevitas/core/test_standalone_view_wrapper.py
@@ -1,0 +1,35 @@
+import pytest
+import warnings
+import torch
+import torch.nn as nn
+
+from brevitas.core.stats.view_wrapper import _ViewParameterWrapper, _ViewCatParameterWrapper
+from brevitas.core.stats.stats_op import AbsMax
+
+
+def test_scaling_state_dict():
+    class TestModuleStd(nn.Module):
+        def __init__(self) -> None:
+            super().__init__()
+            self.conv = nn.Conv2d(1, 1, 3)
+            self.scaling_op = _ViewParameterWrapper(self.conv.weight, nn.Identity())
+
+    mod = TestModuleStd()
+
+    with warnings.catch_warnings(record=True) as wlist:
+        mod.state_dict()
+        for w in wlist:
+            assert "Positional args are being deprecated" not in str(w.message)
+
+    class TestModuleCat(nn.Module):
+        def __init__(self) -> None:
+            super().__init__()
+            self.conv = nn.Conv2d(1, 1, 3)
+            self.scaling_op = _ViewCatParameterWrapper(self.conv.weight, nn.Identity(), 1)
+
+    mod = TestModuleCat()
+
+    with warnings.catch_warnings(record=True) as wlist:
+        mod.state_dict()
+        for w in wlist:
+            assert "Positional args are being deprecated" not in str(w.message)
--- a/tests/brevitas/nn/test_recurrent.py
+++ b/tests/brevitas/nn/test_recurrent.py
@@ -40,6 +40,7 @@
 
 from hypothesis import given
 import pytest
+import warnings
 
 import torch
 from brevitas.nn import QuantRNN, QuantLSTM
@@ -87,6 +88,11 @@ class TestRecurrent:
         assert torch.isclose(out[0], ref_out[0], atol=ATOL).all()
         assert torch.isclose(out[1], ref_out[1], atol=ATOL).all()
 
+        # Test that the brevitas model can be saved/loaded without warning
+        with warnings.catch_warnings(record=True) as wlist:
+            qm.load_state_dict(qm.state_dict())
+            for w in wlist:
+                assert "Positional args are being deprecated" not in str(w.message)
 
     @given(
          inp=float_tensor_random_size_st(dims=3, max_size=3),
@@ -129,6 +135,14 @@ class TestRecurrent:
         assert torch.isclose(out[1][0], ref_out[1][0], atol=ATOL).all()
         # cell states
         assert torch.isclose(out[1][1], ref_out[1][1], atol=ATOL).all()
+
+        # Test that the brevitas model can be saved/loaded without warning
+        # Test that the brevitas model can be saved/loaded without warning
+        with warnings.catch_warnings(record=True) as wlist:
+            qm.load_state_dict(qm.state_dict())
+            for w in wlist:
+                assert "Positional args are being deprecated" not in str(w.message)
+
 
     @pytest.mark.parametrize("batch_first", [True, False])
     @pytest.mark.parametrize("bidirectional", [True, False, 'shared_input_hidden_weights'])
--- a/tests/brevitas/nn/test_recurrent.py
+++ b/tests/brevitas/nn/test_recurrent.py
@@ -72,7 +72,7 @@ class TestRecurrent:
             num_layers=num_layers,
             bias=bias)
         qm = QuantRNN(
-            inp_size, 
+            inp_size,
             hidden_size,
             weight_quant=None,
             bias_quant=None,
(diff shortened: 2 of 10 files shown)


andrei-stoian-zama:
Reposting @volcacius 's message :+1: 

Hi Andrei,

Thanks for the fix. We should add some tests to retrieve the state_dict of each component to check that it doesn't trigger issues with older versions. Right now we still lack some general infrastructure to easily allocate classes of components through pytest fixtures, but let's not try to solve that now and go for something basic. If you could add a test that allocates the respective modules, retrieve the state_dict and verifies that the expected values are there for, that would be great:

- Hadamard under tests/brevitas/nn/test_hadamard.py (not there, needs to be created)
- RNN and LSTM under tests/brevitas/nn/test_recurrent.py
- View wrapper under tests/brevitas/core/test_stats_view_wrapper.py (needs to be created). For arguments like view_shape_impl we can just put an Identity module since we want to only test the behaviour of the wrapper itself.
- Standalone scaling under tests/brevitas/core/test_standalone_scaling.py similarly for the stats_impl let's have something simple that simply does a max, as long as the returned shape is coherent with the shape of the buffer it's okay.

Giuseppe5:
Hi Andrei,
I was wondering if you had time to follow up on @volcacius comment and add some tests for this fix?

Please let us know if there is anything we could help with!

andrei-stoian-zama:
> Hi Andrei, I was wondering if you had time to follow up on @volcacius comment and add some tests for this fix?
> 
> Please let us know if there is anything we could help with!

I'm working on it, should have them done by tomorrow

andrei-stoian-zama:
> Hi Andrei, I was wondering if you had time to follow up on @volcacius comment and add some tests for this fix?
> 
> Please let us know if there is anything we could help with!

I added the tests as @volcacius described

tests/brevitas/nn/test_recurrent.py
```diff
@@ -87,6 +88,11 @@ def test_rnn_quant_disabled_fwd(
         assert torch.isclose(out[0], ref_out[0], atol=ATOL).all()
         assert torch.isclose(out[1], ref_out[1], atol=ATOL).all()
 
+        # Test that the brevitas model can be saved/loaded without warning
+        with warnings.catch_warnings(record=True) as wlist:
```
Giuseppe5, line 92: Generally speaking, I prefer having a separate function for each aspect we want to test, even at the cost of duplicating code (which could be prevented with fixtures and parametrization).

Keeping things simple, for this assert my understanding is that we don't need to parametrize the different variations of RNN/LSTM (but please correct me if I'm wrong). Definitely we don't need to run it for every input tensor generated by hypothesis.

I would create two separate tests with a single version of RNN/LSTM, and test this assert there.

In a separate PR we can figure out a way to leverage pytest_cases to test different implementations of RNN/LSTMs for all tests here.

Another option would be to leave it like this for the moment, and address everything in a separate PR. What's your opinion @volcacius?

tests/brevitas/nn/test_recurrent.py
```diff
@@ -130,6 +136,14 @@ def test_lstm_quant_disabled_fwd(
         # cell states
         assert torch.isclose(out[1][1], ref_out[1][1], atol=ATOL).all()
 
+        # Test that the brevitas model can be saved/loaded without warning
```
Giuseppe5, line 139: Duplicated comment

tests/brevitas/core/test_standalone_view_wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+            super().__init__()
+            self.conv = nn.Conv2d(1, 1, 3)
+            self.scaling_op = _ViewCatParameterWrapper(self.conv.weight, nn.Identity(), 1)
+
+    mod = TestModuleCat()
+
+    with warnings.catch_warnings(record=True) as wlist:
+        mod.state_dict()
+        for w in wlist:
+            assert "Positional args are being deprecated" not in str(w.message)
```
Giuseppe5, line 35: As in the other case, I would split this into two tests

andrei-stoian-zama: please see my comment above.

tests/brevitas/nn/test_recurrent.py
andrei-stoian-zama: removed

tests/brevitas/core/test_standalone_view_wrapper.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+import warnings
+import torch
+import torch.nn as nn
+
+from brevitas.core.stats.view_wrapper import _ViewParameterWrapper, _ViewCatParameterWrapper
+from brevitas.core.stats.stats_op import AbsMax
+
+
+def test_scaling_state_dict():
```
Giuseppe5, line 10: Could you please create two tests:
- One called test_scaling_state_dict_viewparameterwrapper with the class TestModuleStd and the respective assert
- One called test_scaling_state_dict_viewcatparameterwrapper with the class TestModuleCat and the respective assert

tests/brevitas/nn/test_recurrent.py
```diff
@@ -87,6 +88,11 @@ def test_rnn_quant_disabled_fwd(
         assert torch.isclose(out[0], ref_out[0], atol=ATOL).all()
         assert torch.isclose(out[1], ref_out[1], atol=ATOL).all()
 
+        # Test that the brevitas model can be saved/loaded without warning
```
Giuseppe5, line 91: Could you please move this assert to a new test called test_quantrnn_state_dict where you instantiate a single QuantRNN with constant inp_size/hidden_size/etc. where you check only this assert?

tests/brevitas/nn/test_recurrent.py
```diff
@@ -130,6 +136,13 @@ def test_lstm_quant_disabled_fwd(
         # cell states
         assert torch.isclose(out[1][1], ref_out[1][1], atol=ATOL).all()
 
+        # Test that the brevitas model can be saved/loaded without warning
```
Giuseppe5, line 139: Could you please move this assert to a new test called test_quantlstm_state_dict where you instantiate a single QuantLSTM with constant inp_size/hidden_size/etc. where you check only this assert?

Giuseppe5 (changes requested):

tests/brevitas/nn/test_recurrent.py
andrei-stoian-zama: yes, done!

tests/brevitas/nn/test_recurrent.py
```diff
@@ -194,3 +282,46 @@ def test_quant_lstm_fwd_call(
...
+    @pytest.mark.parametrize("batch_first", [True, False])
+    @pytest.mark.parametrize("bidirectional", [True, False, 'shared_input_hidden_weights'])
+    @pytest.mark.parametrize("num_layers", [1, 2])
+    @pytest.mark.parametrize("bias", [True, False])
+    @pytest.mark.parametrize("coupled_input_forget_gates", [True, False])
+    @pytest.mark.parametrize("return_quant_tensor", [True, False])
+    @pytest.mark.parametrize("shared_intra_layer_weight_quant", [True, False])
+    @pytest.mark.parametrize("shared_intra_layer_gate_acc_quant", [True, False])
+    @pytest.mark.parametrize("shared_cell_state_quant", [True, False])
+    @pytest.mark.skip("FIXME: many warnings due to __torch_function__")
```
andrei-stoian-zama, line 295: note I'm disabling this test for now since there are many __torch__function warnings that need fixing

tests/brevitas/nn/test_recurrent.py
```diff
@@ -156,6 +215,35 @@ def test_quant_rnn_fwd_call(
             return_quant_tensor=return_quant_tensor)
         assert m(inp) is not None
 
+    @pytest.mark.parametrize("batch_first", [True, False])
+    @pytest.mark.parametrize("bidirectional", [True, False, 'shared_input_hidden_weights'])
+    @pytest.mark.parametrize("num_layers", [1, 2])
+    @pytest.mark.parametrize("bias", [True, False])
+    @pytest.mark.parametrize("return_quant_tensor", [True, False])
+    @pytest.mark.skip("FIXME: many warnings due to __torch_function__")
```
andrei-stoian-zama, line 223: note I'm disabling this test for now since there are many __torch__function warnings that need fixing

andrei-stoian-zama:
@Giuseppe5 : I split the tests to have all warnings tested in separate tests. I added the warnings tests for the non-quantdisabled tests too, but had to disable these tests since there are many warnings in __torch_function in my version of brevitas. If you think these __torch__function warnings are already fixed (I did not rebase on latest `dev`) let me know

tests/brevitas/nn/test_recurrent.py
Giuseppe5: I would let this and the other test run anyway.
The warnings during tests are a good reminder to go back to fix them.

volcacius:
Looks good to me, we can refactor anything appropriate in a separate PR after the release.

PR #474, opened 2022-12-06
fix: better initialize quant tensor, avoir torch warning

### LINKED ISSUE #456: QuantTensor is never valid / is never is_not_none on master
The constructor for QuantTensor is broken on master, especially the "clone_value" functionality. Could a merge from `dev` which may work (checking the code it seems `clone_value` is fixed) be envisaged ?

On master:

```
class QuantTensor(QuantTensorBase):

    def __new__(
            cls, value, scale=None, zero_point=None, bit_width=None, signed=None, training=None):

        def clone_value(x, torch_dtype):
            if isinstance(x, torch.Tensor):
                x = x.clone().detach() 
            else:
                x = torch.tensor(x, dtype=torch_dtype)
```

`clone_value` does not return anything

andrei-stoian-zama:
Pytorch issues a warning if you create a `tensor.Tensor` based on another tensor. They propose this solution with `.clone().detach()`

Closes #456

### DIFF
--- a/src/brevitas/quant_tensor/__init__.py
+++ b/src/brevitas/quant_tensor/__init__.py
@@ -86,15 +86,16 @@ class QuantTensor(QuantTensorBase):
 
     def __new__(
             cls, value, scale=None, zero_point=None, bit_width=None, signed=None, training=None):
+
         if scale is not None and not isinstance(scale, torch.Tensor):
             scale = torch.tensor(scale, dtype=torch.float)
         if zero_point is not None and not isinstance(zero_point, torch.Tensor):
             zero_point = torch.tensor(zero_point, dtype=torch.float)
         if bit_width is not None and not isinstance(bit_width, torch.Tensor):
             bit_width = torch.tensor(bit_width, dtype=torch.float)
-        if signed is not None:
+        if signed is not None and not isinstance(signed, torch.Tensor):
             signed = torch.tensor(signed, dtype=torch.bool)
-        if training is not None:
+        if training is not None and not isinstance(training, torch.Tensor):
             training = torch.tensor(training, dtype=torch.bool)
         return super().__new__(cls, value, scale, zero_point, bit_width, signed, training)
 


Giuseppe5:
Hi Andrei,
Thanks for looking into this and opening this PR.
Could you please provide an example of when this warning occurs?

I am not sure we want to go for clone().detach() every time the value passed to the constructor is a tensor (e.g., we may want to preserve the gradient information for the scale factor).

andrei-stoian-zama:
> Hi Andrei, Thanks for looking into this and opening this PR. Could you please provide an example of when this warning occurs?
> 
> I am not sure we want to go for clone().detach() every time the value passed to the constructor is a tensor (e.g., we may want to preserve the gradient information for the scale factor).

Ok, I see maybe we should use `tensor.clone()` instead ? It seems to preserve gradient flow.

Here's the warning with torch 1.12:

```
>>> import torch
>>> t = torch.tensor([0, 1, 2])
>>> t
tensor([0, 1, 2])
>>> t2 = torch.tensor(t)
<stdin>:1: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
```

Giuseppe5:
I was double checking this, and it seems that the default behavior for:
```python
t = torch.tensor([0, 1, 2])
t2 = torch.tensor(t)
```
is 

```python
t = torch.tensor([0, 1, 2])
t2 = t.clone().detach()
```

Which means that currently we are not propagating the grad information anyway.

In which use case did you see this warning exactly? I am trying to understand what the expected behavior in that case would be!

andrei-stoian-zama:
> I was double checking this, and it seems that the default behavior for:
> 
> ```python
> t = torch.tensor([0, 1, 2])
> t2 = torch.tensor(t)
> ```
> 
> is
> 
> ```python
> t = torch.tensor([0, 1, 2])
> t2 = t.clone().detach()
> ```
> 
> Which means that currently we are not propagating the grad information anyway.
> 
> In which use case did you see this warning exactly? I am trying to understand what the expected behavior in that case would be!

I think I see the behavior when calling `q.int()` on a `QuantTensor` or `q.quant_weight()` on a quantized Layer

andrei-stoian-zama:
```
>> from brevitas.quant_tensor import QuantTensor
>>> import torch
>>> q = QuantTensor(torch.tensor([0.0]), scale=torch.tensor([1.0]), zero_point=0, bit_width=8, signed=torch.tensor([1]), training=False)
/home/stoiana/Documents/Work/brevitas/src/brevitas/quant_tensor/__init__.py:96: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
```

The warning is issued when "signed" is a tensor. This occurs quite often when using the quantized layers.

src/brevitas/quant_tensor/__init__.py
```diff
@@ -86,16 +86,22 @@ class QuantTensor(QuantTensorBase):
...
         if scale is not None and not isinstance(scale, torch.Tensor):
-            scale = torch.tensor(scale, dtype=torch.float)
+            scale = copy_value(scale, dtype=torch.float)
         if zero_point is not None and not isinstance(zero_point, torch.Tensor):
-            zero_point = torch.tensor(zero_point, dtype=torch.float)
+            zero_point = copy_value(zero_point, dtype=torch.float)
         if bit_width is not None and not isinstance(bit_width, torch.Tensor):
-            bit_width = torch.tensor(bit_width, dtype=torch.float)
+            bit_width = copy_value(bit_width, dtype=torch.float)
         if signed is not None:
```
Giuseppe5, line 101: I guess the simplest solution is to add the "and not isinstance(bit_width, torch.Tensor)" condition for both signed and training (line 95 and 97 in the original file), without the necessity to define and use the `copy_value` function.

andrei-stoian-zama: ok, it's done !

Giuseppe5 (approved):

PR #476, opened 2022-12-08
Graph quantization tests

Giuseppe5:


Giuseppe5:
There are still some OOM crashes, mostly due to VGG11 on windows, during torchscript export.
Maybe we can figure a way to make it less memory intensive, since ONNX export does not show any problems @volcacius.

tests/brevitas_end_to_end/test_torchvision_models.py
```diff
@@ -0,0 +1,71 @@
...
+
+@requires_pt_ge('1.8.1')
+def test_torchvision_graph_quantization_qcdq_onnx(torchvision_model):
+    if torchvision_model is None:
+        pytest.skip('Model not instantiated')
+
+    inp =  torch.randn(BATCH, IN_CH, HEIGHT, WIDTH)
+    export_standard_qcdq_onnx(
+        torchvision_model, args=inp, export_path='model_onnx_qcdq.onnx')
+    # os.remove('model_onnx_qcdq.onnx')
```
volcacius, line 61: Should this be uncommented?

tests/brevitas_end_to_end/test_torchvision_models.py
```diff
@@ -0,0 +1,71 @@
...
+        model = model_fn(pretrained=False)
+
+    model.eval()
+    model = preprocess_flexml(model, inp)
+    model = quantize_flexml(model)
+    return model
+
+
+@requires_pt_ge('1.8.1')
+def test_torchvision_graph_quantization_qcdq_onnx(torchvision_model):
```
volcacius, line 54: this is specific to flexml, should be named as such. Same as for the torch one.

src/brevitas/quant_tensor/__init__.py
```diff
@@ -298,7 +298,7 @@ def cat(tensors, dim, out=None):
             if all([isinstance(qt, QuantTensor) and qt.is_not_none for qt in tensors]):
                 for qt in tensors[1:]:
                     first_qt.check_scaling_factors_same(qt)
-                    first_qt.check_zero_points_same(qt)
+                    first_qt.check_zero_points_same(qt) # Torch.allclose fails with different data types
```
volcacius, line 301: Not clear to me what this comment is trying to say

src/brevitas/quant_tensor/__init__.py
```diff
@@ -310,7 +310,7 @@ def cat(tensors, dim, out=None):
                 return QuantTensor(
                     value=output_value,
                     scale=output_scale,
-                    zero_point=output_zero_point,
+                    zero_point=output_zero_point.type_as(tensors[0].zero_point), # We assume all zero point have the same type, otherwise the check_zero_point fails anyway
```
volcacius, line 313: if we assume that all zero point have the same type, why are we casting it? Is "anyway" always? Comment is not really clear

tests/brevitas_end_to_end/test_torchvision_models.py
Giuseppe5: Yes it should

src/brevitas/quant_tensor/__init__.py
Giuseppe5: if you have two zero point tensors with different data types (e.g., torch.uint8 and torch.int, even if both tensors are 0), this check fails

src/brevitas/quant_tensor/__init__.py
Giuseppe5: The two zero point compared have to have the same type.
This type can vary anyway (in some circumstances could be torch.int, in other torch.uin8), from what I saw during this end-to-end testing

volcacius: do you have an example for this? in principle it should always be a float

src/brevitas/quant_tensor/__init__.py
volcacius: I would clarify then the check can failr when the zero point data types are different. Torch.allclose doesn't show up here

src/brevitas/quant_tensor/__init__.py
Giuseppe5: In export mode, the zero points are cast to either int8 or uint8 (if the quantization is signed or unsigned).

After removing the line `type_as`, I wasn't able to reproduce the original error so this will be removed for the moment.

.github/workflows/utils.py
```diff
@@ -20,11 +20,12 @@ class CustomTemplate(Template):
 
 class Action:
 
-    def __init__(self, name, exclude_list, matrix, step_list):
+    def __init__(self, name, exclude_list, matrix, step_list, strategy_params = od([('fail-fast', 'false')])):
```
Giuseppe5, line 23: We can leave it like this, or have a default STRATEGY constant in the gen_github_actions.py file (similar to what we do for EXCLUDE_LIST.

I think I prefer this second option now that I think about it

volcacius (approved):

volcacius:
Rebase changes from the last commit into the previous ones plz

Giuseppe5:
Yes boss

PR #478, opened 2022-12-09
Onnx notebook export

Giuseppe5:


Giuseppe5:
Docs wired done in #486

PR #482, opened 2022-12-12
Feat (flexml): now finalize_collect_stats is automatically applied at…

Giuseppe5:
… the end of calibration


volcacius:
Calibration tests need to be updated too

volcacius (approved):

Issue #400, opened 2022-05-18
Improvement of equalization method.


ZhangZhiPku:
Hi, developers of brevitas.
I once worked with vitisAI team in Xilinx for serval months (2020~2021, internship), and now I still work for creating better network quantization tools in Sensetime.

In 2020, we found some equalization tricks that can greatly enhance the performance(accuracy) of quantized network. As we all know, Xilinx FPGA requires bias to be quantized with 8 bit-width, 8-bit bias is a very distinctive and restrictive policy in network quantization. To meet this requirement, we have to control and limit the range of bias, especially for algorithm like equalization, as it will apply a scale factor on both weight and bias. There are some scale factors can reduce both weight and bias range, however if we don't take bias into consideration at first, scale factor solved by torch.sqrt(srcs_range / sinks_range) might lead to an extremely large bias. 

So here is our solution: when calculating the scale factor of layers, we first concatenate weight and bias into a single tensor, then use the generated tensor to compute the scale factor instead. Scale factor generated this way will be affected by the value of bias and balance weight and bias together. In our Expriments, this little trick serves as a crucial improvement in vitisAI, would you like to introduce this feature into brevitas in the future?

refer to: https://github.com/Xilinx/Vitis-AI/blob/master/tools/Vitis-AI-Quantizer/vai_q_pytorch/nndct_shared/optimization/commander.py

refer to https://github.com/openppl-public/ppq/blob/master/ppq/quantization/algorithm/equalization.py

volcacius:
HI @ZhangZhiPku,

Thanks for the suggestion. I experimented with the strategy you are describing in the past, but at the time I didn't notice much of a difference with the models I was working with, probably because they had higher precision integer bias (16b or 32b). I should probably revisit it and incorporate it.

Giuseppe5:
#534 implemented this method, closing this issue.

Thanks for your contribution!

PR #510, opened 2023-01-17
Graph Eq improvements

Giuseppe5:
- ~~Improved computation of scale for equalization~~ (see #400)
- ~~Equalize through concat~~
- ~~Equalize through BN~~

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -4,6 +4,8 @@
 
 import copy
 from functools import partial
+from copy import deepcopy
+from typing import Dict, Any
 import operator
 from random import sample
 from typing import Any, Dict
@@ -115,7 +117,36 @@ def _get_output_axis(module):
         raise RuntimeError(f"Module {module} not supported.")
 
 
-def _cross_layer_equalization(srcs, sinks):
+def _combine_weights_bias(m, axis, bias_shrinkage):
+    bias = m.bias.data
+    weight = m.weight.data if axis == 0 else m.weight.data.transpose(1,0)
+    weight = deepcopy(weight).view(weight.shape[0], -1)
+    bias = deepcopy(bias).view(-1, 1)
+    
+    weight = torch.where(torch.abs(weight) < 1e-8, torch.tensor(1e-8), weight)
+    factor = torch.abs(bias)/torch.abs(weight)
+    
+    # From https://github.com/Xilinx/Vitis-AI/blob/master/src/vai_quantizer/vai_q_pytorch/nndct_shared/optimization/commander.py#L450
+    if bias_shrinkage == 'vaiq':
+        if torch.abs(bias).max() < 10 and (torch.abs(bias).max()/torch.abs(weight).max() ) < 20:
+            if torch.median(factor) > 100 or torch.mean(factor) > 1000:
+                shrink_factor = 5
+            else:
+                shrink_factor = 2
+        else:
+            if  torch.median(factor) > 30 or torch.mean(factor) > 500:
+                shrink_factor = 20
+            elif torch.median(factor) > 15 or torch.mean(factor) > 250:
+                shrink_factor = 10
+            else:
+                shrink_factor = 5
+    else:
+        shrink_factor = bias_shrinkage
+    weight_bias = torch.cat([weight, bias/shrink_factor], 1)
+    return weight_bias
+
+
+def _cross_layer_equalization(srcs, sinks, merge_bias=True, bias_shrinkage='vaiq'):
     """
     Given two adjacent tensors', the weights are scaled such that
     the ranges of the first tensors' output channel are equal to the
@@ -135,7 +166,10 @@ def _cross_layer_equalization(srcs, sinks):
         raise RuntimeError("Output channels of sources do not match input channels of sinks")
 
     transpose = lambda module, axis: module.weight if axis == 0 else module.weight.transpose(0, 1)
-    src_weights = [transpose(m, axis) for m, axis in src_axes.items()]
+    if merge_bias:
+        src_weights = [_combine_weights_bias(m, axis, bias_shrinkage) for m, axis in src_axes.items()]
+    else:
+        src_weights = [transpose(m, axis) for m, axis in src_axes.items()]
     sink_weights = [transpose(m, axis) for m, axis in sink_axes.items()]
     srcs_range = _channel_range(torch.cat([w.reshape(w.size(0), -1) for w in src_weights], 1))
     sinks_range = _channel_range(torch.cat([w.reshape(w.size(0), -1) for w in sink_weights], 1))
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -29,7 +29,10 @@ _supported_layers = (
     torch.nn.Conv1d,
     torch.nn.Conv2d,
     torch.nn.Conv3d,
-    torch.nn.Linear)
+    torch.nn.Linear,
+    torch.nn.BatchNorm1d,
+    torch.nn.BatchNorm2d,
+    torch.nn.BatchNorm3d,)
 
 _scale_invariant_layers = (
     torch.nn.Dropout,
@@ -51,14 +54,15 @@ _residual_methods = (
     'add_'
 )
 
-_residual_cat_fns = (
+_residual_fns = (
     torch.add,
     operator.add,
     operator.iadd,
     operator.__add__,
     operator.__iadd__,
-    torch.cat
 )
+
+_cat_fns = (torch.cat,)
 
 _batch_norm = (
     torch.nn.BatchNorm1d,
@@ -91,6 +95,8 @@ def _get_size(axes, check_same=False):
 def _get_input_axis(module):
     if isinstance(module, torch.nn.Linear):
         return 1
+    elif isinstance(module, (torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d)):
+        return 0
     elif isinstance(module, (torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d)):
         if module.groups == 1:
             return 1
@@ -110,7 +116,7 @@ def _get_input_axis(module):
 
 
 def _get_output_axis(module):
-    if isinstance(module, (torch.nn.Linear, torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d)):
+    if isinstance(module, (torch.nn.Linear, torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d, torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d)):
         return 0
     elif isinstance(module, (torch.nn.ConvTranspose1d, torch.nn.ConvTranspose2d, torch.nn.ConvTranpose3d)):
         return 1
@@ -119,6 +125,9 @@ def _get_output_axis(module):
 
 
 def _combine_weights_bias(m, axis, bias_shrinkage):
+    if m.bias is None:
+        return m.weight.data
+
     bias = m.bias.data
     weight = m.weight.data if axis == 0 else m.weight.data.transpose(1,0)
     weight = deepcopy(weight).view(weight.shape[0], -1)
@@ -146,7 +155,7 @@ def _combine_weights_bias(m, axis, bias_shrinkage):
     weight_bias = torch.cat([weight, bias/shrink_factor], 1)
     return weight_bias
 
-def _cross_layer_equalization(srcs, sinks, merge_bias=True, bias_shrinkage='vaiq'):
+def _cross_layer_equalization(srcs, sinks, merge_bias, bias_shrinkage):
     """
     Given two adjacent tensors', the weights are scaled such that
     the ranges of the first tensors' output channel are equal to the
@@ -169,8 +178,6 @@ def _cross_layer_equalization(srcs, sinks, merge_bias=True, bias_shrinkage='vaiq
         is_concat = False
     else:
         return
-        # raise RuntimeError("Output channels of sources do not match input channels of sinks")
-
 
     transpose = lambda module, axis: module.weight if axis == 0 else module.weight.transpose(0, 1)
     if merge_bias:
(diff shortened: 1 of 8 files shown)


Giuseppe5:
Inception v3 does not work because of consecutive concat nodes without any Weight Layer in between

Giuseppe5:
Tests fails because this PR depends on #509

src/brevitas/graph/equalize.py
```diff
@@ -196,22 +293,44 @@ def walk_region(graph_model: GraphModule, starting_node: Node, history, srcs, si
...
+                _clear_sinks(sinks)
+                continue
+            # If this is the first concat we see, we register it while we keep exploring the graph.
+            # NOTE: seen_concat=node is better than seen_concat=True if we decide to support
+            # partial equalization through concats. No difference otherwise.
+
+            # If any of the input to cat has multiple users, we cannot equalize through
+            for n in node.all_input_nodes:
+                if len(n.users)>1:
+                    return
```
Giuseppe5, line 325: Not sure if here we should do `continue` of the outer loop or return or it's the same thing @volcacius

src/brevitas/graph/equalize.py
```diff
@@ -107,15 +116,46 @@ def _get_input_axis(module):
...
+def _combine_weights_bias(m, axis, bias_shrinkage):
+    if m.bias is None:
+        return m.weight.data
+
+    bias = m.bias.data
+    weight = m.weight.data if axis == 0 else m.weight.data.transpose(1,0)
+    weight = deepcopy(weight).view(weight.shape[0], -1)
+    bias = deepcopy(bias).view(-1, 1)
+
+    weight = torch.where(torch.abs(weight) < 1e-8, torch.tensor(1e-8), weight)
```
volcacius, line 136: Two things:
- can we reuse the existing small epsilon constant that we already defined, instead of 1e-8?
- Whenever you are creating a new tensor with torch.tensor, you should always be mindful of its device and dtype. For all we know weight could be float16 on GPU, and in that case torch.where would fail. In this case it's a good idea to set device and dtype to the same as weight

src/brevitas/graph/equalize.py
```diff
@@ -107,15 +116,46 @@ def _get_input_axis(module):
...
+            else:
+                shrink_factor = 2
+        else:
+            if  torch.median(factor) > 30 or torch.mean(factor) > 500:
+                shrink_factor = 20
+            elif torch.median(factor) > 15 or torch.mean(factor) > 250:
+                shrink_factor = 10
+            else:
+                shrink_factor = 5
+    else:
```
volcacius, line 153:
```
elif isinstance(bias_shrinkage, (int, float)):
    ....
else:
   raise RuntimeError(f"{bias_shrinkage} not supported.")
```

Always check the correctness of each case, and raise an error otherwise.

src/brevitas/graph/equalize.py
```diff
@@ -166,8 +225,16 @@ def _equalize(model, regions, iterations):
...
+        print(f"Equalize iter {i}")
+        scale_factor_max = torch.tensor(0.0)
         for region in regions:
-            _cross_layer_equalization([name_to_module[n] for n in region[0]], [name_to_module[n] for n in region[1]])
+            scale_factors_region = _cross_layer_equalization([name_to_module[n] for n in region[0]], [name_to_module[n] for n in region[1]], merge_bias, bias_shrinkage)
+            scale_factor_region_max = torch.max(torch.abs(1-scale_factors_region))
+            scale_factor_max = torch.max(scale_factor_max, scale_factor_region_max)
+        if threshold is not None:
+            if scale_factor_max < threshold:
+                print("Treshold reached, early stopping")
```
volcacius, line 236: No prints in the codebase unless it's something that a user can easily remove. If anything, we should add support for a proper logger

src/brevitas/graph/equalize.py
```diff
@@ -196,22 +303,45 @@ def walk_region(graph_model: GraphModule, starting_node: Node, history, srcs, si
...
+                walk_region(graph_model, node, history, srcs, sinks, walk_forward=True, seen_concat=seen_concat)
+                walk_region(graph_model, node, history, srcs, sinks, walk_forward=False, seen_concat=seen_concat)
+        elif node.target in _cat_fns:
+            if seen_concat is not None:
+                # If we go through another concat, then no equalization should be done.
+                # Removing all the sinks should guarantee that.
+                _clear_sinks(sinks)
+                continue
+            # If this is the first concat we see, we register it while we keep exploring the graph.
+            # NOTE: seen_concat=node is better than seen_concat=True if we decide to support
```
volcacius, line 329: This comment is not clear to me. Do we support seen_concat=node?

src/brevitas/graph/equalize.py
```diff
@@ -222,27 +352,34 @@ def _extract_regions(graph_model: GraphModule):
         if node.op == 'call_module':
             module = get_module(graph_model, node.target)
             if isinstance(module, _supported_layers):
-                srcs, sinks = {node.target}, set()
+                srcs = {node.target: 0}
+                sinks = dict()
                 walk_region(graph_model, node, set(), srcs, sinks, walk_forward=True)
                 if sinks:
                     # each region should appear only once, so to make it hashable
                     # we convert srcs and sinks to ordered lists first, and then to tuples
```
volcacius, line 360: This comment doesn't seem to apply anymore given that they are not lists anymore

src/brevitas/graph/equalize.py
```diff
@@ -222,27 +352,34 @@ def _extract_regions(graph_model: GraphModule):
...
-                    regions.add((tuple(sorted(srcs)), tuple(sorted(sinks))))
-    # for clarity, sort by the of the first source
```
volcacius, line 231: Why is the comment removed? Aren't we still sorting?

src/brevitas/graph/equalize.py
```diff
@@ -222,27 +352,34 @@ def _extract_regions(graph_model: GraphModule):
...
+                    regions.add((tuple(srcs_ordered), tuple(sorted(sinks.keys()))))
     regions = sorted(regions, key=lambda region: region[0][0])
     return regions
 
 
+
 class EqualizeGraph(GraphTransform):
 
-    def __init__(self, iterations) -> None:
+    def __init__(self, iterations, threshold=None, merge_bias=True, bias_shrinkage='vaiq') -> None:
```
volcacius, line 372: We need to provide a good default threshold and number of iterations. Something like 0.95 and 10?

tests/brevitas/graph/test_equlization.py
```diff
@@ -0,0 +1,78 @@
...
+    model = value_trace(model)
+    model = TorchFunctionalToModule().apply(model)
+    model = DuplicateSharedStatelessModule().apply(model)
+    model = ModuleToModuleByClass(nn.ReLU6, nn.ReLU).apply(model)
+    model = MeanMethodToAdaptiveAvgPool2d().apply(model)
+    model = AdaptiveAvgPoolToAvgPool().apply(model, inp)
+    model = CollapseConsecutiveConcats().apply(model)
+    model = MoveSplitBatchNormBeforeCat().apply(model)
+    model = MergeBatchNorm().apply(model)
+    model, regions = EqualizeGraph(1).apply(model)
```
volcacius, line 61: Testing with just 1 iterations doesn't seem like a good idea, do like 3.

src/brevitas/graph/equalize.py
```diff
@@ -166,8 +225,16 @@ def _equalize(model, regions, iterations):
         if name in name_set:
             name_to_module[name] = module
     for i in range(iterations):
+        print(f"Equalize iter {i}")
```
volcacius, line 228: No print

src/brevitas/graph/equalize.py
```diff
@@ -166,8 +225,16 @@ def _equalize(model, regions, iterations):
...
+        scale_factor_max = torch.tensor(0.0)
```
volcacius, line 229: Same issue as before when it comes to torch.tensor. I would initialize it to None, and the allocate a tensor later on, after you compute scale_factor_region_max,  with the same device and dtype of scale_factor_region_max, checking if scale_factor_max is None

volcacius (changes requested):
We need as many comments on the logic as we can, right now we have almost zero and I don't understand anymore how it works, and docs

volcacius:
I would also add tests checking that things work on small group of layers, with patterns of combination of layers we care about, so that when it breaks when know what breaks.

src/brevitas/graph/equalize.py
```diff
@@ -106,16 +119,58 @@ def _get_input_axis(module):
...
         return 1
     else:
         raise RuntimeError(f"Module {module} not supported.")
 
 
-def _cross_layer_equalization(srcs, sinks):
+def _combine_weights_bias(module: torch.nn.Module, axis: int, bias_shrinkage: Union[int, float, str]):
+    """Combine weights and bias before graph equalizattion
+
+    This method merges the weight and bias sources, so that the resulting equalizer scale factor
```
Giuseppe5, line 134: weight and bias **of the** sources, typo

tests/brevitas/graph/test_equalization.py
```diff
@@ -0,0 +1,89 @@
...
+SEED = 123456
+IN_SIZE = (16,3,224,224)
+ATOL = 1e-3
+
+
+@pytest.mark.parametrize("model_name", MODELS)
+def test_equalization_torchvision_models(model_name: str):
+    try:
+        model = getattr(models, model_name)(pretrained=True, transform_input=False)
+    except:
```
volcacius, line 42: Can we specify the right exception to catch?

tests/brevitas/graph/test_equalization.py
```diff
@@ -0,0 +1,89 @@
...
+        if _is_supported_module(model, n):
+            count += 1
+
+    print(f"Source coverage {len(srcs)/count}")
+    print(f"Sink coverage {len(sinks)/count}")
+    assert torch.allclose(expected_out, out, atol=ATOL)
+
+
+
+
```
volcacius: Useless white space, should be 2 lines

volcacius (changes requested):

src/brevitas/graph/equalize.py
```diff
@@ -56,38 +59,48 @@
...
     torch.nn.BatchNorm1d,
     torch.nn.BatchNorm2d,
     torch.nn.BatchNorm3d,
 )
 
 
-def _channel_range(inp):
-    mins, _ = inp.min(dim=1)
-    maxs, _ = inp.max(dim=1)
+def _channel_range(module: torch.nn.Module):
```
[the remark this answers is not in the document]
Giuseppe5: It's a tensor, fixed

src/brevitas/graph/equalize.py
```diff
@@ -106,16 +119,58 @@ def _get_input_axis(module):
...
+    """
+    if module.bias is None:
+        return module.weight.data
+
+    bias = module.bias.data
+    weight = module.weight.data if axis == 0 else module.weight.data.transpose(1,0)
+    weight = deepcopy(weight).view(weight.shape[0], -1)
+    bias = deepcopy(bias).view(-1, 1)
+
+    weight = torch.where(torch.abs(weight) < EPSILON, torch.tensor(EPSILON).type_as(weight), weight)
```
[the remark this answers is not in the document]
Giuseppe5: Yes, check this: https://github.com/pytorch/pytorch/issues/33662

src/brevitas/graph/equalize.py
```diff
@@ -126,65 +181,169 @@ def _cross_layer_equalization(srcs, sinks):
...
+    sink_axes = OrderedDict([(m, _get_input_axis(m)) for m in sinks])
+    src_sink = _get_size(src_axes)
+    sink_size = _get_size(sink_axes, check_same=True)
 
-    if src_size != sink_size:
-        raise RuntimeError("Output channels of sources do not match input channels of sinks")
+
+    if torch.sum(src_sink) == sink_size[0] and len(src_sink)>1:
+        is_concat = True
+    elif torch.mean(src_sink) == sink_size[0]: # Workaround to check src_sink have all the same size
```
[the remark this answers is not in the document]
Giuseppe5: Fixed

src/brevitas/graph/equalize.py
```diff
@@ -126,65 +181,169 @@ def _cross_layer_equalization(srcs, sinks):
...
+        module.weight.data = module.weight.data * torch.reshape(partial_inverse_scaling, src_broadcast_size)
+        if is_concat:
+            start_index += channels
     for module, axis in sink_axes.items():
         src_broadcast_size = [1] * module.weight.ndim
         src_broadcast_size[axis] = module.weight.size(axis)
+
         module.weight.data = module.weight.data * torch.reshape(scaling_factors, src_broadcast_size)
+        if isinstance(module, (torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d)):
+            module.running_mean.data = module.running_mean.data * torch.reshape(inverse_scaling_factors, src_broadcast_size)
```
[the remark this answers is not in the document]
Giuseppe5: Yes, modify running mean and the learned weight allows us to perform the rescale without altering the "additive term" of BN (that depends on the learned parameters and the running mean)

src/brevitas/graph/equalize.py
```diff
@@ -126,65 +181,169 @@ def _cross_layer_equalization(srcs, sinks):
...
+def _equalize(model: GraphModule, regions: tuple, iterations: int, threshold: Union[float, None], merge_bias: bool, bias_shrinkage: Union[int, float, str]):
     """
     Generalized version of section 4.1 of https://arxiv.org/pdf/1906.04721.pdf
     """
     name_to_module : Dict[str, torch.nn.Module] = {}
-    name_set = {name for region in regions for module_set in region for name in module_set}
+    # name_set = {name for region in regions for module_set in region for name in module_set}
+    name_set = set()
+    for region in regions:
+        for src_reg in region[0]:
```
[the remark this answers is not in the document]
Giuseppe5: Regions is a set, I made a mistake in the annotations

src/brevitas/graph/target/flexml.py
```diff
@@ -378,7 +378,7 @@ def flexml_avgpool_handler(model, *model_args, avgpool_to_depthwise_conv=False,
     return model
 
 
-def preprocess_flexml(model, *model_args, equalization_iters = 0, **model_kwargs):
+def preprocess_flexml(model, *model_args, equalization_iters = 30, equalize_threshold = 0.05, **model_kwargs):
```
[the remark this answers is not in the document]
Giuseppe5: We don't, I put a "high" value because of equalize_threshold would stop sooner if needed. I could revert it to zero if needed

PR #514, opened 2023-01-20
Flexml: minor improvements

Giuseppe5:


volcacius:
Please separate into commits for disabling narrow range on FlexML target and fixing transformation for disabling return quant tensor.

Giuseppe5:
Done

PR #516, opened 2023-01-23
Remove Empty QuantTensor

Giuseppe5:


src/brevitas/proxy/runtime_quant.py
```diff
@@ -162,7 +162,7 @@ def forward(self, x: Union[Tensor, QuantTensor]) -> QuantTensor:
             if isinstance(x, QuantTensor):  # passthrough
```
Giuseppe5, line 162: just return x

src/brevitas/nn/quant_layer.py
```diff
@@ -357,12 +368,15 @@ def forward_impl(self, inp: Union[Tensor, QuantTensor]) -> Union[Tensor, QuantTe
             elif quant_input.zero_point is not None and output_zero_point is None:
```
Giuseppe5, line 368: quant_input may be normal Tensor

src/brevitas/nn/quant_layer.py
```diff
@@ -357,12 +368,15 @@ def forward_impl(self, inp: Union[Tensor, QuantTensor]) -> Union[Tensor, QuantTe
...
                 output_zero_point = quant_input.zero_point
 
-        quant_output = QuantTensor(
-            value=output_tensor,
-            scale=output_scale,
-            zero_point=output_zero_point,
-            bit_width=output_bit_width,
-            signed=output_signed,
-            training=self.training)
+        if isinstance(quant_input, QuantTensor):
```
Giuseppe5, line 371: Clarify condition with comments.

Giuseppe5: Add isinstance(quant_weight, QuantTensor)

src/brevitas/nn/quant_layer.py
Giuseppe5: It can't be normal tensor because I check before for this condition and raise an error if it happens

Giuseppe5:
This is blocked by #518 and #519

PR #518, opened 2023-01-25
CI: Re-enable JIT tests

Giuseppe5:
Re-enable JIT tests with torch >= 1.10

### DIFF
--- a/.github/workflows/gen_github_actions.py
+++ b/.github/workflows/gen_github_actions.py
@@ -39,7 +39,8 @@ FINN_PLATFORM_LIST = ['windows-latest', 'ubuntu-latest']
 
 STRATEGY = od([('fail-fast', 'false')])
 
-EXCLUDE_LIST = [
+EXCLUDE_LIST = []
+JIT_EXCLUDE_LIST = [
     od([('pytorch_version', ['1.5.1', ]),
          ('jit_status', ['jit_enabled',])]),
     od([('pytorch_version', ['1.6.0', ]),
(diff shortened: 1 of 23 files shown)


tests/marker.py
```diff
@@ -28,3 +29,10 @@ def skip_wrapper(f):
         return pytest.mark.skipif(skip, reason=f'Requires Pytorch < {pt_version}')(f)
 
     return skip_wrapper
+
+def requires_jit_disabled():
```
volcacius, line 33: we make the reason an arg with default the string below, so that it can be edited depending on context. When we disable because of export, we say so, when it's because of mock, we say so

tests/brevitas/core/test_ternary_quant.py
```diff
@@ -32,6 +33,7 @@ class TestTernaryUnit:
         inp=float_tensor_random_shape_st(),
         scale_init=scalar_float_p_tensor_st(),
         threshold=st.floats(min_value=0.0, max_value=1.0))
+    @requires_jit_disabled() # Mock conflicts with JIT
```
volcacius, line 36: Missing newline

.github/workflows/gen_github_actions.py
```diff
@@ -40,7 +40,18 @@
 STRATEGY = od([('fail-fast', 'false')])
 
 EXCLUDE_LIST = []
-
+JIT_EXCLUDE_LIST = [
```
volcacius, line 43: this can be a function generating the list

Giuseppe5: Creating the function that generalizes to the different configuration of Exclude List is possible but not necessarily quick, so I would post-pone this

volcacius (approved):
LGTM

PR #522, opened 2023-02-02
Model updates

Giuseppe5:

Issue #517, opened 2023-01-23
Looking for guidance to replicate MobileNet-v1 quantized model


verderog:
I'm trying to replicate the pre-trained quantized version of MobileNet-v1 and am looking for some guidance into what I doing wrong or may be fundamentally misunderstanding. I've created a simple Python script (below) to perform a one-shot inference to compare the output of the floating-point model (https://github.com/osmr/imgclsmob/blob/master/pytorch/pytorchcv/models/mobilenet.py), pre-trained quantized model (https://github.com/Xilinx/brevitas/blob/master/src/brevitas_examples/imagenet_classification/models/mobilenetv1.py), and untrained quantized model retrained with the floating point.  It does this inference for the first image from the ImageNet validation dataset:

```
#!/usr/bin/env python3

from pytorchcv.model_provider import get_model as ptcv_get_model
import torch
from brevitas_examples.imagenet_classification import models
from brevitas import config
import torchvision.transforms.functional as TF

from PIL import Image

config.IGNORE_MISSING_KEYS = True

test_file = "/srv/nfs/imagenet-data/imagenet/ILSVRC2012_val_00000001.JPEG"

def tf_pre_process(img):
    img = TF.resize(img,(224,224))
    img = TF.to_tensor(img) # NOTE: this converts from [0,255] range to [0,1]
    img = TF.normalize(img,mean=[0.485, 0.456, 0.406],std=[0.229, 0.224, 0.225])
    return img

def main():
    # instantiate the floating point model, including pre-trained weights and activations
    net_float = ptcv_get_model("mobilenet_w1", pretrained=True) # https://github.com/osmr/imgclsmob/blob/master/pytorch/pytorchcv/models/mobilenet.py

    # instantiate the quantized model, with pre-training.
    net_quant = models.quant_mobilenet_v1_4b(pretrained=True) # https://github.com/Xilinx/brevitas/blob/master/src/brevitas_examples/imagenet_classification/models/mobilenetv1.py

    # instantiate the quantized model, with no pre-training.  Want to propagate the pre-trained floating point weights and activations    
    net_quant_retrained = models.quant_mobilenet_v1_4b(pretrained=False)
    net_quant_retrained.load_state_dict(net_float.state_dict()) # 'retrain' the untrained, quantized model with the floating-point dictionary (https://xilinx.github.io/brevitas/tutorials/tvmcon2021.html#Retraining-from-floating-point)

    net_float.eval()
    net_quant.eval()
[...]

PR #523, opened 2023-02-07
Fix: eval mode and steps fix

Giuseppe5:
This PR should address part of the issues of #517, and fix all of them with #522. 

Changes:

- ~~The model is put in eval mode before calibration and eval~~
- The _collect\_stats\_steps_ is extended during PTQ
- _finalize\_collect\_stats_ now avoids that the pre-process is applied twice, with incorrect results

### DIFF
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -4,6 +4,7 @@
 
 from abc import ABC
 from functools import partial
+import sys
 
 from torch import nn
 import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


src/brevitas/graph/calibrate.py
```diff
@@ -45,10 +46,15 @@
     nn.ConvTranspose2d,
     nn.ConvTranspose3d)
 
+# We extend the collect steps in PTQ to match potentially long calibrations
+def extend_collect_stats_steps(module):
+    if hasattr(module, 'collect_stats_steps') and not JIT_ENABLED:
+        module.collect_stats_steps = 1e5
```
volcacius, line 52: extract this to a COLLECT_STATS_STEPS constant at the top

Giuseppe5:
@volcacius rebased

volcacius (approved):

PR #528, opened 2023-02-13
Feat (QuantTensor): support to move QuantTensor across devices

Giuseppe5:


src/brevitas/quant_tensor/__init__.py
```diff
@@ -320,6 +320,37 @@ def __neg__(self):
...
                 training=self.training)
 
+
+    def to(self, *args, **kwargs):
+        return QuantTensor(
+            self.value.to(*args, **kwargs),
+            self.scale.to(*args, **kwargs) if self.scale is not None else None,
+            self.zero_point.to(*args, **kwargs) if self.zero_point is not None else None,
+            self.bit_width.to(*args, **kwargs) if self.bit_width is not None else None,
+            self.signed,
```
volcacius, line 330: this should be the signed_t tensor, and similarly for training_t, across all methods. The retrieval property should also be updated to make sure it doesn't break whenever signed_t and training_t are not on CPU (item() fails I think on a GPU tensor).

volcacius (changes requested):

PR #533, opened 2023-02-14
Graph Eq: torchvision end to end fix and tests

Giuseppe5:
This depends on #532

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -95,7 +95,7 @@ def _get_size(axes: Dict[nn.Module, int]) -> int:
     size = m0.weight.size(axis0)
     for m, axis in axes.items():
         if m.weight.size(axis) != size:
-            raise RuntimeError("Weights sizes don't match")
+            return None
     return size
 
 
@@ -109,17 +109,13 @@ def _get_input_axis(module: nn.Module) -> int:
             return 1
         elif module.groups == module.out_channels:
             return 0
-        else:
-            raise RuntimeError("Group convolution not supported")
     elif isinstance(module, (nn.ConvTranspose1d, nn.ConvTranspose2d, nn.ConvTranspose3d)):
         if module.groups == 1:
             return 0
         elif module.groups == module.out_channels:
             return 1
-        else:
-            raise RuntimeError("Group convolution not supported")
     else:
-        raise RuntimeError(f"Module {module} not supported.")
+        return None
 
 
 def _get_output_axis(module: nn.Module) -> int:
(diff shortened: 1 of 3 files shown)


tests/brevitas/graph/test_equalization.py
```diff
@@ -1,20 +1,54 @@
...
+    count= 0
+    for r in regions:
+        srcs.update(list(r[0]))
+        sinks.update(list(r[1]))
+
+    for n in model.graph.nodes:
+        if _is_supported_module(model, n):
+            count += 1
+
+    print(f"Source coverage {len(srcs)/count}")
```
volcacius, line 49: can we put asserts here of the expected coverage rather than prints?

tests/brevitas/graph/test_equalization.py
```diff
@@ -1,20 +1,54 @@
...
+    except TypeError:
+        model = getattr(models, model_name)(pretrained=True)
+
+    torch.manual_seed(SEED)
+    inp = torch.randn(IN_SIZE)
+    model.eval()
+    expected_out = model(inp)
+
+    input_name = getfullargspec(model.forward)[0][0]
+    model = value_trace(model, {input_name: inp})
```
volcacius, line 33: do we need to do value trace? can we get away with pytorch symbolic traces?

volcacius (changes requested):

src/brevitas/graph/equalize.py
```diff
@@ -145,11 +145,16 @@ def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module]):
 
     src_axes = {m: _get_output_axis(m) for m in srcs}
     sink_axes = {m: _get_input_axis(m) for m in sinks}
-    src_size = _get_size(src_axes)
-    sink_size = _get_size(sink_axes)
+    try:
+        src_size = _get_size(src_axes)
+        sink_size = _get_size(sink_axes)
+    except RuntimeError:
```
volcacius, line 151: RuntimeError is not supposed to be internally catched like this. We should define an an exception, and then catch that.

tests/brevitas/graph/equalization_fixtures.py
```diff
@@ -3,6 +3,15 @@
 import torch
```
volcacius, line 3: add copyright and license headers

volcacius (changes requested):

src/brevitas/graph/equalize.py
```diff
@@ -110,16 +110,16 @@ def _get_input_axis(module: nn.Module) -> int:
...
         if module.groups == 1:
             return 0
         elif module.groups == module.out_channels:
             return 1
         else:
-            raise RuntimeError("Group convolution not supported")
+            return None
     else:
-        raise RuntimeError(f"Module {module} not supported.")
+        return None
```
volcacius, line 122: You can just return None once at the end and remove all the else branches

volcacius (approved):

PR #535, opened 2023-02-14
Graph Eq: early exit based on threshold value

Giuseppe5:
This depends on #534

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -169,16 +169,20 @@ def _combine_weights_bias(weight: nn.parameter.Parameter, bias_shrinkage: Union[
     return weight_bias
 
 
-def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], merge_bias: bool, bias_shrinkage: Union[float, str]):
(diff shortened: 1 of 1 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -174,29 +174,33 @@ def _combine_weights_bias(weight: nn.parameter.Parameter, bias_shrinkage: Union[
...
 
 
-def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], merge_bias: bool, bias_shrinkage: Union[float, str]):
+def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], merge_bias: bool, bias_shrinkage: Union[float, str]) -> torch.Tensor:
     """
     Given two adjacent tensors', the weights are scaled such that
     the ranges of the first tensors' output channel are equal to the
     ranges of the second tensors' input channel
     """
+    # Determine device of tensors
```
volcacius, line 183: The dtype should be extract and used too, otherwise it might lead to mismatches.

volcacius (changes requested):

volcacius (approved):

PR #541, opened 2023-02-20
Feat (ptq): activation equalization support

Giuseppe5:
Changes:

- walk_region extended to keep track of activations encoutered in the path, 
- walk_region can walk through scale varying activations (e.g., sigmoid)

**Activation Equalization support**:
- Possibility to set and tune alpha
- For now, if there are multiple activations in one region, no equalization is performed (it could be mathematically possible but I have doubt about its usefulness)
- If scale varying activations are used, then a standalone MUL op is added after the activation. 

~~Missing tests. On a FP model, this transformation does not alter the math of the model, similar to graph equalization. Tests will be performed in a similar fashion.~~

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -11,7 +11,9 @@ from typing import Callable, Dict, List, Optional, Set, Tuple, Union
 import warnings
 
 import torch
+from torch.fx import GraphModule as TorchGraphModule
 import torch.nn as nn
+import torch.nn.functional as F
 
 from brevitas.fx import GraphModule
 from brevitas.fx import Node
@@ -65,7 +67,16 @@ _select_op = (operator.getitem, operator.__getitem__)
 _scale_invariant_activations = (torch.nn.ReLU,)
 
 _scale_varying_activations = (
-    torch.nn.Sigmoid, torch.nn.Tanh, torch.nn.ReLU6, torch.nn.GELU, torch.nn.SiLU)
+    torch.nn.Sigmoid,
+    torch.nn.Tanh,
+    torch.nn.ReLU6,
+    torch.nn.GELU,
+    torch.nn.SiLU,
+    F.sigmoid,
+    F.tanh,
+    F.relu6,
+    F.gelu,
+    F.silu)
 
 _residual_methods = ('add', 'add_')
 
@@ -110,16 +121,21 @@ def _select_scale_computation_fn(
 
 class activation_equalization_mode:
 
-    def __init__(self, graph_model, alpha, add_mul_node, layerwise=False, enabled=True) -> None:
-        self.graph_model = graph_model
+    def __init__(self, model, alpha, add_mul_node=True, layerwise=True, enabled=True) -> None:
+        self.model = model
         self.alpha = alpha
         self.enabled = enabled
         self.add_mul_node = add_mul_node
         if layerwise:
-            self.add_mul_node = True
-            self.graph_act_eq = LayerwiseActivationEqualization(self.graph_model)
+            if not self.add_mul_node:
+                raise ValueError("Layerwise activation equalization requires add_mul_node")
+            self.graph_act_eq = LayerwiseActivationEqualization(self.model)
         else:
-            self.graph_act_eq = GraphActivationEqualization(self.graph_model, self.add_mul_node)
+            if not isinstance(self.model, (TorchGraphModule, GraphModule)):
+                raise TypeError(
+                    "A Torch FX representation of the model is needed for Graph Activation Equalization"
+                )
+            self.graph_act_eq = GraphActivationEqualization(self.model, self.add_mul_node)
         self.scale_factors = None
 
     def __enter__(self):
@@ -293,8 +309,12 @@ def _combine_weights_bias(
     return weight_bias
 
 
-def transpose(module, axis):
-    shape = list(range(module.weight.ndim))  #list(module.weight.shape)
+def transpose(module: torch.nn.Module, axis: int):
+    """
+    Given a module and an axis, this function re-arranges the module's weights so that the axis and
+    the first dimension are swapped
+    """
+    shape = list(range(module.weight.ndim))
     axis = shape[axis]
     shape.insert(0, axis)
     del shape[axis + 1]
@@ -315,9 +335,15 @@ def _cross_layer_equalization(
     the ranges of the first tensors' output channel are equal to the
     ranges of the second tensors' input channel
     """
+
     # Determine device and type of tensors
     device = next(sinks[0].parameters()).device
     dtype = next(sinks[0].parameters()).dtype
+
+    # If equalization criteria are not met, we return a scalar one to indicate that no equalization
+    # has been performed
+    def _no_equalize():
+        return torch.tensor(1., dtype=dtype, device=device)
 
     src_axes = {}
     sink_axes = {}
@@ -327,7 +353,7 @@ def _cross_layer_equalization(
     for i, module in enumerate(srcs):
         # If module is not supported, do not perform graph equalization
         if not isinstance(module, _supported_layers):
-            return torch.tensor(1., dtype=dtype, device=device)
+            return _no_equalize()
         if isinstance(module, nn.MultiheadAttention):
             srcs[i] = module.out_proj
         src_axes[srcs[i]] = _get_output_axis(module)
@@ -336,18 +362,18 @@ def _cross_layer_equalization(
     for i, module in enumerate(sinks):
         # If module is not supported, do not perform graph equalization
         if not isinstance(module, _supported_layers):
-            return torch.tensor(1., dtype=dtype, device=device)
+            return _no_equalize()
         # For MultiheadAttention, we support only self-attetion
         if isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is not None:
             # For sinks, we only need to modify the weight but not the bias
             sinks[i] = WeightBiasTuple(weight=module.in_proj_weight)
         elif isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is None:
-            return torch.tensor(1., dtype=dtype, device=device)
+            return _no_equalize()
         sink_axes[sinks[i]] = _get_input_axis(module)
         act_sink_axes[sinks[i]] = _get_act_axis(module)
 
     # If act_val is enabled, use source or sink weights to determine the activation channel
-    # For example, if the source is BatchNorm, we need to use the sink information
+    # For example, if the source is BatchNorm, we need to use the information coming from the sinks
     if list_of_act_val is not None:
         list_of_sink_axes = [x for x in list(act_sink_axes.values()) if x is not None]
         list_of_source_axes = [x for x in list(act_sources_axes.values()) if x is not None]
@@ -356,23 +382,23 @@ def _cross_layer_equalization(
         elif len(list_of_source_axes) > 0:
             act_axis = list_of_source_axes[0]
         else:
-            return torch.tensor(1., dtype=dtype, device=device)
+            return _no_equalize()
         # If there is a mismatch in the activation channel (e.g. a transpose/flatten op in between),
         # do not perform equalization
         if any([act_axis != axis for axis in list_of_source_axes + list_of_sink_axes]):
-            return torch.tensor(1., dtype=dtype, device=device)
+            return _no_equalize()
 
     # Check if any of the axis is None, which means that the module is not supported.
     # In that case, do not perform graph equalization
     axes_to_check = [*src_axes.values(), *sink_axes.values()]
     if None in axes_to_check:
-        return torch.tensor(1., dtype=dtype, device=device)
+        return _no_equalize()
 
     # Check if the sink_size is None,
     # which means that the some of the sinks do not have the same size as the others.
     sink_size = _get_size(sink_axes)
     if None in [sink_size]:
-        return torch.tensor(1., dtype=dtype, device=device)
+        return _no_equalize()
 
     scale_fn = _select_scale_computation_fn(scale_computation_type)
     sink_weights = [transpose(m, axis) for m, axis in sink_axes.items()]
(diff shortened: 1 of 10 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -93,22 +108,65 @@ def _select_scale_computation_fn(
...
-    maxs, _ = inp.max(dim=1)
+class activation_equalization_mode:
+
+    def __init__(self, graph_model, alpha, add_mul_node, layerwise=False, enabled=True) -> None:
+        self.graph_model = graph_model
+        self.alpha = alpha
+        self.enabled = enabled
+        self.add_mul_node = add_mul_node
+        if layerwise:
+            self.add_mul_node = True
```
volcacius, line 119: raise a ValueError when add_mul_node is False and layerwise is True.

src/brevitas/graph/equalize.py
```diff
@@ -93,22 +108,65 @@ def _select_scale_computation_fn(
         raise RuntimeError(f"Scale computation type {scale_computation_type} not supported")
 
 
-def _channel_range(inp: torch.Tensor) -> torch.Tensor:
-    mins, _ = inp.min(dim=1)
-    maxs, _ = inp.max(dim=1)
+class activation_equalization_mode:
+
+    def __init__(self, graph_model, alpha, add_mul_node, layerwise=False, enabled=True) -> None:
```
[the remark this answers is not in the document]
volcacius: Rename graph_model to model

src/brevitas/graph/equalize.py
```diff
@@ -93,22 +108,65 @@ def _select_scale_computation_fn(
...
+        self.graph_model = graph_model
+        self.alpha = alpha
+        self.enabled = enabled
+        self.add_mul_node = add_mul_node
+        if layerwise:
+            self.add_mul_node = True
+            self.graph_act_eq = LayerwiseActivationEqualization(self.graph_model)
+        else:
+            self.graph_act_eq = GraphActivationEqualization(self.graph_model, self.add_mul_node)
```
volcacius, line 122: We need to verify that it's indeed a GraphModule. Needs to be done both against the backport and the pytorch one.

src/brevitas/graph/equalize.py
```diff
@@ -220,23 +293,36 @@ def _combine_weights_bias(
     return weight_bias
 
 
+def transpose(module, axis):
+    shape = list(range(module.weight.ndim))  #list(module.weight.shape)
```
volcacius, line 297: Is the comment needed?

src/brevitas/graph/equalize.py
```diff
@@ -220,23 +293,36 @@ def _combine_weights_bias(
     return weight_bias
 
 
+def transpose(module, axis):
```
volcacius, line 296: Please add a comment of what this is doing, it's not very readable

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
             return torch.tensor(1., dtype=dtype, device=device)
         sink_axes[sinks[i]] = _get_input_axis(module)
+        act_sink_axes[sinks[i]] = _get_act_axis(module)
+
+    # If act_val is enabled, use source or sink weights to determine the activation channel
+    # For example, if the source is BatchNorm, we need to use the sink information
+    if list_of_act_val is not None:
+        list_of_sink_axes = [x for x in list(act_sink_axes.values()) if x is not None]
+        list_of_source_axes = [x for x in list(act_sources_axes.values()) if x is not None]
+        if len(list_of_sink_axes) > 0:
```
volcacius, line 354: What is happening here and why are we talking only the first one? Sould we check that they are the same? Add some comments

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+            act_axis = list_of_sink_axes[0]
+        elif len(list_of_source_axes) > 0:
+            act_axis = list_of_source_axes[0]
+        else:
+            return torch.tensor(1., dtype=dtype, device=device)
+        # If there is a mismatch in the activation channel (e.g. a transpose/flatten op in between),
+        # do not perform equalization
+        if any([act_axis != axis for axis in list_of_source_axes + list_of_sink_axes]):
+            return torch.tensor(1., dtype=dtype, device=device)
```
volcacius, line 363: can we have a return _no_equalize() instead defined inline above? More readable

src/brevitas/graph/equalize.py
```diff
@@ -74,13 +86,16 @@ class WeightBiasTuple:
 class Region:
     srcs: Tuple = field(default_factory=tuple)
     sinks: Tuple = field(default_factory=tuple)
+    acts: Tuple = field(default_factory=tuple)
```
volcacius, line 89: What is acts in region? Why do we need it?

src/brevitas/graph/equalize.py
```diff
@@ -192,7 +265,7 @@ def _combine_weights_bias(
         return weight.data
 
     bias = bias.data
-    weight = weight.reshape(weight.shape[0], -1)
+    weight = weight.data.reshape(weight.shape[0], -1)
```
volcacius, line 268: why do we need data?

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
-        src_weights = [
-            _combine_weights_bias(transpose(m, axis), bias_shrinkage, m.bias) for m,
-            axis in src_axes.items()]
-    else:
-        src_weights = [transpose(m, axis) for m, axis in src_axes.items()]
     sink_weights = [transpose(m, axis) for m, axis in sink_axes.items()]
-    srcs_range = scale_fn(torch.cat([w.reshape(w.size(0), -1) for w in src_weights], 1))
     sinks_range = scale_fn(torch.cat([w.reshape(w.size(0), -1) for w in sink_weights], 1))
-    sinks_range += EPSILON
+    sinks_range = torch.clamp(sinks_range, EPSILON)
```
volcacius, line 380: this isn't the same as adding epsilon. What happens when it's too small?

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+    # Determine the srcs_range based on where we are performing activation equalization or
+    # weight equalization
+    if list_of_act_val is not None:
+        act_shape = list_of_act_val[0].shape
+        # act_val = [WeightBiasTuple(act_val) for act_val in list_of_act_val]
+        list_of_act_val = [
+            transpose(WeightBiasTuple(act_val), act_axis) for act_val in list_of_act_val]
+        srcs_range = scale_fn(
+            torch.cat([act_val.reshape(act_val.size(0), -1) for act_val in list_of_act_val],
+                      1))  # scale_fn(act_val.reshape(act_val.size(0), -1))
```
volcacius, line 391: What is the commented out part?

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+    else:
+        # If we do weight equalization, perform additional check on source size
+        src_size = _get_size(src_axes)
+        # Exit if source and sink have different different sizes, or if sources contains None
+        if src_size != sink_size or None in [src_size]:
```
volcacius, line 396: Should we raise a warning on different sizes? Is that something that can happen during a correct execution?

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+            src_weights = [
+                _combine_weights_bias(transpose(m, axis), bias_shrinkage, m.bias) for m,
+                axis in src_axes.items()]
+        else:
+            src_weights = [transpose(m, axis) for m, axis in src_axes.items()]
+        srcs_range = scale_fn(torch.cat([w.reshape(w.size(0), -1) for w in src_weights], 1))
+
+    # If there is a mismatch between srcs and sinks values, exit
+    if srcs_range.shape != sinks_range.shape:
+        return torch.tensor(1., dtype=dtype, device=device)
```
volcacius, line 409: raise a warning?

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+
+    srcs_range = torch.pow(srcs_range, alpha)
+    sinks_range = torch.pow(sinks_range, 1 - alpha)
+    scaling_factors = srcs_range / sinks_range
+    scaling_factors = torch.clamp(scaling_factors, EPSILON)
```
volcacius, line 414: clamp_min?

src/brevitas/graph/equalize.py
```diff
@@ -361,7 +492,18 @@ def _is_supported_module(graph_model: GraphModule, node: Node) -> bool:
 
 def _is_scale_invariant_module(graph_model: GraphModule, node: Node) -> bool:
     return node.op == 'call_module' and isinstance(
-        get_module(graph_model, node.target), _scale_invariant_layers)
+        get_module(graph_model, node.target),
+        _scale_invariant_layers + _scale_invariant_activations)
+
+
+def _is_scale_invariant_activation(graph_model, node):
+    return node.op == 'call_module' and isinstance(
```
volcacius, line 500: shouldn't we check functionals too?

src/brevitas/graph/equalize.py
```diff
@@ -431,18 +577,30 @@ def find_sinks(graph_model: GraphModule, starting_node: Node,
             state.sinks.add(None)
```
volcacius, line 577: I think we should define a _UNSUPPORTED_OP = object() at the top and then add that and check with is when it's present, cause we are using None already for too much stuff

src/brevitas/graph/equalize.py
```diff
@@ -482,73 +640,206 @@ def apply(self,
...
+        if name not in self.float_act_map:
+            self.float_act_map[name] = self.scale_fn(out, dim=0)
+        else:
+            batch_data = torch.cat([self.float_act_map[name].unsqueeze(0), out], dim=0)
+            self.float_act_map[name] = self.scale_fn(batch_data, dim=0)
+
+    def insert_mul_node(self, scale, shape, axis, act_node):
+        broadcastable_shape = [1] * len(shape)
+        broadcastable_shape[axis] = shape[axis]
+        # Add Batch Dim
```
volcacius, line 837: I think we need more work around determining the batch size, e.g. for whenever batch_first=False in attention-y implementation.

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
+        if isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is not None:
             # For sinks, we only need to modify the weight but not the bias
             sinks[i] = WeightBiasTuple(weight=module.in_proj_weight)
-        elif isinstance(module, nn.MultiheadAttention) and not hasattr(sinks[i], 'in_proj_weight'):
+        elif isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is None:
             return torch.tensor(1., dtype=dtype, device=device)
         sink_axes[sinks[i]] = _get_input_axis(module)
+        act_sink_axes[sinks[i]] = _get_act_axis(module)
+
+    # If act_val is enabled, use source or sink weights to determine the activation channel
```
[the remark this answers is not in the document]
Giuseppe5: Sink weights are used immediately after we define scale_fn.
After computing sinks_range, we use the srcs weights in case of weight equalization or the activation runtime values in case of act equalization to determine the srcs_range.

src/brevitas/graph/equalize.py
Giuseppe5: It's convenient (but not fundamental) for the following:

- When equalizing through scale_varying activations, these are added as srcs, and we need to filter them out before equalizing
- When equalizing through scale_varying activations, we cannot equalize if a region contains two of them, so we need to keep track of how many we have encoutered

Originally there were more restrictions and acts was more useful than now. In principle we should be able to drop it

src/brevitas/graph/equalize.py
Giuseppe5: Coherence with respect to what we do with the bias I suppose

src/brevitas/graph/equalize.py
```diff
@@ -245,55 +331,100 @@ def _cross_layer_equalization(
...
-    srcs_range = scale_fn(torch.cat([w.reshape(w.size(0), -1) for w in src_weights], 1))
     sinks_range = scale_fn(torch.cat([w.reshape(w.size(0), -1) for w in sink_weights], 1))
-    sinks_range += EPSILON
+    sinks_range = torch.clamp(sinks_range, EPSILON)
+
+    # Determine the srcs_range based on where we are performing activation equalization or
+    # weight equalization
+    if list_of_act_val is not None:
+        act_shape = list_of_act_val[0].shape
+        # act_val = [WeightBiasTuple(act_val) for act_val in list_of_act_val]
```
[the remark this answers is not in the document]
Giuseppe5: Nope

src/brevitas/graph/equalize.py
Giuseppe5: It was used for when activation equalization supported only one scale invariant activation at a time. Not needed anymore

src/brevitas/graph/equalize.py
Giuseppe5: The original act eq was based on FX and potentially the FunctionalToModule transformation.
I guess it's too strong of an assumption

src/brevitas/graph/equalize.py
```diff
@@ -482,73 +664,235 @@ def apply(self,
...
-        for name, module in model.named_modules():
-            if isinstance(module, (_supported_layers)):
-                hook_fn = partial(self.collect_inp_shape_hook, name=name)
-                hook = module.register_forward_pre_hook(hook_fn)
-                self.collect_inp_shape_hooks.append(hook)
-        model(inp)
-        for hook in self.collect_inp_shape_hooks:
+
+        # Compute stats along the batch dimension
+        if hasattr(inp, 'names') and 'N' in inp.names:
```
Giuseppe5, line 749: I believe this can be dropped since we check for batch_first attribute

Giuseppe5: Also, this would be present only in our QuantMHA but act_eq works on float modules

volcacius (approved):

PR #544, opened 2023-03-01
Fix and Tests for Adaround

Giuseppe5:
~~Tests might be needed before merging.~~


volcacius (approved):

volcacius:
Let's move AdaroundSte to adaround_impl.py . Let's add an AdaroundIntErrorInitSte that takes in bit_width_impl, scaling_impl and int_scaling_impl. We can then drive scaling_impl at init time by passing in from tracked_parameter_list, and int_scaling_impl by passing in from bit_width_impl. We then add it as an option to the Enum as ADAROUND_INT_ERROR_INIT, together with the existing one.

src/brevitas/core/function_wrapper/ops_ste.py
```diff
@@ -46,7 +46,7 @@ class AdaRoundSte(brevitas.jit.ScriptModule):
     that decides if "ceil" or "floor" rounding type has to be used.
     """
 
-    def __init__(self, tracked_parameter_list: List[torch.nn.Module], adaround_impl: brevitas.jit.ScriptModule) -> None:
+    def __init__(self, tracked_parameter_list: List[torch.nn.Module], adaround_impl: torch.nn.Module) -> None:
```
volcacius, line 49: You haven't removed AdaRoundSte here.

volcacius (changes requested):

volcacius (approved):

PR #555, opened 2023-03-09
Graph Eq: Transformers, LayerNorm, Better BN

Giuseppe5:


src/brevitas/graph/equalize.py
```diff
@@ -55,6 +57,8 @@
     operator.imul,
     operator.__mul__,
     operator.__imul__,
+    operator.getitem,
```
Giuseppe5, line 60: Test with torch symbolic tracer if this op is needed. In case move to _select_op category

src/brevitas/graph/equalize.py
```diff
@@ -179,13 +184,24 @@ def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], mer
...
+            return torch.tensor(1., dtype=dtype, device=device) # If module is not supported, do not perform graph equalization
+        if isinstance(module, nn.MultiheadAttention):
+            srcs[i] = StandardizedModule(module.out_proj.weight, (module.out_proj.bias))
+        src_axes[srcs[i]] = _get_output_axis(module)
+
+    for i, module in enumerate(sinks):
+        if not isinstance(module, _supported_layers):
+            return torch.tensor(1., dtype=dtype, device=device) # If module is not supported, do not perform graph equalization
+        if isinstance(module, nn.MultiheadAttention) and hasattr(sinks[i], 'in_proj_weight'): # Supports only self-attention
+            sinks[i] = StandardizedModule(module.in_proj_weight, (module.in_proj_bias, module.bias_k, module.bias_v))
```
Giuseppe5, line 201: Rename to WeightBiasTuple

Giuseppe5:
Review done:
- We now use torch symbolic tracer, and it still traces _getitem_ ops
- I tested that graph equalization fails when it is not the "self attention" case, and I can confirm it does fail. The good news is we have a way to detect when it is self-attention vs when it is not
- We could extend to equalize through any generic MHA but it would require some refactoring of _cross_equalize_region, to bridge better the information coming from the graph tracing and the information of the nn.Modules

src/brevitas/graph/equalize.py
```diff
@@ -115,16 +121,28 @@ def _get_input_axis(module: nn.Module) -> int:
             return 0
         elif module.groups == module.out_channels:
             return 1
+    elif isinstance(module, nn.LayerNorm):
+        # We assume normalization happens only along the channel dimension
```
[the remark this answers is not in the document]
Giuseppe5: LayerNorm operates starting from the last dim, and going backwards for more than 1 dim. For Transformers, the last channel is also the channel dimension, so this is correct. This assumption wouldn't hold for CNN, where the last channel is a spatial one, but LayerNorm for CNN has generally 3 dims (the channel dim, and the last 2 spatial dims).

Unless very weird configurations of LayerNorm with CNN where for some reason you normalize only the last spatial dimension, this should hold.

In this current form we wouldn't be able to equalize through layernorm applied to CNN but it's a very rare circumstance.

volcacius (approved):

tests/brevitas/graph/test_equalization.py
```diff
@@ -75,13 +76,57 @@ def test_equalization_torchvision_models(model_dict: dict, merge_bias: bool):
...
+            modules.train()
+            pre_bn_stats.append(modules.bias.data.clone())
+    for _ in range(100):
+        model(torch.randn(IN_SIZE_CONV))
+
+    model.eval()
+    post_bn_stats = []
+    for name, modules in model.named_modules():
+        if hasattr(modules, 'orig_bias'):
+            modules.train()
```
Giuseppe5, line 124: Clean-up

tests/brevitas/graph/test_equalization.py
```diff
@@ -8,36 +8,53 @@
...
+from brevitas.graph.equalize import _extract_regions
 from brevitas.graph.equalize import _is_supported_module
 
 from .equalization_fixtures import *
 
 SEED = 123456
 ATOL = 1e-3
 
+
+def equalize(model, regions, merge_bias, bias_shrinkage, scale_computation_type):
```
Giuseppe5, line 21: rename to test_equalize

src/brevitas/graph/equalize.py
```diff
@@ -229,17 +276,38 @@ def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], mer
...
         src_broadcast_size = [1] * module.weight.ndim
         src_broadcast_size[axis] = module.weight.size(axis)
         module.weight.data = module.weight.data * torch.reshape(inverse_scaling_factors, src_broadcast_size)
     for module, axis in sink_axes.items():
         src_broadcast_size = [1] * module.weight.ndim
         src_broadcast_size[axis] = module.weight.size(axis)
-        if isinstance(module, (torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d)):
-            additive_factor = module.running_mean.data * module.weight.data / torch.sqrt(module.running_var.data + module.eps)
-            module.bias.data = module.bias.data + additive_factor * (scaling_factors - 1)
+        if isinstance(module, _batch_norm):
```
volcacius, line 289: Please move this logic to a separate function

src/brevitas/graph/equalize.py
```diff
@@ -296,7 +370,10 @@ def walk_region(graph_model: GraphModule, starting_node: Node, history: Set[Node
             continue
         if _is_supported_module(graph_model, node):
             if walk_forward:
-                sinks.add(node.target)
+                module = get_module(graph_model, node.target)
+                # It is not possible equalize through LayerNorm as sink
```
volcacius, line 374: to equalize

src/brevitas/graph/equalize.py
```diff
@@ -229,17 +276,38 @@ def _cross_layer_equalization(srcs: List[nn.Module], sinks: List[nn.Module], mer
...
+            # If a new training is performed, the weight and the pre-equalized bias and weight will
+            # be learned.
+            class_name = module.__class__.__name__ + 'Equalized'
+            module.register_parameter('orig_bias', torch.nn.Parameter(module.bias.clone().detach()))
+            module.register_parameter('orig_weight', torch.nn.Parameter(module.weight.clone().detach()))
+            module.scaling_factors = scaling_factors.clone()
+            module.inverse_scaling_factor = torch.ones_like(module.orig_bias)
+
+            del module.bias
+            new_bias_lambda = lambda self: self.inverse_scaling_factor * \
```
volcacius, line 302: Can we define this as a function

src/brevitas/graph/equalize.py
```diff
@@ -22,14 +24,18 @@
 
 EPSILON = 1e-9
 
+Region = namedtuple('Region', ['sources', 'sinks'])
```
volcacius, line 27: rename to srcs

src/brevitas/graph/equalize.py
```diff
@@ -73,8 +82,29 @@
...
+
+
+WeightBiasTuple = namedtuple('WeightBiasTuple', ['weight', 'bias'], defaults=[None])
+
+
+def _equalize_bn(bn_module: nn.Module, scaling_factors: torch.Tensor):
+    class_name = bn_module.__class__.__name__ + 'Equalized'
+    bn_module.register_parameter('orig_bias', torch.nn.Parameter(bn_module.bias.clone().detach()))
+    bn_module.register_parameter('orig_weight', torch.nn.Parameter(bn_module.weight.clone().detach()))
+    bn_module.scaling_factors = scaling_factors.clone()
```
volcacius, line 95: this should be a buffer

Giuseppe5:
There's currently a bug for Batch Norm when equalizing for more than 1 iteration

Giuseppe5:
> There's currently a bug for Batch Norm when equalizing for more than 1 iteration

Now fixed

volcacius (approved):

volcacius (approved):

PR #566, opened 2023-03-28
Add support for slicing a QuantTensor

V0XNIHILI:
See title for the reason of change.

My only concern is that `QuantTensorBase` is originally a `NamedTuple`. With this change, it is however not possible anymore to use that get_item functionality of the named tuple. Ie, originally `a = QuantTensor(...); print(a[0])` would return `self.value`. Is this a problem for the codebase? Maybe it would be better to remove the `NamedTuple` base class from `QuantTensorBase` just for clarity?


Giuseppe5:
About this PR, I will discuss with @volcacius the implications it could have.
We are also planning a refactoring of QuantTensor, so maybe we can take this issue into account while working on that.

V0XNIHILI:
@Giuseppe5 @volcacius any update on this?

Giuseppe5:
I feel we can't evaluate all the implications of this change at the moment.
We might consider if/how to do it when refactoring the QuantTensor class.

Is there a specific use-case where you needed to do this?

PR #568, opened 2023-04-03
Fix mha

volcacius:


Giuseppe5 (approved):

Giuseppe5:
Typing of Query, Key, and Value needs to be updated to support the QuantTensor case

PR #570, opened 2023-04-03
Ci update

Giuseppe5:


Giuseppe5:
It seems that switching to _logical_ from _auto_ brings no benefit (2 physical and logical core)

PR #573, opened 2023-04-12
Learned round

Giuseppe5:


src/brevitas/core/function_wrapper/learned_round.py
```diff
@@ -0,0 +1,95 @@
...
+            self,
+            tracked_parameter_list: List[torch.nn.Module],
+            learnedround_impl: torch.nn.Module,
+            learnedround_init: Optional[torch.Tensor] = None) -> None:
+        super(LearnedRoundSte, self).__init__()
+        if len(tracked_parameter_list) > 1:
+            raise RuntimeError('LearnedRound does not support shared quantizers')
+        self.learnedround_impl = learnedround_impl
+        if learnedround_init == None:
+            self.value = torch.nn.Parameter(torch.full(tracked_parameter_list[0].shape, 0.))
```
volcacius, line 71: let's move this to a value function called learned_round_init where the enum is solved too that depends on tracked_parameter_list, that also does the check for shared quantizers. This way in the init here we just need to take in learned_round_init.

volcacius (approved):

PR #574, opened 2023-04-14
Fix (graph_quant): split sign and precision preserving modules

Giuseppe5:


src/brevitas/graph/quantize_impl.py
```diff
@@ -104,7 +107,7 @@ def are_inputs_quantized_and_aligned(model, node, quantized_modules_list, quant_
             inp_module = get_module(model, inp_node.target)
             if isinstance(inp_module, tuple(quant_act_map.keys())):
                 quantized_modules_list.append(None)
-            elif isinstance(inp_module, tuple(SIGN_PRESERVING_MODULES)):
+            elif isinstance(inp_module, tuple(PRECISION_PRESERVING_MODULES)):
                 are_inputs_quantized_and_aligned(
                     model, inp_node, quantized_modules_list, quant_act_map, same_sign)
```
volcacius, line 112: assuming PRECISION_PRESERVING_MODULES might not preserve the sign (in the future), is passing same_sign here correct, or should it be same_sign only if inp_module is in SIGN_PRESERVING_MODULES too?

Giuseppe5: Fixed now

volcacius (approved):

PR #582, opened 2023-04-24
Feat (graph_eq): improvements to graph equalization

Giuseppe5:
Moved to a stateless implementation of walk_region, which is not split into two functions:
- find_sources (to be named find_srcs)
- find_sinks

Clean-up of minor other issues

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -173,8 +173,8 @@ def _combine_weights_bias(
         return weight.data
 
     bias = bias.data
-    weight = deepcopy(weight).view(weight.shape[0], -1)
-    bias = deepcopy(bias).view(-1, 1)
+    weight = weight.reshape(weight.shape[0], -1)
+    bias = bias.reshape(-1, 1)
 
     weight = torch.where(torch.abs(weight) < EPSILON, torch.tensor(EPSILON).type_as(weight), weight)
     factor = torch.abs(bias) / torch.abs(weight)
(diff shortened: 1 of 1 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -350,57 +350,75 @@ def _is_reshaping_op(node: Node) -> bool:
...
-            if walk_forward:
-                walk_region(graph_model, node, history, srcs, sinks, walk_forward=True)
-            else:
-                walk_region(graph_model, node, history, srcs, sinks, walk_forward=True)
-                walk_region(graph_model, node, history, srcs, sinks, walk_forward=False)
+            state['srcs'].add(node.target)
+            find_sinks(graph_model, node, state)
+        elif _is_scale_invariant_module(
+                graph_model, node) or _is_scale_invariant_function(node) or _is_reshaping_op(node):
+            find_srcs(graph_model, node, state)
```
Giuseppe5, line 368: Missing state =

src/brevitas/graph/equalize.py
```diff
@@ -350,57 +350,76 @@ def _is_reshaping_op(node: Node) -> bool:
...
 
 
 def _extract_regions(graph_model: GraphModule) -> Set[Tuple[str]]:
     regions = set()
     for node in graph_model.graph.nodes:
         if _is_supported_module(graph_model, node):
-            srcs, sinks = {node.target}, set()
-            walk_region(graph_model, node, set(), srcs, sinks, walk_forward=True)
-            if sinks:
+            state = dict()
```
volcacius, line 414: Can we have a dataclass for the state rather than a dict?

src/brevitas/graph/equalize.py
```diff
@@ -350,57 +350,76 @@ def _is_reshaping_op(node: Node) -> bool:
...
+            state['srcs'] = {node.target}
+            state['sinks'] = set()
+            state['history'] = set()
+            state = find_sinks(graph_model, node, state)
```
volcacius, line 418: do we need to reassign the state variable?

src/brevitas/graph/equalize.py
```diff
@@ -173,8 +173,8 @@ def _combine_weights_bias(
         return weight.data
 
     bias = bias.data
-    weight = deepcopy(weight).view(weight.shape[0], -1)
-    bias = deepcopy(bias).view(-1, 1)
+    weight = weight.reshape(weight.shape[0], -1)
```
volcacius, line 176: can we do a view instead of a reshape?

src/brevitas/graph/equalize.py
Giuseppe5: No I did it mostly for clarity, to make clear that it gets changed and updated. I would rather keep it if it's not too ugly or non-pythonic

src/brevitas/graph/equalize.py
Giuseppe5: No for some models/configurations it fails with view

src/brevitas/graph/equalize.py
volcacius, line 1: Can we move to dataclasses for things that are currently namedtuples?

volcacius (approved):

PR #585, opened 2023-04-28
Examples (espcn): Adding super resolution example

i-colbert:
Adding single-image super resolution Brevitas examples training a FINN-friendly variant of the Efficient Sub-Pixel Convolution Network (ESPCN) to upsample images by 2x using the BSD300 dataset. Including accumulator-aware quantization (A2Q) usage example with floating-point and vanilla per-channel weight quantization benchmarks.


src/brevitas_examples/super_resolution/models/espcn.py
```diff
@@ -0,0 +1,207 @@
...
+            in_channels=num_channels,
+            out_channels=64,
+            kernel_size=5,
+            stride=1,
+            padding=2,
+            bias=True,
+            input_bit_width=IO_BIT_WIDTH,
+            input_quant=CommonUintActQuant,
+            weight_bit_width=IO_BIT_WIDTH,
+            weight_accumulator_bit_width=32,
```
volcacius, line 110: Let's capture this in a constant.

PR #586, opened 2023-05-01
Fix (graph_quant): correct lambda evaluation

Giuseppe5:


volcacius (approved):

volcacius:
Add a test for this please, something simple but that checks that it doesn't break.

volcacius (approved):

PR #595, opened 2023-05-16
Feat (ptq): new benchmark results with GPTQ

Giuseppe5:


volcacius:
The reported git commit is wrong, can we fix that plase.

volcacius (approved):

PR #604, opened 2023-05-22
Examples (imagenet/ptq): add BN calibration support

volcacius:


src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
```diff
@@ -128,6 +128,7 @@
 add_bool_arg(parser, 'gptq', default=True, help='GPTQ (default: enabled)')
 add_bool_arg(
     parser, 'gptq-act-order', default=False, help='GPTQ Act order heuristic (default: disabled)')
+add_bool_arg(parser, 'calibrate-bn', default=False, help='Calibrate BN (default: disabled)')
```
Giuseppe5, line 131: We might want to add it to the config variable (which is already very long to be fair)

PR #606, opened 2023-05-31
Fix (bias_corr): correct use of float output for bias_correction

Giuseppe5:


src/brevitas/graph/calibrate.py
```diff
@@ -283,19 +260,18 @@ def register_hook_to_wbiol(self, model, hooks):
...
-        self.register_collect_float_mean_hook(module, name)
-        module.forward(*inp)  # Required to avoid infinite recursion
-        self.float_mean_hooks_cleanup()
+        out_float = module.forward(*inp)  # Required to avoid infinite recursion
+        self.collect_float_mean(module, out_float, name)
         self.enable_act_quantization(module, is_training=False)
         self.enable_param_quantization(module, is_training=False)
-        self.register_correct_bias_hook(module, name)
-        out = module.forward(*inp)  # Required to avoid infinite recursion
+        out_quant = module.forward(*inp)  # Required to avoid infinite recursion
```
volcacius, line 274: We need to disable output_quant while out_quant is being computed or it's gonna do BC on post quant values, instead of pre quant, which means that then we can't really merge into bias.

PR #608, opened 2023-06-02
Feat (gptq): support for parallel layer execution

Giuseppe5:
- Possibility to apply gptq to group of layers that are executed in parallel.
- Support for Conv 1d/2d and ConvTranspose 1d/2d

### DIFF
--- a/src/brevitas/graph/gptq.py
+++ b/src/brevitas/graph/gptq.py
@@ -125,7 +125,7 @@ class gptq_mode:
                         act_order=self.act_order,
                         parallel_layers=parallel_layers)
                     hook_fn = partial(gptq.update_batch, current_layer=self.current_layer)
-                    self.hook_dict[name] = module.register_forward_pre_hook(hook_fn)
+                    self.hook_dict[name] = module.register_forward_hook(hook_fn)
                     self.gptq_layers[name] = gptq
         if not self.use_quant_activations:
             self.disable_quant_inference.disable_act_quantization(
@@ -200,11 +200,13 @@ class GPTQ():
         self.blocksize = math.ceil(self.columns / self.num_blocks)
 
         # Initialize Hessian matrix and counter. We need it in float32 to compute the inverse
-        self.H = torch.zeros((self.groups, self.columns, self.columns), device=dev)
+        self.H = torch.zeros((self.groups, self.columns, self.columns),
+                             device=dev,
+                             dtype=torch.float32)
         self.nsamples = 0
         self.parallel_layers = parallel_layers
 
-    def update_batch(self, module, input, current_layer):
+    def update_batch(self, module, input, output, current_layer):
         # Update reference to current layer
         current_layer.layer_names.add(self.name)
 
@@ -215,16 +217,20 @@ class GPTQ():
             inp = inp.value
 
         # If input is unbatched, add batch_size = 1
-        if len(inp.shape) == 2:
+        if len(inp.shape) == 1:
+            warnings.warn("Found unbatched input, adding batch dimension equal to 1")
             inp = inp.unsqueeze(0)
 
         # Define batch size before re-organizing the input
+        if hasattr(inp, 'names') and 'N' in inp.names:
+            batch_dim = inp.names.index('N')
+            inp = inp.transpose(0, batch_dim)
         batch_size = inp.shape[0]
 
         # Preprocess the input to compute the Hessian
         if isinstance(self.layer, qnn.QuantLinear):
-            if len(inp.shape) == 3:
-                inp = inp.reshape((-1, inp.shape[-1]))
+            if len(inp.shape) > 2:
+                inp = inp.reshape((-1, torch.sum(inp.shape[2:])))
             inp = inp.t()
             # For QuantLinear layer, groups will be 1
             inp_processed = inp.unsqueeze(0)
@@ -248,7 +254,7 @@ class GPTQ():
             # Preprocess input by group
             for i, inp in enumerate(inp_by_group):
                 inp = unfold(inp)
-                inp = inp.permute([1, 0])
+                inp = inp.transpose([1, 0])
                 inp = inp.flatten(1)
                 inp_processed.append(inp)
             inp_processed = torch.stack(inp_processed)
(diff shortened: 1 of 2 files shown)


src/brevitas/core/scaling/runtime.py
```diff
@@ -88,6 +95,8 @@ def __init__(
             scaling_stats_input_view_shape_impl: Module,
             restrict_scaling_impl: Module,
             scaling_shape: Tuple[int, ...],
+            dtype: Optional[torch.dtype] = None,
+            device: Optional[torch.device] = None,
```
volcacius, line 99: Let's add new args always at the bottom, after scaling_min_val

src/brevitas/core/scaling/runtime.py
```diff
@@ -114,13 +125,19 @@ def forward(self, x: torch.Tensor):
...
-    def __init__(self, scaling_shape, shift_scale):
+    def __init__(
+            self,
+            scaling_shape,
+            shift_scale,
+            dtype: Optional[torch.dtype],
+            device: Optional[torch.device]):
         super(_AffineRescaling, self).__init__()
-        self.affine_weight = Parameter(torch.ones(scaling_shape))
+        self.affine_weight = Parameter(torch.ones(scaling_shape, dtype=dtype, device=device))
```
volcacius, line 135: What happens when they are None?

src/brevitas/quant/solver/weight.py
```diff
@@ -101,7 +102,8 @@ class WeightQuantSolver(SolveWeightScalingStatsInputDimsFromModule,
                         SolveParameterScalingInitFromEnum,
                         SolveParameterScalingShape,
                         SolveWeightScalingPerOutputChannelShapeFromModule,
-                        SolveWeightTensorQuantFromEnum):
+                        SolveWeightTensorQuantFromEnum,
+                        SolveDtypeDeviceFromTrackedParameterList):
```
volcacius, line 106: Should be done for bias too, for when we have a standalone scale.

src/brevitas/quant/solver/common.py
```diff
@@ -150,13 +150,6 @@ def learned_round_init(tracked_parameter_list):
         return torch.full(tracked_parameter_list[0].shape, 0.)
 
 
-class SolveTensorQuantFloatToIntImplFromEnum(ExtendedInjector):
```
volcacius, line 153: What happened here? Why is it being removed?

src/brevitas/graph/gptq.py
```diff
@@ -331,24 +395,33 @@ def single_layer_update(self, percdamp=.01):
             else:
                 perm = permutation_list[0]
                 weight[:, perm[i2:]] -= error_block.matmul(h_inv[0, i1:i2, i2:])
+            weight.to(dtype)
```
volcacius, line 398: What's the reasoning for this casting? Can you add a comment?

src/brevitas/graph/gptq.py
```diff
@@ -331,24 +395,33 @@ def single_layer_update(self, percdamp=.01):
...
+                    quant_weight = quant_weight[:, permutation_list[0]]
+                else:
+                    for ii, perm in enumerate(permutation_list):
+                        quant_weight[ii, :] = quant_weight[ii, perm]
+
+            quant_weight_block = quant_weight[:, i1:i2]
+            q = quant_weight_block[:, i]
+        else:
+            pass
+        # else:
```
volcacius, line 423: Can you raise a NotImplemented?

src/brevitas/graph/gptq.py
```diff
@@ -204,7 +262,7 @@ def update_batch(self, module, input, out, current_layer):
             # Preprocess input by group
             for i, inp in enumerate(inp_by_group):
                 inp = unfold(inp)
-                inp = inp.permute([1, 0, 2])
+                inp = inp.permute([1, 0])
```
volcacius, line 265: What happened here?

src/brevitas/graph/gptq.py
```diff
@@ -64,40 +73,69 @@ def __init__(
...
+                return False
         elif isinstance(module, qnn.QuantLinear):
             return True
         else:
             return False
 
     def __enter__(self):
+        # If the user specifies which layers to execute in parallel, we follow that structure
+        # Otherwise each supported layer will be executed standalone
+        if self.group_of_parallel_layers is not None:
```
volcacius, line 100: Can we do it differently, where we allow to pass in only some groups of layers, and we can infer that the others have to run by themselves?

src/brevitas/quant/solver/common.py
Giuseppe5: If you check, it is defined twice. I guess something happened during a rebase/merge

src/brevitas/graph/gptq.py
Giuseppe5: This should not be here since it will addressed in another PR entirely, apologies

src/brevitas/graph/gptq.py
Giuseppe5: We are just interested in permuting the first two dimensions, leaving all the others unchanged. The `2` was redundant.

src/brevitas/graph/gptq.py
Giuseppe5: Yes I can work on that

volcacius (commented):
A few notes:
- Put the device and dtype changes on a separate PR we can rebase this one onto.
- Device and dtype for learned bit width are missing.
- Device and dtype for QuantRNN and QuantLSTM are missing.

src/brevitas/graph/gptq.py
```diff
@@ -64,40 +72,68 @@ def __init__(
...
+                return False
         elif isinstance(module, qnn.QuantLinear):
             return True
         else:
             return False
 
     def __enter__(self):
+        # The user can specify on which layers to apply gptq in parallel.
+        # All the others will be executed sequentially
+        list_of_layers = {
```
volcacius, line 95: does this need to be a dict? It's called list and the key is not used below

volcacius: Nvm I see where it's used. Let's call it dict

src/brevitas/graph/gptq.py
```diff
@@ -162,13 +199,14 @@ def __init__(self, layer, name, num_blocks, act_order) -> None:
         # Define how many columns to update in each mini-block
         self.blocksize = math.ceil(self.columns / self.num_blocks)
 
-        # Initialize Hessian matrix and counter
-        self.H = torch.zeros((self.groups, self.columns, self.columns), device=dev, dtype=dtype)
+        # Initialize Hessian matrix and counter. We need it in float32 to compute the inverse
+        self.H = torch.zeros((self.groups, self.columns, self.columns), device=dev)
```
volcacius, line 203: set dtype explicitly to float32 instead of relying on default

volcacius (approved):

PR #612, opened 2023-06-06
Examples (mobilenetv1): exposing new args to quant_mobilenet_v1

i-colbert:
exposing control of weight_quant and separating act/weight bit-widths.


src/brevitas_examples/imagenet_classification/models/mobilenetv1.py
```diff
@@ -137,10 +148,11 @@ def __init__(
             in_channels=in_channels,
             out_channels=init_block_channels,
             kernel_size=3,
-            stride=2,
+            stride=first_layer_stride,
             weight_bit_width=FIRST_LAYER_BIT_WIDTH,
```
volcacius, line 152: can you move this to an argument too?

volcacius (approved):
LGTM

PR #615, opened 2023-06-06
Examples (resnet18): exposing more args to quant_resnet18

i-colbert:
Exposing control over weight quants.

### DIFF
--- a/src/brevitas_examples/bnn_pynq/models/resnet.py
+++ b/src/brevitas_examples/bnn_pynq/models/resnet.py
@@ -6,9 +6,9 @@ from typing import List
 from torch import Tensor
 import torch.nn as nn
 
-from brevitas.inject.defaults import Int8WeightPerTensorFloat
 import brevitas.nn as qnn
 from brevitas.quant import TruncTo8bit
+from brevitas.quant import Int8WeightPerChannelFloat
 from brevitas.quant_tensor import QuantTensor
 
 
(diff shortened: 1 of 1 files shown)


src/brevitas_examples/bnn_pynq/models/resnet.py
```diff
@@ -39,10 +45,11 @@ def __init__(
             in_planes,
             planes,
             stride=1,
+            bias=False,
             shared_quant_act=None,
-            weight_bit_width=8,
             act_bit_width=8,
-            bias=False):
+            weight_bit_width=8,
+            weight_quant=Int8WeightPerTensorFloat):
```
volcacius, line 52: You can set as default Int8WeightPerChannelFloat and get rid of the weight_scaling_per_output_channel=True set explicitly

src/brevitas_examples/bnn_pynq/models/resnet.py
```diff
@@ -1,28 +1,34 @@
...
 def make_quant_conv2d(
-        in_channels, out_channels, kernel_size, weight_bit_width, stride=1, padding=0, bias=False):
+        in_channels,
+        out_channels,
+        kernel_size,
+        weight_bit_width,
+        stride=1,
+        padding=0,
+        bias=False,
+        weight_quant=Int8WeightPerTensorFloat):
```
volcacius, line 23: I would remove this default since it's defined in the caller already, and for quantizers I prefer to have defaults set only once.

volcacius (approved):
LGTM

PR #629, opened 2023-06-15
Feat (nn): set dim names in QuantMHA Linear

volcacius:


Giuseppe5:
Need to be rebased after merging #647

PR #630, opened 2023-06-15
Feat (quantize): support for name/node in lambdas

Giuseppe5:


src/brevitas_examples/imagenet_classification/ptq/ptq_common.py
```diff
@@ -67,7 +67,7 @@ def quantize_model(
     maps = [deepcopy(quant_map) for quant_map in LAYER_MAP[backend]]
 
     def bit_width_fn(module, other_bit_width):
-        if backend != 'layerwise':
+        if backend != 'layerwise' or layerwise_first_last_bit_width is None:
```
Giuseppe5, line 70: This should not be here

Giuseppe5:
Add name in quantization lambdas

PR #642, opened 2023-06-27
Activation eq extension

Giuseppe5:


src/brevitas/graph/equalize.py
```diff
@@ -745,30 +767,21 @@ def apply(self, alpha):
...
+        elif not use_inp:
+            inp = args[-1]
+        elif len(kwargs) > 0:
+            inp = kwargs[kwarg_name]
         # Extra check for batch_dim
         if hasattr(inp, 'names') and 'N' in inp.names:
             batch_dim = inp.names.index('N')
             inp = inp.transpose(0, batch_dim)
 
-        if len(inp) > 1:
```
Giuseppe5, line 762: This has to be restored and improved with the presence of kwargs

src/brevitas/graph/equalize.py
```diff
@@ -723,11 +719,20 @@ def setup(self):
...
+                batch_dim=batch_dim,
+                kwarg_name=kwarg_name,
+                use_inp=True)
+            new_instance = KwargsForwardHook(region, hook_fn)
+            ModuleInstanceToModuleInstance(region, new_instance).apply(self.model)
+            self.hooks.append(new_instance)
 
     def apply(self, alpha):
         scale_factors = []
+        self.remove_hooks()
```
Giuseppe5, line 735: Since we perform it here, we need to remove it from __exit__

src/brevitas/nn/equalized_layer.py
```diff
@@ -10,8 +10,20 @@ def __init__(self, scale_module, layer) -> None:
         self.scale = scale_module
         self.layer = layer
 
-    def forward(self, x, *args, **kwargs):
+    def forward(self, *args, **kwargs):
         args = list(args)
+
+        if len(args) > 0:
```
Giuseppe5, line 16: We should probably standardize this better, in case of a kwargs with different names maybe.

volcacius (approved):

PR #645, opened 2023-06-30
Feat (llm): QuantizableBert

Giuseppe5:


volcacius:
Split the attention mask handling part and we can merge that into llm.

volcacius:
We should move these attention modules under `brevitas.nn` .

Giuseppe5:
Closed in favor of #961

PR #685, opened 2023-07-24
Fix single linear test

jinchen62:
Got the following error without the changes.

> python exception: Failure while executing pass pipeline:
error: unknown: unsupported by backend contract: module initializers
note: unknown: see current operation: "torch.initialize.global_slots"(%4, %5, %6) {slotSymNames = [@_tensor_constant1, @_tensor_constant2, @_tensor_constant3]} : (!torch.tensor<[256,128],ui8>, !torch.tensor<[256,1,1],f32>, !torch.tensor<[256,1,1],f32>) -> ()
note: unknown: this is likely due to InlineGlobalSlots being unable to inline a global slot

Tested with the latest torch-mlir, no other changes required.

### DIFF
--- a/src/brevitas_examples/llm/test_linear_mlir_export.py
+++ b/src/brevitas_examples/llm/test_linear_mlir_export.py
@@ -1,4 +1,5 @@
 import argparse
+from typing import List, Tuple
 
 import torch
 from torch import nn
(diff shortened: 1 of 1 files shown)


jinchen62:
@volcacius

jinchen62:
The fix also applies to src/brevitas_examples/llm/main.py.

Issue #449, opened 2022-11-11
Shape inference support for qonnx symbolic operators


volcacius:
Setting shape inference is currently broken in upstream PyTorch for custom ONNX operators https://github.com/pytorch/pytorch/issues/81693 . This impact qonnx export, andin  some use cases lack of shape propagations means that certain other operators (e.g. globalmaxpool) that require correct shape inference are not exported correctly.

lix19937:
You can ref https://github.com/Xilinx/brevitas/blob/master/src/brevitas/export/onnx/standard/function.py

PR #692, opened 2023-08-04
Fixes ConvNd usability and shape inference for qonnx export on Torch 2.0.1

bwintermann:
As mentioned in issue #449 when using Torch>=2.0.1, sometimes shape inference doesn't work properly when exporting to QONNX. In some cases it only produces warnings, but for some cases with Conv2D layers specifically the export won't work at all. This would fix both the warnings and errors, atleast regarding the ConvNd layer types. I tested the export on multiple example networks afterwards.

### DIFF
--- a/src/brevitas/export/onnx/qonnx/manager.py
+++ b/src/brevitas/export/onnx/qonnx/manager.py
@@ -12,6 +12,8 @@ from brevitas.export.manager import _set_recurrent_layer_export_handler
 from brevitas.export.manager import _set_recurrent_layer_export_mode
 from brevitas.export.onnx.debug import DebugMarkerFunction
 from brevitas.export.onnx.manager import ONNXBaseManager
+from brevitas.export.onnx.finn.function.parameter import QuantizedConvNdFn
+from brevitas.export.onnx.finn.handler.parameter import FINNQuantConvNdHandler
 
 from .function import BrevitasBinaryQuantFn
 from .function import BrevitasQuantFn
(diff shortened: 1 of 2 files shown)


volcacius:
Hey @bwintermann , thanks for the PR, this is definitely something that needs fixing. Do you have an example for why you are adding the FINNQuantConvNdHandler to the handlers of QONNX? I don't think it's necessary but I would like to understand your use case. Thanks

bwintermann:
With the big Torch update to 2.0 the export of certain layers in the QONNX format doesn't work anymore, seemingly due to missing shape information, which, it looks like, could be inferred before, but is now required. Bringing the FINN handler and function into scope was the easiest way to provide the required information. However it is completely possible that I've misunderstood the usage of the export manager and that this is not the proper way to fix it. I would of course also say that importing the handler and function from another export subdirectory is not the cleanest solution, so that if this were to persist, maybe the handlers and functions could be moved into a shared location for both the FINN- and the QONNX-export.

src/brevitas/export/onnx/qonnx/manager.py
```diff
@@ -42,14 +44,16 @@ class QONNXManager(ONNXBaseManager):
         BrevitasDecoupledWeightQuantProxyHandler,
         BrevitasDecoupledWeightQuantWithInputProxyHandler,
         BrevitasTruncQuantProxyHandler,
-        BrevitasQuantLSTMLayerHandler]
+        BrevitasQuantLSTMLayerHandler,
+        FINNQuantConvNdHandler]
```
volcacius, line 48: Remove this (FINN ONNX is being deprecated).

src/brevitas/export/onnx/qonnx/manager.py
```diff
@@ -42,14 +44,16 @@ class QONNXManager(ONNXBaseManager):
...
 
     custom_fns = [
         DebugMarkerFunction,
         BrevitasQuantFn,
         BrevitasBinaryQuantFn,
         BrevitasTruncFn,
-        BrevitasQuantLSTMCellFn]
+        BrevitasQuantLSTMCellFn,
+        QuantizedConvNdFn]
```
volcacius, line 56: Remove this (FINN ONNX is being deprecated).

Giuseppe5:
Included in #760, closing this

PR #693, opened 2023-08-17
Rename brevitas quant custom op

jinchen62:
I have tested the op would be `torch.operator "quant.matmul_rhs_group_quant"`.


jinchen62:
Pls review, thanks! @Giuseppe5 @volcacius @ harish

Giuseppe5:
Would you be able to run 
`pre-commit run --all-files`

And add also the changes that cause the pre-commit check to fail?
It should be changing something in one of the files you are already modifying.

Thanks!

PR #704, opened 2023-09-12
Fix (test): ignore hypothesis health check

Giuseppe5:


Giuseppe5:
Moving to draft as it seems it has been fixed by hypothesis

PR #716, opened 2023-10-03
Fix errors in test

Giuseppe5:


src/brevitas/utils/quant_utils.py
```diff
@@ -7,6 +7,13 @@
 from brevitas.inject.enum import FloatToIntImplType
 
 
+def _is_proxy_in_export_mode(model):
```
volcacius, line 10: doesn't check whether it's a proxy

src/brevitas/nn/quant_rnn.py
```diff
@@ -675,9 +674,8 @@ def forward(self, inp, hidden_state, cell_state):
             self.output_gate_params, quant_input)
         if self.cifg:
             # Avoid dealing with None and set it the same as the forget one
-            quant_weight_if = quant_weight_ii
-            quant_weight_hf = quant_weight_hi
-            quant_bias_forget = quant_bias_input
+            quant_weight_if, quant_weight_hf, quant_bias_forget = self.gate_params_fwd(
```
volcacius, line 677: Increase min version to 1.10 and deal with None since it's supported in TS.

volcacius (approved):

PR #751, opened 2023-11-07
Fix (llm): add checks for group dimensions

Giuseppe5:


src/brevitas_examples/llm/llm_quant/quantizers.py
```diff
@@ -32,22 +32,22 @@ class WeightSymmetricGroupQuantMixin(ExtendedInjector):
     @value
     def expanded_scaling_shape(module, block_size):
         if isinstance(module, nn.Conv2d):
-            return module.weight.size(0), module.weight.size(1) // block_size, block_size, module.weight.size(2), module.weight.size(3)
+            return module.weight.size(0), (module.weight.size(1) + block_size - 1) // block_size, block_size, module.weight.size(2), module.weight.size(3)
```
volcacius, line 35: Let's just put an assert to check that the block size divides size(1) of the weights.

PR #754, opened 2023-11-13
changes class_implementation to init_class in gpxq_mode

fabianandresgrob:
This change cleans up the implementations for GPFQ and GPTQ. It also allows for easier changes in the future.

### DIFF
--- a/src/brevitas/graph/gptq.py
+++ b/src/brevitas/graph/gptq.py
@@ -84,12 +84,12 @@ class gptq_mode(gpxq_mode):
                 return out
 
     def initialize_module_optimizer(
-            self, layer, name, act_order, parallel_layers, create_weight_orig):
+            self, layer, name, act_order, len_parallel_layers, create_weight_orig):
         return GPTQ(
             layer=layer,
             name=name,
             act_order=act_order,
-            parallel_layers=parallel_layers,
+            len_parallel_layers=len_parallel_layers,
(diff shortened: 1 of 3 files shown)


src/brevitas/graph/gptq.py
```diff
@@ -104,15 +111,22 @@ class GPTQ(GPxQ):
...
+            create_weight_orig=True,
+            num_blocks=100) -> None:
         super().__init__(layer, name, act_order, parallel_layers, create_weight_orig)
 
         dev = self.layer.weight.device
 
         # Define how many columns to update in each mini-block
-        self.blocksize = math.ceil(self.columns / GPTQ.num_blocks)
+        self.blocksize = math.ceil(self.columns / num_blocks)
+        self.num_blocks = num_blocks
```
Giuseppe5, line 129: Do we need num_blocks after init?
If no, let's not store it

src/brevitas/graph/gpxq.py
```diff
@@ -98,7 +98,7 @@ def __enter__(self):
 
                 # Attach hooks for GPTQ
                 if self._is_module_supported(module):
-                    gpxq = self.class_implementation(
+                    gpxq = self.init_class(
```
Giuseppe5, line 101: let's call the function `initialize_module_optimizer`
And then this line would be:
gpxq_module_optimizer = self.initialize_module_optimizer

PR #760, opened 2023-11-15
Setup: CI tests against pytorch 2.x

Giuseppe5:


volcacius:
Check that the ONNX debug flow is indeed still working.

volcacius (approved):
LGTM

PR #785, opened 2023-12-13
Feat: improve Brevitas compatibility with `torch.compile`

nickfraser:
Currently crashes occasionally occur when using `BREVITAS_JIT=1` & `torch.compile`, ideally these should interoperate without crashing until `TorchScript` is deprecated by upstream PyTorch.

### DIFF
--- a/src/brevitas_examples/bnn_pynq/bnn_pynq_train.py
+++ b/src/brevitas_examples/bnn_pynq/bnn_pynq_train.py
@@ -69,6 +69,7 @@ def parse_args(args):
     parser.add_argument("--network", default="LFC_1W1A", type=str, help="neural network")
     parser.add_argument("--pretrained", action='store_true', help="Load pretrained model")
     parser.add_argument("--strict", action='store_true', help="Strict state dictionary loading")
+    parser.add_argument("--compile", action='store_true', help="Compile model with `torch.compile` (PyTorch version >=2 only)")
     parser.add_argument(
         "--state_dict_to_pth",
         action='store_true',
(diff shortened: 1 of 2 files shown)


Giuseppe5:
Current findings after applying `compile` to quantized models (e.g., ready for inference):
- Current structure of QuantTensor is not amenable to compile
- It seems possible to get fullgraph compilation with some modifications to the codebase (e.g., removing global variables, adding zero_zero_point flag to avoid data-dependent conditional flow, `allow_in_graph` for QuantTensor) for _**layerwise**_ only quantization. However it seems a bit brittle.
- Inheriting from another NamedTuple (IntQuantTensorBase or FloatQuantTensorBase) breaks the `is_namedtuple` check within PyTorch, which could be patched.
- Defining a `__new__`  method (as we do for type checking) is also not supported. It seems that at compile time the type instance is passed multiple time within the call to new.
- The above problems are currently being investigate (hopefully) here :
- Even if that were solved, it seems that compile does not like torch.bool dtypes for constant values (e.g., zero_zero_point or training or signed) and it rather prefers simple `bool`. torch.bool dtype causes the NamedTuple to decay to tuple, e.g. when adding 2 QT.
- Calling `.item()` to get a bool from a torch.bool is kind of supported with certain compile flags

After this, my suggestions for this PR:
- It might be worth getting rid of training vs inference behavior for QT to avoid "data dependent" checks
- It might be worth dropping switching to support `compile` while simultaneously dropping support to PyTorch less than 2.0
- It might be worth considering the switch to Tensor subclass which might be more amenable with respect to compile
- Or we could have a compile supports in `export_mode`, where we never instantiate  QuantTensor at proxy level, and we only deal with dequantized values. Doing this in export mode ensures that we don't need to propagate QuantTensors since everything is cached (e.g., metadata for bias quantization)

PR #797, opened 2024-01-19
Feat (a2q+): improving accumulator-aware weight quantization

i-colbert:


Giuseppe5:
@i-colbert Would you mind taking a look at the failing tests? It seems that in some instances where the JIT is enabled (but not all) something fails.

src/brevitas/quant/base.py
```diff
@@ -400,6 +403,20 @@ def accumulator_bit_width_impl(accumulator_bit_width):
     float_to_int_impl = RoundToZeroSte  # required to ensure no upwards rounding violates constraints
 
 
+class AccumulatorAwareZeroCenterWeightQuant(AccumulatorAwareWeightQuant):
+    """Experimental improved accumulator-aware weight quantized based on: `A2Q+: Improving
```
Giuseppe5, line 407: Let's change to experimental zerocenter etc.etc.

Giuseppe5 (approved):

Issue #795, opened 2024-01-16
Implement QuantConv3d and QuantConvTranspose3d


capnramses:
This issue extends the family of `QuantConv` functionality (see `QuantConv1d`, `QuantConv2d`, etc.) to include:

* QuantConv3d
* QuantConvTranspose3d

PR #805, opened 2024-01-23
Feat: Add QuantConv3d and QuantConv3dTranspose

costigt-dev:
## Issue
Implements https://github.com/Xilinx/brevitas/issues/795

## Details

- Extends QuantConv and QuantConvTranspose to have 3d variant.
- Adds test for QuantConv3d
- Updates output_channel_dim for QuantConv1d to no longer throw RuntimeError
- Also removes conv**X**d_same_zero_pad_stride as it is only invoked by an invalid combination of padding/stride parameters.
- Adds tests for MergeBatchNorm and AvgPoolToQuantDepthwiseConv


src/brevitas/nn/quant_conv.py
```diff
@@ -211,3 +213,113 @@ def max_acc_bit_width(self, input_bit_width: Tensor, weight_bit_width: Tensor):
...
+            x = F.pad(
+                x,
+                [
+                    pad_w // 2,
+                    pad_w - pad_w // 2,
+                    pad_h // 2,
+                    pad_h - pad_h // 2,
+                    pad_d // 2,
+                    pad_d - pad_d // 2])
+        out = F.conv2d(x, weight, bias, self.stride, 0, self.dilation, self.groups)
```
nickfraser, line 304: I think this should be `F.conv3d` here.

src/brevitas/nn/quant_conv.py
```diff
@@ -175,8 +176,9 @@ def per_elem_ops(self):
     @property
     def output_channel_dim(self):
         if self.transposed:
-            raise RuntimeError("Transposed kernels not supported")
```
nickfraser, line 178: @Giuseppe5, is there a reason why this RuntimeError was raised? Or is Tim's fix for this OK?

src/brevitas/graph/target/flexml.py
```diff
@@ -45,6 +45,12 @@
             'weight_quant': Int8WeightPerTensorFixedPoint,
             'bias_quant': Int16Bias,
             'return_quant_tensor': True}),
+    nn.Conv3d: (
+        qnn.QuantConv3d,
+        {
+            'weight_quant': Int8WeightPerTensorFixedPoint,
+            'bias_quant': Int16Bias,
+            'return_quant_tensor': True}),
```
nickfraser, line 53: @Giuseppe5, do we need Conv3d for the flexml target?

nickfraser:
General comments / questions:

I think our graph transformations are not exhaustively tested, so with that in mind I'd like to know:
 - Are your changes to graph transformations tested? (Even just locally, without a unit test...)
 - Specifically, I'm curious about the following:
   - `src/brevitas/graph/fixed_point.py:MergeBatchNorm` does a 3d batch norm get merged into Conv3d correctly?
   - `src/brevitas/graph/<gptq|gpxq>.py` - do the weight manipulations still make sense in the 3d case? (@Giuseppe5, your comments here are appreciated)
   - `src/brevitas/graph/per_input.py:replace_modules` does the conversion from `AvgPool3d` to `QuantConv3d` work and is it correct (does it produce the correct output)?

nickfraser:
In your notes:
> Updates output_channel_dim for QuantConv2d to no longer throw RuntimeError

I think you updated QuantConv1d ;)

capnramses:
Just checking if this comment changes the issue _status_: /review

costigt-dev:
> General comments / questions:
> 
> I think our graph transformations are not exhaustively tested, so with that in mind I'd like to know:
> 
> * Are your changes to graph transformations tested? (Even just locally, without a unit test...)
> * Specifically, I'm curious about the following:
>   
>   * `src/brevitas/graph/fixed_point.py:MergeBatchNorm` does a 3d batch norm get merged into Conv3d correctly?
>   * `src/brevitas/graph/<gptq|gpxq>.py` - do the weight manipulations still make sense in the 3d case? (@Giuseppe5, your comments here are appreciated)
>   * `src/brevitas/graph/per_input.py:replace_modules` does the conversion from `AvgPool3d` to `QuantConv3d` work and is it correct (does it produce the correct output)?

I didn't test these specifically, I'll take a look at them now to make sure they work properly

src/brevitas/graph/target/flexml.py
Giuseppe5: No

costigt-dev: shall I remove this part?

costigt-dev:
> >   * `src/brevitas/graph/fixed_point.py:MergeBatchNorm` does a 3d batch norm get merged into Conv3d correctly?
From local testing it seems it is correctly merged

src/brevitas/nn/quant_conv.py
Giuseppe5: I believe this to be correct

tests/brevitas_ort/test_quant_module.py
```diff
@@ -44,8 +45,10 @@ def test_ort_wbiol(model, export_type, current_cases):
         in_size = (1, IN_CH)
     elif impl in ('QuantConv1d', 'QuantConvTranspose1d'):
         in_size = (1, IN_CH, FEATURES)
-    else:
+    elif impl in ('QuantConv2d', 'QuantConvTranspose2d'):
         in_size = (1, IN_CH, FEATURES, FEATURES)
+    else:
```
Giuseppe5, line 50: Out of clarity, let's specify explicitly this case as well

src/brevitas/graph/target/flexml.py
Giuseppe5: Yeah

Giuseppe5:
> > * `src/brevitas/graph/fixed_point.py:MergeBatchNorm` does a 3d batch norm get merged into Conv3d correctly?
> 
> From local testing it seems it is correctly merged

Let's add some tests about that as well

src/brevitas/graph/gpxq.py
```diff
@@ -18,7 +18,12 @@
 from brevitas.quant_tensor import QuantTensor
 
 SUPPORTED_CONV_OP = (
-    qnn.QuantConv2d, qnn.QuantConv1d, qnn.QuantConvTranspose1d, qnn.QuantConvTranspose2d)
+    qnn.QuantConv3d,
```
Giuseppe5, line 21: Reorder

tests/brevitas/graph/test_transforms.py
```diff
@@ -50,6 +54,92 @@ def test_rewriter_merge_bn(model_name: str, pretrained: bool):
...
+        args = torch.randn(20, 16, 10, 50)
+    else:
+        args = torch.randn(20, 16, 10, 50, 100)
+
+    graph = symbolic_trace(model)
+    graph = AvgPoolToQuantDepthwiseConv().apply(graph, args)
+
+    has_quant_conv = False
+    for m in graph.modules():
+        if isinstance(m, QuantConv1d):
```
Giuseppe5, line 129: Collapse into:
`isinstance(m, (QuantConv1d, QuantConv2d, QuantConv3d))`

Giuseppe5 (approved):

src/brevitas/export/onnx/handler.py
```diff
@@ -88,6 +88,38 @@ def kernel_shape(module):
...
+        if isinstance(module.padding, int):
+            padding = [module.padding] * 6
+        else:
+            padding = list(module.padding) + list(module.padding)
+        return padding
+
+    @staticmethod
+    def stride(module):
+        if isinstance(module.stride, int):
+            return [module.stride] * 4
```
nickfraser, line 104: I think this should actually be `return [module.stride] * 3`

src/brevitas/export/onnx/handler.py
```diff
@@ -88,6 +88,38 @@ def kernel_shape(module):
...
+        else:
+            return list(module.stride)
+
+    @staticmethod
+    def dilation(module):
+        if isinstance(module.dilation, int):
+            return [module.dilation] * 4
```
nickfraser, line 111: `return [module.dilation] * 3`

src/brevitas/export/onnx/handler.py
```diff
@@ -88,6 +88,38 @@ def kernel_shape(module):
...
+        else:
+            return list(module.dilation)
+
+    @staticmethod
+    def kernel_shape(module):
+        if isinstance(module.kernel_size, int):
+            return [module.kernel_size] * 4
```
nickfraser, line 118: `return [module.kernel_size] * 3`

src/brevitas/export/onnx/handler.py
costigt-dev: ah - I thought they were going up by 2 - fixing now

src/brevitas/nn/quant_convtranspose.py
```diff
@@ -218,3 +220,102 @@ def max_acc_bit_width(self, input_bit_width, weight_bit_width):
...
+        else:
+            raise NotImplementedError(f"Padding mode {self.padding_mode} not supported.")
+
+    def max_acc_bit_width(self, input_bit_width, weight_bit_width):
+        max_uint_input = max_int(bit_width=input_bit_width, signed=False, narrow_range=False)
+        max_kernel_val = self.weight_quant.max_uint_value(weight_bit_width)
+        group_size = self.out_channels // self.groups
+        overlapping_sums = max(round(self.kernel_size[0] / self.stride[0]), 1)
+        overlapping_sums *= max(round(self.kernel_size[1] / self.stride[1]), 1)
+        overlapping_sums *= max(round(self.kernel_size[2] / self.stride[2]), 1)
```
nickfraser, line 318: Does this actually calculate the right thing? The way I read this function is that `group_size` * `overlapping_sums` should equal the number of terms in my dot product. If so, why is the stride important?

@i-colbert, @Giuseppe5, can you enlighten me here?

tests/brevitas/nn/test_a2q.py
```diff
@@ -102,12 +102,18 @@ def test_quant_wbiol_a2q(model_input, current_cases):
...
+            'model_type'] == 'QuantConv3d':  # shape = (out_channels, in_channels, kernel_size, kernel_size)
+        quant_weight_per_channel_l1_norm = quant_weight.norm(p=1, dim=(1, 2, 3, 4))
     elif kwargs[
             'model_type'] == 'QuantConvTranspose1d':  # shape = (in_channels, out_channels, kernel_size)
         quant_weight_per_channel_l1_norm = quant_weight.norm(p=1, dim=(0, 2))
     elif kwargs[
             'model_type'] == 'QuantConvTranspose2d':  # shape = (in_channels, out_channels, kernel_size)
         quant_weight_per_channel_l1_norm = quant_weight.norm(p=1, dim=(0, 2, 3))
+    elif kwargs[
+            'model_type'] == 'QuantConvTranspose3d':  # shape = (in_channels, out_channels, kernel_size)
```
nickfraser, line 115: Comment: `# shape = (in_channels, out_channels, kernel_size, kernel_size)`

src/brevitas/nn/quant_convtranspose.py
i-colbert: I have only worked with the 2D case, but if it is a trivial extension of that then this is almost correct. Unlike the convolution, which traverses the input space in strides, the deconvolution traverses the output space in strides and creates overlapping sums that are a function of the kernel size and stride.  [This paper](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9592768) introduces an algorithm the traverses the output space of the deconvolution (see Algorithm 8) and you can see the stride skipping there. However, you can see that it should be a floor rounding rather than a halfway rounding. I would propose the following:

```python
...
patch_size = (self.kernel_size[0] // self.stride[0]) * (self.kernel_size[1] // self.stride[1]) * (self.kernel_size[2] // self.stride[2])
max_uint_output = max_uint_input * max_kernel_val * patch_size * group_size
...
```

The kernel size shouldn't really outgrow the stride, since it wouldn't make much sense unless there is a niche use case. Not sure PyTorch throws an error in that case and also not sure if it'd be the job of the quant layer to protect against this.

nickfraser: Thank you! <3

nickfraser (approved):
LGTM!

Issue #799, opened 2024-01-21
FP Quantization test suite


Giuseppe5:
We are missing tests for FP quantization.

The corresponding integer ones can be used as reference.

PR #815, opened 2024-01-30
Fix (tests): adding tests for FloatQuant

fabianandresgrob:
### Issue:
Implements #799 

### Details:

- tests different minifloat configurations
- test default values for FloantQuant
- tests if input is quantized correctly


fabianandresgrob:
A problem that occurs is, that for some minifloat configurations, the inner_scale reaches `inf`, when encountering very small fp values. The question is how to handle an inner scale of 'inf'.

src/brevitas/core/quant/float.py
```diff
@@ -13,6 +13,9 @@
 from brevitas.function.ops import max_float
 from brevitas.function.ops_ste import floor_ste
 
+# max int that can be passed to torch.exp2() without running into inf
+MAX_REPRESENTABLE_INT = 127
```
Giuseppe5, line 17: I am changing my mind about this. Maybe we can handle the `inf` case in the tests somehow? As expected results?

tests/brevitas/hyp_helper.py
```diff
@@ -218,3 +220,21 @@ def min_max_tensor_random_shape_st(draw, min_dims=1, max_dims=4, max_size=3, wid
     min_tensor = torch.tensor(min_list).view(*shape)
     max_tensor = torch.tensor(max_list).view(*shape)
     return min_tensor, max_tensor
+
+
+@st.composite
+def random_minifloat_format(draw, min_bit_width=MIN_INT_BIT_WIDTH, max_bit_with=MAX_INT_BIT_WIDTH):
+    """"
+    Generate a minifloat format. Returns bit_width, exponent, mantissa, and signed.
```
Giuseppe5, line 228: Add a `TODO` here. The minifloat format will be different once we have support for a FloatQuantTensor (Soon TM)

tests/brevitas/hyp_helper.py
```diff
@@ -218,3 +220,21 @@ def min_max_tensor_random_shape_st(draw, min_dims=1, max_dims=4, max_size=3, wid
...
+    """
+    bit_width = draw(st.integers(min_value=min_bit_width, max_value=max_bit_with))
+    exponent_bit_width = draw(st.integers(min_value=0, max_value=bit_width))
+    signed = draw(st.booleans())
+    # if no budget is left, return
```
Giuseppe5, line 233: What does this mean?

fabianandresgrob: It means that the exponent_bit_width + signed at that point might already sum up to bit_width. If that's the case, we can't draw mantissa bits anymore, so returning it with a 0 for mantissa_bit_width is the only option

Giuseppe5: No the base case is 1 bit for everything (e.g., 3 bit floating point)

fabianandresgrob: Agreed, added the checks for that later, so this is still from the earlier tests.

src/brevitas/core/quant/float.py
```diff
@@ -35,10 +38,16 @@ def __init__(
         self.bit_width = StatelessBuffer(torch.tensor(float(bit_width), device=device, dtype=dtype))
```
fabianandresgrob, line 38: @Giuseppe5 if the sign bit is always there, this check also doesn't make sense anymore. If it is unsigned, the sign is just going to be 0, right? So one bit has to be reserved for the sign at all times anyways

PR #839, opened 2024-02-09
Fix (core): deprecate bitwidth-less bias

Giuseppe5:
Breaking change


src/brevitas/proxy/parameter_quant.py
```diff
@@ -180,39 +172,28 @@ def scale(self):
...
-                out, out_scale, out_zp, out_bit_width = impl(x, input_scale, input_bit_width)
-            elif self.requires_input_scale and not self.requires_input_bit_width:
+            if self.requires_input_scale:
                 input_scale = input_scale.view(-1)
                 out, out_scale, out_zp, out_bit_width = impl(x, input_scale)
-            elif not self.requires_input_scale and not self.requires_input_bit_width:
+            elif not self.requires_input_scale:
                 out, out_scale, out_zp, out_bit_width = impl(x)
             else:
                 raise RuntimeError("Internally defined bit-width required")
```
Giuseppe5, line 199: Rewrite

src/brevitas/quant/solver/bias.py
```diff
@@ -34,29 +32,17 @@ def scaling_per_output_channel_shape(module):
...
-    def tensor_quant(quant_type, requires_input_bit_width, requires_input_scale):
+    def tensor_quant(quant_type, requires_input_scale):
         if quant_type == QuantType.FP:
             return None
         elif quant_type == QuantType.INT:
-            if not requires_input_bit_width and requires_input_scale:
+            if requires_input_scale:
                 return PrescaledRestrictIntQuant
-            elif not requires_input_bit_width and not requires_input_scale:
+            elif not requires_input_scale:
```
Giuseppe5, line 44: else

Giuseppe5, line 44: else

Giuseppe5:
This depends on #847

src/brevitas_examples/imagenet_classification/qat/README.md
```diff
@@ -5,12 +5,12 @@ and by no means a direct mapping to hardware should be assumed.
 
 Below in the table is a list of example pretrained models made available for reference.
 
-| Name         | Cfg                   | Scaling Type               | First layer weights | Weights | Activations | Avg pool | Top1  | Top5  | Pretrained model                                                                                | Retrained from                                                |
```
nickfraser, line 8: Any reason why you also removed Top-5?

nickfraser (approved):
Other than the question about Top-5, LGTM!

src/brevitas_examples/imagenet_classification/qat/README.md
Giuseppe5: We don't compute top5 accuracy anymore because:
- It is not as relevant as the Top 1
- I went down the rabbit hole of removing/reducing variability while performing 2 runs with the same seed, and this has led to removing the computation of top5 accuracy. Since we don't generate it while running the script, I thought I would also remove it from the results as it could be confusing.

I am happy to restore it in the readme, but less inclined to restore its computation.

PR #846, opened 2024-02-13
Fix (QuantLayer): make bias for QuantLayer optional

fabianandresgrob:
### Description:
Currently, the user always has to set the bias parameter when initializing a QuantLayer. This PR makes it optional to pass the parameter, as wished in #736 .
Changed this for the following layer types:
- QuantLinear
- QuantConvXd
- QuantConvTransposeXd

### Open Questions:
- what about bias for MHA?


Giuseppe5:
Extend also to QuantRNN e QuantLSTM.

Issue #843, opened 2024-02-13
Explicitly print output of asserts in notebook examples


nickfraser:
This will help running & validating notebooks from the commandline.

Issue #842, opened 2024-02-13
Make notebook examples deterministic


nickfraser:
Some cells have a seed set, but we need to do this everywhere where it's necessary. This would make it much easier to verify them with a script.

fabianandresgrob:
As `torch.manual_seed()` is creating a generator, it should be sufficient to set it once at the start of each notebook.
Or does a seed per cell (given that any numbers are generated) have any other advantages?

nickfraser:
Start of the notebook should be fine, but we shouldn't change the cells where it is set explicitly.

PR #847, opened 2024-02-13
Notebooks: makes notebooks deterministic and prints output of asserts

fabianandresgrob:
PR for #842 and #843


fabianandresgrob:
Checked using that script, everything looks good. Only minor things differ like kernel id. I added one more `torch.manual_seed()` to the ONNX export.

Giuseppe5:
When doing multiple runs to compare, restart the kernels, just to be sure

notebooks/02_quant_activation_overview.ipynb
```diff
@@ -315,16 +326,27 @@
...
+    }
+   ],
    "source": [
     "return_disabled_quant_relu = QuantReLU(act_quant=None, return_quant_tensor=True)\n",
     "relu_out_tensor = return_disabled_quant_relu(out_tensor)\n",
-    "assert relu_out_tensor.is_valid==True\n",
-    "assert relu_out_tensor.scale == out_tensor.scale\n",
-    "assert relu_out_tensor.zero_point == out_tensor.zero_point\n",
-    "assert relu_out_tensor.bit_width == out_tensor.bit_width"
+    "assert_with_message((relu_out_tensor.is_valid==True))\n",
```
Giuseppe5, line 346: Double parenthesis

Giuseppe5: Also, let's remove `assert condition == True`, and change it to `assert condition`.
Similar with `assert condition == False` that becomes `assert not condition`
This line becomes :
```suggestion
    "assert_with_message(relu_out_tensor.is_valid)\n",
```
I think there are other instances around the notebooks

nickfraser (approved):
LGTM!

Issue #827, opened 2024-02-06
Importing quantized models after bias correction


i-colbert:
When quantizing floating-point models that don't have a bias in their layer (e.g., `nn.Linear(in_features=10, out_features=2, bias=False)`), bias correction currently will add a bias to the layer. This leads to the new bias being exported with the state dictionary. However, when loading this modified state dictionary in a new instance of original model, there is a missing keys error from pytorch because there is no bias in the floating-point (or even quantized model) without first running bias correction.

The issue can be resolved by first running bias correction before loading the modified state dictionary (see below), but a more flexible solution may be to add support into the state dictionary loading mechanism itself.

```python
def _prepare_bias_corrected_quant_model(model: nn.Module):
    model.eval()
    dtype = next(model.parameters()).dtype
    device = next(model.parameters()).device
    images = torch.randn(10, 3, 32, 32)
    images = images.to(device)
    images = images.to(dtype)
    with torch.no_grad():
        with bias_correction_mode(model):
            model(images)
```

PR #868, opened 2024-02-20
Enhance: Importing quantized models after bias correction

costigt-dev:
## Issue
Implements https://github.com/Xilinx/brevitas/issues/827

## Details

- Adds new context manager **_load_quant_model_** that silently adds (and removes if unused) biases to help load any state dictionary that may have had biases unexpectedly added via use of the **_bias_correction_mode_** context manager prior to saving
- Adds flag to **_bias_correction_mode_** and **__BiasCorrection_** to disable the adding of corrections to layers that don't have a bias, default behaviour is the same as before so without modifying the flag, correction will still be applied
- Adds two tests for the above changes

### DIFF
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -113,9 +113,9 @@ class allow_unexpected_bias_keys:
 
 class bias_correction_mode:
 
-    def __init__(self, model, enabled=True, only_layers_with_bias=False):
+    def __init__(self, model, enabled=True, skip_if_no_bias=False):
         self.model = model
-        self.bias_correction = _BiasCorrection(only_layers_with_bias=only_layers_with_bias)
+        self.bias_correction = _BiasCorrection(skip_if_no_bias=skip_if_no_bias)
         self.enabled = enabled
         self.hooks = []
 
@@ -235,7 +235,7 @@ class _BiasCorrection(DisableEnableQuantization):
 
     LAYERS = (QuantWBIOL,)
 
-    def __init__(self, layers=LAYERS, only_layers_with_bias=False):
+    def __init__(self, layers=LAYERS, skip_if_no_bias=False):
         super(_BiasCorrection, self).__init__()
         self.layers = layers
         self.iterations = {}
@@ -243,7 +243,7 @@ class _BiasCorrection(DisableEnableQuantization):
         self.float_mean_map = {}
         self.collect_float_mean_hooks = []
         self.correct_bias_hooks = []
-        self.only_layers_with_bias = only_layers_with_bias
+        self.skip_if_no_bias = skip_if_no_bias
 
     def compute_mean(self, inp, transpose_dim):
         inp = inp.transpose(0, transpose_dim)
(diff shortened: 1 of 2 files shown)


src/brevitas/graph/calibrate.py
```diff
@@ -85,11 +89,33 @@ def __exit__(self, type, value, traceback):
...
+                if module.bias is None:
+                    module.register_parameter(
+                        'bias',
+                        nn.Parameter(torch.empty(module.weight.shape[0])).to(module.weight.device))
+                    self.tracked_modules.append(module)
+
+    def __exit__(self, type, value, traceback):
+        for module in self.tracked_modules:
+            # empty tensor has a numel result of 0
+            if torch.numel(module.bias) == 0:
```
Giuseppe5, line 110: let's check that str(module.device) == 'meta'

src/brevitas/graph/calibrate.py
```diff
@@ -85,11 +89,33 @@ def __exit__(self, type, value, traceback):
...
+                module.bias = None
+
+
 class bias_correction_mode:
 
-    def __init__(self, model, enabled=True):
+    def __init__(self, model, enabled=True, only_layers_with_bias=False):
```
Giuseppe5, line 116: rename the flag to:
skip_if_no_bias

src/brevitas/graph/calibrate.py
costigt-dev: the approach i've used doesn't use the meta device but rather an empty tensor as the meta tensor seemed to add a lot of complication and warnings

Giuseppe5: If I'm not mistaken, an empty tensor has "meta" as device by default. Or am I missing something?

costigt-dev: My testing has the default as 'cpu'
`torch.empty(10).device` 
gives 'cpu'

Giuseppe5: Out of curiosity, what warnings do you get with meta tensors?

costigt-dev: I didn't go too deep into a meta tensor implementation but warnings like:
> UserWarning: for net.0.bias: copying from a non-meta parameter in the checkpoint to a meta parameter in the current model, which is a no-op.

src/brevitas/graph/calibrate.py
```diff
@@ -24,7 +24,11 @@
 from .base import Transform
 
 __all__ = [
-    'ClipFloatWeights', 'DisableEnableQuantization', 'bias_correction_mode', 'calibration_mode']
+    'ClipFloatWeights',
+    'DisableEnableQuantization',
+    'bias_correction_mode',
+    'calibration_mode',
+    'allow_unexpected_bias_keys']
```
Giuseppe5, line 31: Let's rename `allow_unexpected_bias_keys` to `load_quant_model`.
In the future we can use this as a more general entrypoint to load checkpoint with special requirements.

For example, at the moment we have a global flag to ignore missing keys when loading a floating point dict in a quantized model, and it's up to the user to enable that flag. 

This is a matter for another PR anyway.

src/brevitas/graph/calibrate.py
```diff
@@ -85,11 +89,33 @@ def __exit__(self, type, value, traceback):
             self.model, is_training=self.previous_training_state, quantization_enabled=True)
 
 
+class load_quant_model:
+
+    def __init__(self, model):
+        self.model = model
+
+    def __enter__(self):
+        self.tracked_modules = []
```
Giuseppe5, line 98: Let's define this in the init

Issue #831, opened 2024-02-07
Documentation for Dynamic Quantization export


Giuseppe5:
We are expanding our support for Dynamic Quantization, and there is already the possibility to export it to ONNX Op DynamicQuantizeLinear (when matching its constraints).

This should be documented in corresponding notebook with examples.

PR #877, opened 2024-02-26
Feat (ONNX Export): move dynamic quantizers and add example for export

fabianandresgrob:
PR for #831 
- Adding an example export to ONNX for dynamic quantization. Also included limitations for the export in the ONNX Export notebook. 
- Moving the dynamic quantizers under `src/brevitas/quant/generative`


Giuseppe5:
We don't need to move the quantizers, just add the example

Giuseppe5 (approved):

PR #819, opened 2024-02-02
Quant tensor not empty

Giuseppe5:

PR #880, opened 2024-02-26
Fix (proxy): fix for attributes retrieval

Giuseppe5:
We don't have tests to check for these changes, that should have failed when we merged #819

### DIFF
--- a/tests/brevitas/proxy/test_proxy.py
+++ b/tests/brevitas/proxy/test_proxy.py
@@ -1,0 +1,82 @@
+import pytest
+
+from brevitas.nn import QuantLinear
+from brevitas.nn.quant_activation import QuantReLU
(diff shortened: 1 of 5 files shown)


tests/brevitas/proxy/test_proxy.py
```diff
@@ -0,0 +1,80 @@
...
+from brevitas.quant.scaled_int import Int8WeightPerChannelFloatDecoupled
+from brevitas.quant.scaled_int import Int8WeightPerTensorFloat
+from brevitas_examples.common.generative.quantizers import Int8DynamicActPerTensorFloat
+
+
+class TestProxy:
+
+    def test_bias_proxy(self):
+        model = QuantLinear(10, 5, bias_quant=Int8BiasPerTensorFloatInternalScaling)
+        assert model.weight_quant.scale() is not None
```
Giuseppe5, line 16: model.bias_quant

PR #890, opened 2024-03-01
Fix (GPFQ): use max/mean to avoid running out of memory

fabianandresgrob:
For the GPFQ algorithm, larger models lead to an OutOfMemory error. This is because the GPFQ algorithm uses all calibration data provided. Instead of iteratively concatenating the new input to all previous ones, we could use a running average, or always select the max of the new input and the previous ones.


Giuseppe5:
Closing this since it seems to provide poorer performances compared to standard implementation

PR #892, opened 2024-03-02
Fix (gpxq): handling empty tensors with GPxQ and adding unit tests

i-colbert:


Giuseppe5:
In general, it's better not to require multiple reviews because too many tests are triggered

tests/brevitas/graph/test_gpxq.py
```diff
@@ -0,0 +1,135 @@
...
+import torch.nn as nn
+from torch.utils.data import DataLoader
+from torch.utils.data import TensorDataset
+
+from brevitas.graph.gpfq import gpfq_mode
+from brevitas.graph.gptq import gptq_mode
+import brevitas.nn as qnn
+
+
+class QuantConvModel(nn.Module):
```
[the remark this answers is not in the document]
Giuseppe5: For Conv/ConvTranspose, we should have group == 1 and group > 1 as test cases

src/brevitas/graph/gpfq.py
```diff
@@ -302,13 +308,36 @@ def __init__(
             p=p)
         self.accumulator_bit_width = accumulator_bit_width
         assert self.accumulator_bit_width is not None
-        self.requires_quant_input = True  # force true
+
+    def process_input(self, inp):
+        inp = super().process_input(inp)
+        inp = self.layer.input_quant(inp)
```
Giuseppe5, line 314: Is this here because we attach forward-pre-hook to the layer?

i-colbert: Exactly. So the quant layer input is a tensor, but the input to the inner forward pass is after the input quant. I added a unit test to make sure that works as well since it wasn't before.

tests/brevitas/graph/test_gpxq.py
```diff
@@ -0,0 +1,143 @@
...
+        request):
+
+    test_id = request.node.callspec.id
+
+    torch.manual_seed(SEED)
+
+    name, apply_gpxq = apply_gpxq_tuple
+
+    if (name == 'gptq' and acc_bit_width < 32):
+        pytest.skip("GPTQ does not support accumulator-aware quantization.")
```
Giuseppe5, line 105: I would skip tests only when the testing configuration does not make sense.

For this case, I would raise an error in the code and catch it here with `pytest.raises`.

Similar for the one below.

tests/brevitas/graph/equalization_fixtures.py
Giuseppe5, line 1: Not sure these changes should live in this file to be honest. In any case, let's clearly separate the old vs new so it will be easier to migrate

Issue #727, opened 2023-10-18
QuantTensor division 


volcacius:
We should test the correctness of division over QuantTensor

capnramses:
* add a test for this
* consider generalising to test more elements of QuantTensor

PR #894, opened 2024-03-05
Enhance: Quant Tensor Test

costigt-dev:
## Issue
Implements https://github.com/Xilinx/brevitas/issues/727

## Details
Creates test file for QuantTensor

Has tests for:
- maths operators
- divide by zero
- transpose
- view

It's not exhaustive but should form a good starting point.


tests/brevitas/core/test_quant_tensor.py
Giuseppe5, line 1: Let's create a folder brevitas/quant_tensor for these tests

costigt-dev: I'm inclined to leave it as is at the moment as the other tests in this area don't have their own folders and I want to keep this PR nice and simple. If it expands later we can create a folder.

Giuseppe5: All tests in brevitas/core refers to file that are srcs/brevitas/core (we don't follow all the subfolders but that's the general idea).

tests/brevitas/quant_tensor/test_quant_tensor.py
```diff
@@ -100,7 +100,7 @@ def test_quant_tensor_view():
 def test_is_valid():
     x = torch.randn(4, 4)
     # directly initialised QuantTensor shouldn't be valid
-    invalid_quant_tensor = QuantTensor(x)
+    invalid_quant_tensor = QuantTensor(x, None, None, None, None, None)
```
Giuseppe5, line 103: It is actually possible to initialize a valid QuantTensor from scratch, if you handpick your values, so the comment is not precise.

Instead of picking None which causes only pain, i would pick random number.

costigt-dev: I added None because some of the test workflows weren't accepting the default arguments.
The comment is just referring to the scenario where its directly initiated without doing anything else but I guess it is ambiguous.

tests/brevitas/quant_tensor/test_quant_tensor.py
```diff
@@ -0,0 +1,97 @@
...
+        quant = qa - qb
+    elif op == Operator.DIVIDE:
+        normal = a / b
+        quant = qa / qb
+    elif op == Operator.MULTIPLY:
+        normal = a * b
+        quant = qa * qb
+    elif op == Operator.MATMUL:
+        normal = a @ b
+        # @ matmul operator not implemented for QuantTensor
```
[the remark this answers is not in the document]
costigt-dev: I don't believe there is a difference so its probably something we should create an issue to implement

tests/brevitas/quant_tensor/test_quant_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+    return (
+        torch.round(normal_tensor / quant_tensor.scale + quant_tensor.zero_point) -
+        quant_tensor.zero_point) * quant_tensor.scale
+
+
+def test_quant_tensor_init():
+    x = torch.randn(4, 4)
+    quant_tensor = to_quant_tensor(x)
+    normal_tensor = torch.Tensor(x)
+    assert torch.allclose(qdq(normal_tensor, quant_tensor), quant_tensor, rtol=0.01)
```
costigt-dev, line 36: difference between the qdq result and quant tensor is extremely close but some error is creeping in from the quanttensor somewhere so added relative tolerance

tests/brevitas/quant_tensor/test_quant_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+    x = torch.randn(4, 4)
+
+    a = torch.Tensor(x)
+    b = torch.Tensor(x)
+
+    qa = to_quant_tensor(a)
+    qb = to_quant_tensor(b)
+
+    # to factor in quantisation error
+    e_a = a - qa
```
costigt-dev, line 51: didn't use qdq approach above as should be covered by the init test, I just need the difference so I can incorporate it into the calculations below

tests/brevitas/quant_tensor/test_quant_tensor.py
```diff
@@ -18,53 +18,61 @@ class Operator(Enum):
...
     elif op == Operator.MATMUL:
-        normal = a @ b
         # @ matmul operator not implemented for QuantTensor
         quant = torch.matmul(qa, qb)
+        normal = (a - e_a) @ (b - e_b)
     else:
         # unrecognised operator
         assert False
 
     # tolerance set to a high value as there is considerable loss of precision
```
Giuseppe5, line 74: Comment is outdated I believe

Giuseppe5 (approved):
Just a small clean-up of comments but otherwise it's ready to go

PR #895, opened 2024-03-06
Ignore

costigt-dev:


costigt-dev:
> Structure of the workflow was tested in separate branch but ultimately the issue_comment trigger will only run if on the default branch.

`Structure of the workflow was tested in separate branch but ultimately the issue_comment trigger will only run if on the default branch.`

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment

Giuseppe5:
> `Structure of the workflow was tested in separate branch but ultimately the issue_comment trigger will only run if on the default branch.`
> 
> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment

So we need to have this in master to work?

costigt-dev:
> > `Structure of the workflow was tested in separate branch but ultimately the issue_comment trigger will only run if on the default branch.`
> > https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment
> 
> So we need to have this in master to work?

yep - it's not a great design on the github actions side. Several events ONLY happen on the master branch so its a bit of a leap of faith adding them. So we'll want to scrutinize what i've added very carefully before merging.

Giuseppe5:
You can try to merge this in master in your fork and see what happens there?

costigt-dev:
> You can try to merge this in master in your fork and see what happens there?

excellent idea!

costigt-dev:
> > You can try to merge this in master in your fork and see what happens there?
> 
> excellent idea!

Works in fork

PR #898, opened 2024-03-06
Fix (ptq): expose uint_sym_act flag and fix issue with minifloat sign

fabianandresgrob:
So far, the `ptq_evaluate` in the imagenet ptq example didn't expose the `uint_sym_act_for_unsigned_values` flag, which makes act quantizers unsigned in order to use full bit width. However, this doesn't work with minifloat configurations, so a check is introduced for the `act_quant_format` and the `uint_sym_act_for_unsigned_values` flag is further exposed to the user running the `ptq_evaluate.py` file.

### DIFF
--- a/src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
+++ b/src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
@@ -251,6 +251,11 @@ add_bool_arg(
     'merge-bn',
     default=True,
     help='Merge BN layers before quantizing the model (default: enabled)')
+add_bool_arg(
+    parser,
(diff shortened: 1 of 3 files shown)


src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
```diff
@@ -251,6 +251,11 @@ def parse_type(v, default_type):
     'merge-bn',
     default=True,
     help='Merge BN layers before quantizing the model (default: enabled)')
+add_bool_arg(
```
Giuseppe5, line 254: You need to add this to the README.

Generate it by running
```python ptq_evaluate.py -h```

PR #909, opened 2024-03-16
Feat (examples): add support for Stable Diffusion XL

Giuseppe5:
Missing:

- [x] GPTQ 
- [x] Activation equalization (layerwise if FX does not work)
- ~Possibly Learned Round but it could be very slow~

Torchscript export is deprecated since it is not used and difficult to make compatible with SD-XL.

More investigations are needed to fix it in case it is required.

### DIFF
--- a/src/brevitas_examples/stable_diffusion/main.py
+++ b/src/brevitas_examples/stable_diffusion/main.py
@@ -7,28 +7,39 @@ import argparse
 from datetime import datetime
 import json
 import os
-import re
 import time
 
 from dependencies import value
-from diffusers import StableDiffusionPipeline
+from diffusers import DiffusionPipeline
 import torch
 from torch import nn
+from tqdm import tqdm
 
 from brevitas.export.onnx.standard.qcdq.manager import StdQCDQONNXManager
 from brevitas.export.torch.qcdq.manager import TorchQCDQManager
+from brevitas.graph.calibrate import bias_correction_mode
+from brevitas.graph.calibrate import calibration_mode
+from brevitas.graph.equalize import activation_equalization_mode
+from brevitas.graph.gptq import gptq_mode
+from brevitas.nn.equalized_layer import EqualizedModule
+from brevitas.utils.torch_utils import KwargsForwardHook
 from brevitas_examples.common.generative.quantize import quantize_model
 from brevitas_examples.common.parse_utils import add_bool_arg
 from brevitas_examples.common.parse_utils import quant_format_validator
 from brevitas_examples.llm.llm_quant.export import BlockQuantProxyLevelManager
 from brevitas_examples.stable_diffusion.sd_quant.constants import SD_2_1_EMBEDDINGS_SHAPE
+from brevitas_examples.stable_diffusion.sd_quant.constants import SD_XL_EMBEDDINGS_SHAPE
 from brevitas_examples.stable_diffusion.sd_quant.export import export_onnx
-from brevitas_examples.stable_diffusion.sd_quant.export import export_torchscript
 from brevitas_examples.stable_diffusion.sd_quant.utils import generate_latents
-from brevitas_examples.stable_diffusion.sd_quant.utils import generate_unet_rand_inputs
+from brevitas_examples.stable_diffusion.sd_quant.utils import generate_unet_21_rand_inputs
+from brevitas_examples.stable_diffusion.sd_quant.utils import generate_unet_xl_rand_inputs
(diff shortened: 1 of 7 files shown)


src/brevitas_examples/stable_diffusion/main.py
```diff
@@ -128,32 +183,41 @@ def bit_width(module):
             pipe, args.resolution, prompts, test_seeds, output_dir, args.device, dtype)
 
     if args.export_target:
+        assert args.weight_quant_format == 'int', "Only integer quantization supported for export."
```
Giuseppe5, line 186: Reword

src/brevitas_examples/stable_diffusion/sd_quant/utils.py
```diff
@@ -47,3 +47,35 @@ def generate_unet_rand_inputs(
...
+    unet_rand_inputs = generate_unet_rand_inputs(
+        embedding_shape, unet_input_shape, batch_size, device, dtype, with_return_dict_false)
+    sample = unet_rand_inputs['sample']
+    del unet_rand_inputs['sample']
+    unet_rand_inputs['timestep_cond'] = None
+    unet_rand_inputs['cross_attention_kwargs'] = None
+    unet_rand_inputs['added_cond_kwargs'] = {
+        "text_embeds": torch.randn(1, 1280, dtype=dtype, device=device),
+        "time_ids": torch.randn(1, 6, dtype=dtype, device=device)}
+    inputs = (sample, unet_rand_inputs)
```
Giuseppe5, line 80: Add comment to explain this magic

src/brevitas_examples/stable_diffusion/main.py
```diff
@@ -89,36 +111,102 @@ def main(args):
...
+    # Move model to target device
+    print(f"Moving model to {args.device}...")
+    pipe = pipe.to(args.device)
+
+    if args.activation_equalization:
+        with activation_equalization_mode(pipe.unet, alpha=0.5, layerwise=True, add_mul_node=True):
+            # Workaround to expose `in_features` attribute from the Hook Wrapper
+            for m in pipe.unet.modules():
+                if isinstance(m, KwargsForwardHook) and hasattr(m.module, 'in_features'):
+                    m.in_features = m.module.in_features
```
Giuseppe5, line 123: @costigt-dev This (and similarly, line 129-131) is why I need the wrapper to expose the attributes of the wrapped module.

In this case, the wrapping is only temporary so I need to keep some sort of reference of wrapper vs wrapped

src/brevitas_examples/stable_diffusion/main.py
```diff
@@ -185,48 +283,98 @@ def bit_width(module):
...
+        '--input-quant-format',
+        type=quant_format_validator,
+        default='int',
+        help=
+        'Weight quantization type. Either int or eXmY, with X+Y==weight_bit_width-1. Default: int.')
     parser.add_argument(
         '--weight-quant-granularity',
         type=str,
-        default='per_group',
+        default='per_channel',
```
capnramses, line 363: could you explain a bit why this changes from per group to per channel?

src/brevitas_examples/stable_diffusion/main.py
```diff
@@ -128,32 +216,42 @@ def bit_width(module):
...
             device=device,
             dtype=dtype)
+
         if args.export_target == 'torchscript':
             if args.weight_quant_granularity == 'per_group':
                 export_manager = BlockQuantProxyLevelManager
             else:
                 export_manager = TorchQCDQManager
-                export_manager.change_weight_export(export_weight_q_node=True)
+                export_manager.change_weight_export(export_weight_q_node=args.export_weight_q_node)
```
capnramses, line 247: does the value from args always resolve to True/False type?

capnramses (approved):
Because there are mostly input and argument changes in this PR my only real question would be - have you run it or tested it to double check there isn't any clash with arguments? I'd like to give it a  run myself (with supervision) since it's a hot topic demo but I don't want to hold up the PR, so I'm hitting 'approve'.

Great to see some improvements to comments in tricky code and notes on PR to add context!

I asked a couple of "put me in context here" questions, but I don't see any code that looks like it could be a bug. I suggest a bit of background info in PR descriptions would be good for next PRs.

src/brevitas_examples/stable_diffusion/main.py
Giuseppe5: We have more robust support for per_channel quantization, including export representation.

We can export per_group quantized network but we're still defining the best way to represent that quantization style.

PR #915, opened 2024-03-20
Fix (GPxQ): unwrap QuantTensor when dealing with QuantLinear

fabianandresgrob:
When running GPxQ and we use quantized activations, once a `QuantLinear` layer is processed, `update_batch` tries to unsqueeze (GPFQ) or transpose (GPTQ) that `QuantTensor`, leading to an error.

This PR solves that issue by unwrapping that tensor. It also adds a quant linear model in the fixtures for reproducing the issue.
Some of the tests still fail as it does not raise a `ValueError` because it actually has a `quant_input` ([here](https://github.com/Xilinx/brevitas/blob/bb4feb293ece5d521bfb15c3e9104a2e210ace29/src/brevitas/graph/gpfq.py#L340)). @i-colbert maybe we need to change the conditions for an expected fail?


Giuseppe5:
@fabianandresgrob would you mind fixing the error so we can merge this?

PR #920, opened 2024-03-22
Fix/remove-numel - Remove numel is zero check from context manager exit method

costigt-dev:
## Reason for this PR

torch.empty and torch.numel don't work as originally expected although this code path was unlikely to be triggered in the original scenario this context manager was added to resolve

## Changes Made in this PR

Removed the contents of the exit method in the load_quant_model context manager as the method to detect if the newly added biases were corrected won't work as intended.

Also replaced empty with zeros so the bias layer won't affect anything if unused.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/graph/calibrate.py
+++ b/src/brevitas/graph/calibrate.py
@@ -13,7 +13,6 @@ import torch.nn.functional as F
 from brevitas.nn import QuantHardTanh
 from brevitas.nn import QuantLinear
 from brevitas.nn.quant_layer import QuantWeightBiasInputOutputLayer as QuantWBIOL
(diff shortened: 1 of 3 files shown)


Giuseppe5:
We would still have an issue if the state dict the user is trying to load has no bias for certain modules, correct?

costigt-dev:
> We would still have an issue if the state dict the user is trying to load has no bias for certain modules, correct?

yes, i'm not sure of the best way to handle that

Giuseppe5:
New change modifies the load state dict function withing a QuantWBIOL layer.

Issue #908, opened 2024-03-15
Adding tests for "quantize" function for CNN PTQ


Giuseppe5:

OscarSavolainenDR:
I've started working on making unit tests for `quantize_model`, and I had a few questions:
- `quantize_model` seems to expect a certain model. E.g., in `ptq_evaluate`, `quantize_model` is called after either [`preprocess_for_quantization`](https://github.com/OscarSavolainenDR/brevitas/blob/200456825f3b4b8db414f2b25b64311f82d3991a/src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py#L364) or [`preprocess_for_flexml_quantize`](https://github.com/OscarSavolainenDR/brevitas/blob/200456825f3b4b8db414f2b25b64311f82d3991a/src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py#L357) is called, depending on the quantization backend used.

    Should I make tests with the understanding that the model should be pre-processed accordingly, or just test `quantize_model`  as it is with any vanilla model? If they should only ever be called together, should those functions be packaged together?


- The function doesn't work for quantizing Transformer models, because it quantizes the inputted tokens. Once those quantized (now float) tokens get to the embedding layer, it fails because of the non-integer tokens. This may be expected behavior for this function if it is built for quantizing CNN-based models, but I thought I would raise the issue. This is also an effect of the quantization of inputs being `True` by default in `quantize_model` (quantization is True for bias, input, and weight, but not for output), which is contrary to what is said to be the standard for layers in one of the [tutorials.](https://github.com/OscarSavolainenDR/brevitas/blob/master/notebooks/01_quant_tensor_quant_conv2d_overview.ipynb):

    ``` 
    By default weight_quant=Int8WeightPerTensorFloat, while bias_quant, input_quant and output_quant are set to None. 

    That means that by default weights are quantized to 8-bit signed integer with a per-tensor floating-point scale factor (a 
    very common type of quantization adopted by e.g. the ONNX standard opset), while quantization of bias, input, and 
    output are disabled. We can easily verify all of this at runtime on an example:
    ```

    This may be desired/expected behavior for this function, but I wasn't sure.
    
[...]

PR #931, opened 2024-04-09
Assert all ptq-common bit widths are positive integers

OscarSavolainenDR:
In [testing](https://github.com/Xilinx/brevitas/issues/908) I found that `ptq_common.py` `quantize_model` bit widths can take negative and zero values. In the case of zero values, the model outputs NaNs, and in the case of negative values it is a bit more complicated (real values get outputted), but it is still undesired behavior.

This PR adds a constraint that all 'ptq_common/quantize_model' bit widths should be positive integers.

### DIFF
--- a/src/brevitas_examples/imagenet_classification/ptq/ptq_common.py
+++ b/src/brevitas_examples/imagenet_classification/ptq/ptq_common.py
@@ -155,7 +155,7 @@ def quantize_model(
     act_scale_type = scale_factor_type
 
     # We check all of the provided values are positive integers
-    check_bit_widths(
+    check_positive_int(
         weight_bit_width,
         act_bit_width,
         bias_bit_width,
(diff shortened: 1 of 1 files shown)


Giuseppe5 (approved):
I was about to suggest somehing like this to fix the issue.

Going a bit deeper, bit width requirements can be further specified, for example bias, weight, and activation bit width for integer quantization can only be greater than 3, since for Ternary and Binary quantization we need a different setup.
For floating point quantization, we can have also 1 bit for mantissa or exponent.

However these checks should probably live somewhere else in the code (e.g., the quantizer should verify that it's `valid`), so for now this assert should be sufficient.

Giuseppe5:
Can you run the pre-commit hooks please?

OscarSavolainenDR:
> Can you run the pre-commit hooks please?

Yep, have done so now!

PR #937, opened 2024-04-17
HQO for scale/zero point

Giuseppe5:
Implement HQO optimization for scale and zero point.
Caveats:
- Scale does not seem to perform that well
- Zero point supports per_tensor/per_channel/per_group (asym only)
- Scale supports per_tensor/per_channel
- More numerical tests are needed to evaluate the impact of HQO compared to other techniques
- Scale implementation is INT-only
- Zero point implementation does not make sense for FP8 since we mostly do symmetric quantization

Maybe worth to merge after #1002 and expand tests

### DIFF
--- a/src/brevitas/core/stats/stats_op.py
+++ b/src/brevitas/core/stats/stats_op.py
@@ -10,6 +10,8 @@ from torch.nn import Parameter
 
 import brevitas
 from brevitas import config
+from brevitas.core.function_wrapper.misc import Identity
+from brevitas.core.function_wrapper.ops_ste import ScalarClampMinSte
 from brevitas.core.utils import StatelessBuffer
 from brevitas.function.ops import max_int
 from brevitas.quant_tensor import _unpack_quant_tensor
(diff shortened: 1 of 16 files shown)


src/brevitas/core/stats/stats_op.py
```diff
@@ -544,3 +546,216 @@ def forward(self, x):
...
+                mask = (num != 0.) & (den != 0.)
+                if self.stats_reduce_dim is None:
+                    candidate = masked_median(num / den, mask)
+                else:
+                    candidate = masked_median(num / den, mask, dim=self.stats_reduce_dim)
+                candidate = self.input_view_shape_impl(candidate)
+                candidate[torch.isnan(candidate)] = self.internal_candidate[torch.isnan(candidate)]
+                candidate = self.clamp_min_ste(candidate)
+                bit_width = self.msb_clamp_bit_width_impl()
+                int_threshold = self.int_scaling_impl(bit_width)
```
Giuseppe5, line 621: Int specific implementation, maybe it should be generalized.

src/brevitas/quant/base.py
```diff
@@ -502,3 +504,53 @@ class MSEWeightZeroPoint(MSEZeroPoint):
 
 class MSEActZeroPoint(MSEZeroPoint):
     zero_point_impl = ParameterFromRuntimeZeroPoint
+
+
+class HQOZeroPoint(ExtendedInjector):
+    """
+    We leverage a sub-injector to avoid a name clash between scale and zero-point.
```
Giuseppe5, line 511: Remove here and next

src/brevitas/core/stats/stats_op.py
```diff
@@ -440,6 +442,21 @@ def _set_local_loss_mode(module, enabled):
             m.local_loss_mode = enabled
 
 
+def _set_quantize_zero_point(module, enabled):
```
Giuseppe5, line 445: Remove

src/brevitas/core/stats/stats_op.py
nickfraser: This looks like it hasn't been resolved to me. Is that true?

nickfraser (commented):
Is it intentional that this only works for Int?

PR #938, opened 2024-04-19
Feat (proxy): scale computation delegated to bias proxy

Giuseppe5:
As part of decoupling QuantLayers and QuantTensor, the computation of output scale for bias quantization is now delegated to the underlying proxy, who takes as input the possibly quantized input and weights, and internally defines the output scale.

Compared to our current setup, the main change would happen in the case where:
- bias quantizer is shared across multiple layers
- These layers are of different types (e.g., Linear and Conv)

Our current setup would have no issue with this case, while after this PR, the user would be required to use internally scaled bias and not externally defined.
I believe this to be a edge case that can be safely ignored.

### DIFF
--- a/src/brevitas/proxy/parameter_quant.py
+++ b/src/brevitas/proxy/parameter_quant.py
@@ -261,7 +261,7 @@ class BiasQuantProxyFromInjector(ParameterQuantProxyFromInjector, BiasQuantProxy
             self,
             x: Tensor,
             input: Optional[Union[Tensor, QuantTensor]] = None,
-            weight: Union[Tensor, QuantTensor] = None) -> Union[Tensor, QuantTensor]:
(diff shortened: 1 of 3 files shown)


src/brevitas/proxy/parameter_quant.py
```diff
@@ -234,10 +235,38 @@ def bit_width(self):
         bit_width = self.__call__(self.tracked_parameter_list[0], zhs).bit_width
         return bit_width
 
-    def forward(self,
-                x: Tensor,
-                input_scale: Optional[Tensor] = None) -> Union[Tensor, QuantTensor]:
+    def quant_output_scale_impl(
+            self, input: QuantTensor, weight: QuantTensor, module: torch.nn.Module) -> Tensor:
+        channel_dim = -1 if isinstance(module, torch.nn.Linear) else 1
```
costigt-dev, line 240: how come its -1 if linear and 1 otherwise?

PR #943, opened 2024-04-25
Fix (gpxq): adding input quant to process input

i-colbert:


src/brevitas/graph/gpxq.py
```diff
@@ -232,6 +232,8 @@ def process_input(self, inp):
             batch_dim = inp.names.index('N')
             inp.rename_(None)
             inp = inp.transpose(0, batch_dim)
+
+        inp = self.layer.input_quant(inp)
```
Giuseppe5, line 236: Unfortunately I think we need to do this before we start reshaping and adjusting dim, etc.
In case of activations with per-channel scale factor we could have unexpected behaviours

PR #949, opened 2024-04-30
Fix (llm): fix device issue for eval when not using default device

fabianandresgrob:
This PR fixes a problem when setting a different device than the default. E.g. when `cuda:0` is free, the call to `cuda()` would place the data on that device. However, when we specified the model to be on `cuda:1`, we run into the problem of having the data on a different device than the model. Simply moving the data to `model.device` solves this. For the creation of a validation_dataloader, I've added an argument to solve this issue.

### DIFF
--- a/src/brevitas_examples/llm/main.py
+++ b/src/brevitas_examples/llm/main.py
@@ -278,7 +278,8 @@ def main():
         nsamples=args.nsamples, tokenizer=tokenizer, seqlen=args.seqlen, seed=0)
     val_data = get_wikitext2(
         nsamples=args.nsamples, tokenizer=tokenizer, seqlen=args.seqlen, split='validation', seed=0)
-    val_data = create_validation_dataloader(val_data, args.seqlen, model.device)
+    device = next(iter(model.parameters())).device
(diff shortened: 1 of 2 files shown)


Giuseppe5:
Tests are failing. Is this because of the PR?

Giuseppe5:
Is this still relevant/needed?

fabianandresgrob:
> Is this still relevant/needed?

yes, still needed. We applied it manually on the separate branch

src/brevitas_examples/llm/main.py
```diff
@@ -278,7 +278,7 @@ def main():
         nsamples=args.nsamples, tokenizer=tokenizer, seqlen=args.seqlen, seed=0)
     val_data = get_wikitext2(
         nsamples=args.nsamples, tokenizer=tokenizer, seqlen=args.seqlen, split='validation', seed=0)
-    val_data = create_validation_dataloader(val_data, args.seqlen)
```
Giuseppe5, line 281: device = next(iter(model.parameters())).device

PR #951, opened 2024-05-01
Feat (examples/stable_diffusion): improvements to SD

Giuseppe5:


src/brevitas/graph/gptq.py
```diff
@@ -9,6 +9,9 @@
 from packaging import version
 import torch
 
+from brevitas.graph.calibrate import disable_return_quant_tensor
```
Giuseppe5, line 12: remove

Issue #958, opened 2024-05-23
Mac OSX Tests for `python=3.8` & `torch==1.9.1` fail when installing dependencies


nickfraser:
Noticed in tests for #956, when OS=`macosx-latest`, `python=3.8` & `torch==1.9.1`, `pip` fails find the necessary dependencies (specifically, `torch==1.9.1`). It is particularly weird because manually searching pypi.org shows that this build [does indeed exist](https://pypi.org/project/torch/1.9.1/#files). Specifically, I believe that `torch-1.9.1-cp38-none-macosx_10_9_x86_64.whl` should get installed.
Even more weird, `1.9.0` is suggested as an installation candidate, despite not listing any extra [builds](https://pypi.org/project/torch/1.9.0/#files).

Not sure if there is some temporary bug with Pypi (in which case, this might resolve itself). Or if this build has been [yanked](https://blog.piwheels.org/new-features-deletion-yanking-and-more/) somehow (they don't appear to be!) Alternatively, we could remove this from the test list for Mac OSX, or move back to `torch==1.9.0`.

Example [test output](https://github.com/Xilinx/brevitas/actions/runs/9206590254/job/25324775792?pr=956):

```
nox > Command python -m pip install torch==1.9.1 failed with exit code 1:
ERROR: Could not find a version that satisfies the requirement torch==1.9.1 (from versions: 1.9.0, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, 2.2.0, 2.2.1, 2.2.2, 2.3.0)
ERROR: No matching distribution found for torch==1.9.1
nox > Session tests_brevitas_cpu-3.8(jit_disabled, pytorch_1.9.1) failed.
```

Giuseppe5:
Related to #906

nickfraser:
In fact, a duplicate!

nickfraser:
Alternatively, we can move the tests to `python=3.9`, `python=3.10`, since `python=3.8` is reaching EOL [soon](https://devguide.python.org/versions/).

PR #960, opened 2024-05-23
Fix (test/actions): Excluded `torch==1.9.1`, `platform=macos-latest` tests

nickfraser:
Removed failing MacOSX test. See #906, #958.

### DIFF
--- a/.github/workflows/develop_install.yml
+++ b/.github/workflows/develop_install.yml
@@ -21,8 +21,7 @@ jobs:
 
 
         exclude:
-          - python_version: '3.8'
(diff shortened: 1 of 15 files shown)


Giuseppe5:
Part of me still leans towards completely removing mac-os from the testing suite. We would need to go through the tests to remove any macos specific exception.

We have already several tests disabled only for MacOS and we can't debug if anything goes wrong there.

nickfraser:
We'll disable this for now, and revisit if/when we want to disable all MacOSX test.

PR #964, opened 2024-05-31
Fix (GPFQ): using random projection for speed up/less memory usage

fabianandresgrob:
Using random projection to reduce dimensionality of calibration data that is used for GPFQ `single_layer_update`. 
Reducing the dimensionality offers speed up and less memory being used, so GPFQ becomes feasible for larger models.

### DIFF
--- a/src/brevitas/graph/gpfq.py
+++ b/src/brevitas/graph/gpfq.py
@@ -109,7 +109,7 @@ class gpfq_mode(gpxq_mode):
         # selecting impl of random proj
         self.compression_rate = compression_rate
         if self.compression_rate < 0.0 or self.compression_rate > 1.0:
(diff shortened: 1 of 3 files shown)


src/brevitas/graph/gpfq.py
```diff
@@ -83,6 +106,11 @@ def __init__(
         self.accumulator_bit_width = accumulator_bit_width
         self.a2q_layer_filter_fnc = a2q_layer_filter_fnc  # returns true when to use GPFA2Q
 
+        # selecting impl of random proj
+        self.compression_rate = compression_rate
+        if self.compression_rate < 0.0 or self.compression_rate > 1.0:
+            raise ValueError(f'Compression rate for random projection cannot be {compression_rate}')
```
Giuseppe5, line 112: Cannot be negative

PR #971, opened 2024-06-17
Feat: Support for Groupwise (MX) quantization

Giuseppe5:
This implements:
- New GroupwiseQuantTensor for Int and Float
- Relevant Proxy classes
- MX Float based quantizers
- One notebook to test instantiation and execution 
- Default MXInt quantizers

Missing:
- Export

# Features
## Dynamic expansion of contracted groupwise tensors
Compared to Int/Float QuantTensor, the main difference of their groupwise equivalent is that `value`, `scale`, and `zero_point` are not direct attributes anymore but properties. The new attributes are `value_`, `scale_`, and `zero_point_`.

The reason for this is shaping. When quantizing a tensor with shapes [O, I], where O is output channel and I is input channel, with groupsize k,  groupwise quantization is normally represented as follow:
- Tensor with shapes [O, k, I/k]
- Scales with shapes [O, k, 1]
- Zero point same as scale

The alternative to this representation is to have all three tensors with shapes [O,I], with a massive increase in memory utilization, especially with QAT + gradients.

The underscored attributes will have the compressed shapes, while the properties (non-underscored naming) will dynamically compute the expanded version of the property. This means:
```python
quant_tensor.scale_.shape
# This will print [O, k, 1]
quant_tensor.scale.shape
# This will print [O, I]
```

Internally, the quantization will happen with groupwise shaping (e.g., contracted). For this reason, there is a preliminary view applied to the tensors before everything goes in `tensor_quant`.

## Deprecation of scaling_per_output_channel
Another important change of this PR is the `deprecation` (i.e., still usable but not recommended anymore) of the flag scaling_per_output_channel, in favor of a ternary flag `scaling_per_output` that can be TENSOR/CHANNEL/GROUP.

Lots of work has gone into maintaining retro-compatibility with the existing binary flag, so that everything will still work as intended.
One thing that is still up for discussion is how to handle shared quantizers for groupwise quantization.

## Automatic OCP definition
When instantiating a OCP FP quantizer, NaN/INF encoding are automatically defined through dep inj based on the bitwidths. This is to avoid to have manually to define all possible minifloat quantizers needed for MX (e.g. MX FP8, FP6, etc.)

## Groupwise Default quantizers
Lots of possible quantizers could be defined by default. For integer (not much of a problem) we have groupwise with float scaling (defined in examples if I'm not mistaken) and MX (groupwise with Po2 scale) defined in the brevitas source.

For float it gets a bit more complicated with inf/nan. The solution adopted is the following:
- Float scale will use non-standard floating point representation, meaning no inf/nan representations for all the bitwidths. Defined in examples
- Po2 scale will use OCP standard for FP8/FP6/FP4, so that we have MX Float compliant quantizers defined out of the box. Defined in brevitas source

I only created quantizers with bitwidth equal to 8, e4m3. Overriding the bit_width, mantissa_bit_width and exponent_bit_width will produce OCP compliant MX quantizers (thanks to the change in Automatic OCP definition)

## Example changes

No longer separated flags for ocp/fnuz standard. Now the user should pass float_e4m3 for general fp8 no standard, float_ocp_e4m3 for OCP, float_fnuz_e4m3 for FNUZ.


src/brevitas/quant_tensor/groupwise_float_quant_tensor.py
```diff
@@ -0,0 +1,411 @@
...
+
+    @property
+    def saturating(self):
+        return self.saturating_t.item()
+
+    def __torch_function__(self, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+
+        # new_args = list()
```
Giuseppe5, line 85: Remove

src/brevitas/proxy/groupwise_int_parameter_quant.py
```diff
@@ -0,0 +1,40 @@
...
+    def forward(self, x: torch.Tensor) -> Union[Tensor, GroupwiseIntQuantTensor]:
+        if self.is_quant_enabled:
+            impl = self.export_handler if self.export_mode else self.tensor_quant
+            x = self.view_impl(x)
+            out, scale, zero_point, bit_width = impl(x)
+            return GroupwiseIntQuantTensor(
+                out,
+                scale,
+                zero_point,
+                bit_width,
```
nickfraser, line 34: Shouldn't `bit_width` go _after_ `group_dim`, `group_size`?

src/brevitas/proxy/groupwise_int_parameter_quant.py
```diff
@@ -0,0 +1,40 @@
...
+                self.group_dim,
+                self.group_size,
```
nickfraser, line 36: Also, I think `group_dim`, `group_size` should be reversed.

src/brevitas/proxy/groupwise_int_runtime_quant.py
```diff
@@ -0,0 +1,75 @@
...
+                if isinstance(y, tuple):
+                    y = y[0]
+                if isinstance(x, GroupwiseIntQuantTensor):
+                    out = GroupwiseIntQuantTensor(
+                        y,
+                        x.scale,
+                        x.zero_point,
+                        self.group_dim,
+                        self.group_size,
+                        x.bit_width,
```
nickfraser, line 59: Also here. `bit_width`, `group_dim`, `group_size` in wrong order?

src/brevitas/proxy/groupwise_float_runtime_quant.py
```diff
@@ -0,0 +1,85 @@
...
+                    saturating,
+                    inf_values,
+                    nan_values,
+                    signed=self.is_signed,
+                    training=self.training)
+            elif self.is_passthrough_act:  # preserve scale/zp/bit/sign even without output quant
+                if isinstance(y, tuple):
+                    y = y[0]
+                if isinstance(x, GroupwiseFloatQuantTensor):
+                    out = GroupwiseFloatQuantTensor(
```
nickfraser, line 58: I think the argument order is wrong here too?

`group_dim` <-> `group_size`?
`mantissa_bit_width` <-> `exponent_bit_width`

Last 5 values also seem out-of-order from class definition:
```
class GroupwiseFloatQuantTensorBase(NamedTuple):
    value_: Tensor
    scale_: Tensor
    zero_point_: Tensor
    group_size: Tensor
    group_dim: Tensor
    exponent_bit_width: Tensor
    mantissa_bit_width: Tensor
    exponent_bias: Tensor
    saturating_t: Tensor
    inf_values: List[str]
    nan_values: List[str]
    signed_t: Tensor
    training_t: Tensor
```

src/brevitas/proxy/float_runtime_quant.py
```diff
@@ -109,3 +109,56 @@ def forward(self, x: Union[Tensor, QuantTensor]) -> Union[Tensor, FloatQuantTens
...
+            # If y is an empty FloatQuantTensor, we need to check if this is a passthrough proxy,
+            # otherwise return a simple Tensor
+            # We exclude the last two values (inf_values and nan_values)
+            if isinstance(y, tuple) and not any(map(lambda f: f is None, y[:-2])):
+                out = FloatQuantTensor(*y, signed=self.is_signed, training=self.training)
+            elif self.is_passthrough_act:  # preserve scale/zp/bit/sign even without output quant
+                if isinstance(y, tuple):
+                    y = y[0]
+                if isinstance(x, FloatQuantTensor):
+                    out = FloatQuantTensor(
```
nickfraser, line 139: I also think these are out-of-order. (and also at line 86 in this same file - can't comment there because you didn't change it, but I think worth fixing in this PR anyway)

src/brevitas/quant_tensor/groupwise_float_quant_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+
+        neg_value = (-neg_deq - zp) * scale
+        # In case the dtype of self.minifloat is different from the one of the scale
+        neg_value = neg_value.type(scale.dtype)
+        neg_value = GroupwiseFloatQuantTensor.from_expanded(
+            neg_value, self.group_dim, self.group_size, compress=False)
+        scale = GroupwiseFloatQuantTensor.from_expanded(
+            scale, self.group_dim, self.group_size, compress=True)
+        if self.signed:
+            return GroupwiseFloatQuantTensor(
```
nickfraser, line 260: nitpicky, but maybe worth putting these in the correct order as well (I know it makes no difference), but the other errors may be due to copy/paste from functions like these...

src/brevitas/quant_tensor/groupwise_float_quant_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+
+            # In case the dtype of self.minifloat is different from the one of the scale
+            abs_value = (neg_deq - zp) * scale
+            # In case the dtype of self.minifloat is different from the one of the scale
+            abs_value = abs_value.type(scale.dtype)
+            abs_value = GroupwiseFloatQuantTensor.from_expanded(
+                abs_value, self.group_dim, self.group_size, compress=False)
+            scale = GroupwiseFloatQuantTensor.from_expanded(
+                scale, self.group_dim, self.group_size, compress=True)
+            return GroupwiseFloatQuantTensor(
```
nickfraser, line 313: Likewise, maybe put in correct order?

src/brevitas/quant_tensor/groupwise_int_quant_tensor.py
```diff
@@ -0,0 +1,315 @@
...
+
+        neg_value = (-neg_deq - zp) * scale
+        # In case the dtype of self.minifloat is different from the one of the scale
+        neg_value = neg_value.type(scale.dtype)
+        neg_value = GroupwiseIntQuantTensor.from_expanded(
+            neg_value, self.group_dim, self.group_size, compress=False)
+        scale = GroupwiseIntQuantTensor.from_expanded(
+            scale, self.group_dim, self.group_size, compress=True)
+        if self.signed:
+            return GroupwiseIntQuantTensor(
```
nickfraser, line 250: Arguments are for `GroupwiseFloatQuantTensor`, not `GroupwiseIntQuantTensor`.

src/brevitas/quant_tensor/groupwise_int_quant_tensor.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def __mul__(self, other):
+        if isinstance(other, QuantTensor):
+            return self.value * other.value
+        else:
+            output = self.value * other
+        return output
+
+    def __str__(self):
+        return f"GroupwiseIntQuantTensor(value={self.value}, scale={self.scale}, zero_point={self.zero_point}, bit_width={self.bit_width}, signed_t={self.signed_t}, training_t={self.training_t})"
```
nickfraser, line 281: Add `group_size`, `group_dim`?

src/brevitas/quant_tensor/groupwise_int_quant_tensor.py
```diff
@@ -0,0 +1,315 @@
...
+
+            # In case the dtype of self.minifloat is different from the one of the scale
+            abs_value = (neg_deq - zp) * scale
+            # In case the dtype of self.minifloat is different from the one of the scale
+            abs_value = abs_value.type(scale.dtype)
+            abs_value = GroupwiseIntQuantTensor.from_expanded(
+                abs_value, self.group_dim, self.group_size, compress=False)
+            scale = GroupwiseIntQuantTensor.from_expanded(
+                scale, self.group_dim, self.group_size, compress=True)
+            return GroupwiseIntQuantTensor(
```
nickfraser, line 303: Arguments should be for `GroupwiseIntQuantTensor`.

src/brevitas_examples/common/generative/quant_blocks.py
```diff
@@ -16,79 +16,6 @@
 from brevitas.function.ops_ste import abs_binary_sign_grad
 
 
-class OverSubChannelBlockView(brevitas.jit.ScriptModule):
```
nickfraser, line 19: Are any of these classes used over at `optimum-amd`?

src/brevitas/graph/quantize.py
```diff
@@ -8,6 +8,7 @@
 from brevitas.core.scaling.standalone import ParameterScaling
 from brevitas.fx.brevitas_tracer import symbolic_trace
 from brevitas.graph.base import ModuleToModuleByClass
+from brevitas.graph.calibrate import DisableEnableQuantization
```
nickfraser, line 11: Unused import?

nickfraser (changes requested):
Maybe worth double-checking all of the `<Datatype>QuantTensor` instantiations where the order of the arguments really matter. I found a few errors that are likely due to some original issue + copy/paste, so they may also exist outside your specific changes.

I think it's worth checking double-checking as many as you can while you're fixing the ones I found.

src/brevitas_examples/common/generative/quant_blocks.py
nickfraser: I suspect not. I see you've changed to definition of the quantizers in common/generative to use the new features, so I expect this will work. We should test it though (not for this PR)

Giuseppe5: I think we should be good but at some point the quantizers will move as well..

src/brevitas/graph/quantize.py
Giuseppe5: Removed a whole bunch of them

notebooks/minifloat_mx_tutorial.ipynb
```diff
@@ -63,6 +108,46 @@
...
+    "The underscored attributes will have the compressed shapes, while the properties (non-underscored naming) will dynamically compute the expanded version of the property. This means:\n",
+    "```python\n",
+    "quant_tensor.scale_.shape\n",
+    "# This will print [O, k, 1]\n",
+    "quant_tensor.scale.shape\n",
+    "# This will print [O, I]\n",
+    "```\n",
+    "\n",
+    "With respect to pre-defined quantizers, Brevitas offers several Groupwise and MX options.\n",
+    "The meain difference between the two is that MX is restricted to group_size=32 and the scale factor must be a power-of-2.\n",
```
nickfraser, line 143: meain -> main

PR #977, opened 2024-06-25
Various SDXL quantization fixes

nickfraser:
Adds the following fixes / enhancements:
 - [x] Set the device correct when checkpoints are loaded
 - [x] Fix issue setting the bitwidth for the output of linear layers when SDP is quantized
 - [x] Fix replacing the LoRACompatible linear layers
   - [x] Fix or workaround (i.e., achieve this another way) failing tests
 - [x] Move weight to CPU before 'param_only' export
 - [x] Add explicit pandas dependency
 - [x] Add option to scale & export a whitelist of layers of VAE, to enable FP16 execution of VAE

### DIFF
--- a/src/brevitas_examples/stable_diffusion/mlperf_evaluation/accuracy.py
+++ b/src/brevitas_examples/stable_diffusion/mlperf_evaluation/accuracy.py
@@ -601,7 +601,8 @@ def compute_mlperf_fid(
         path_to_coco,
         model_to_replace=None,
         samples_to_evaluate=500,
-        output_dir=None):
+        output_dir=None,
+        vae_force_upcast=True):
 
     assert os.path.isfile(path_to_coco + '/tools/val2014.npz'), "Val2014.npz file required. Check the MLPerf directory for instructions"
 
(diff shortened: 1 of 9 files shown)


tests/brevitas/graph/test_transforms.py
```diff
@@ -287,6 +287,6 @@ def forward(self, x):
     model = TestModel()
     assert model.conv.stride == (1, 1)
 
-    kwargs = {'stride': lambda module: 2 if module.in_channels == 3 else 1}
+    kwargs = {'stride': lambda module, name: 2 if module.in_channels == 3 else 1}
```
nickfraser, line 290: Check if anything in `brevitas_examples` will be affected by this change.

src/brevitas_examples/stable_diffusion/mlperf_evaluation/accuracy.py
```diff
@@ -615,7 +615,7 @@ def compute_mlperf_fid(
 
     if model_to_replace is not None:
         model.pipe = model_to_replace
-
+    model.pipe.vae.config.force_upcast = True
```
nickfraser, line 618: Only replace UNet and not the entire pipe.

tests/brevitas/graph/test_transforms.py
Giuseppe5: Lambdas gets inspected to decide what's the correct signature to use, so we have backward compatibility

PR #980, opened 2024-07-03
Fix (ptq/bias_correction): remove unnecessary forward pass

Giuseppe5:


src/brevitas/graph/calibrate.py
```diff
@@ -284,6 +284,9 @@ def update_correction(self, name, error):
 
     def apply_correction(self, model):
         for name, module in model.named_modules():
+            if isinstance(module, self.layers):
+                # Re-enable output quantization
+                self.enable_act_quantization(module.output_quant, is_training=False)
```
nickfraser, line 289: Move this to context manager.

nickfraser (changes requested):

Issue #889, opened 2024-03-01
Update entrypoint for LLM


Giuseppe5:
We need to update our entrypoint for LLM to follow the latest features and changes in:

https://github.com/huggingface/optimum-amd/blob/main/examples/quantization/brevitas/quantize_llm.py

nickfraser:
Solved in #987.

PR #987, opened 2024-07-16
Feat: Update LLM entry-point

nickfraser:
Addresses #889. Updated entry-point to leverage many features of our optimum-amd integration effort, as well as update the example to use available quantizers. ~Builds on #977~ Now merged.

Todo:
 - [x] Use dataset utils from optimum-amd
 - [x] ~Update to use fx tracing method from optimum-amd~ Not necessary, already implemented
 - [x] Decompose quantizer generation and model quantization (like SDXL example)
 - [x] Test and fix all pre-quantization and PTQ techniques:
   - [x] `--ln-affine-merge` **fails**
   - [x] `--weight-equalization`
   - [x] `--act-equalization layerwise`
   - [x] `--act-equalization fx`
   - [x] `--bias-corr`
   - [x] `--act-calibration`
   - [x] `--gptq`
 - [x] Test `--replace-mha`
 - [ ] Update interface and add MX datatypes (depends on #971)
 - [ ] ~Allow optional quantization of first (embedded) layer~ Disabled, see [comment](https://github.com/Xilinx/brevitas/pull/987#issuecomment-2296316074)
 - [x] Allow optional quantization of the last layer
 - [x] Test various export flows:
   - [x] ONNX QCDQ
   - [x] torch QCDQ
   - [ ] ~torchmlir~ fails, won't fix in this PR
   - [ ] ~torchmlir (packed weights)~ fails, won't fix in this PR

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/bias_corr.py
+++ b/src/brevitas_examples/llm/llm_quant/bias_corr.py
@@ -4,6 +4,7 @@ Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 """
 
 import torch
+from tqdm import tqdm
 
 from brevitas.graph.calibrate import bias_correction_mode
 
@@ -11,5 +12,5 @@ from brevitas.graph.calibrate import bias_correction_mode
 @torch.no_grad()
 def apply_bias_correction(model, dataloader):
     with bias_correction_mode(model):
-        for inps in dataloader:
+        for inps in tqdm(dataloader):
             model(**inps)
--- a/src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
+++ b/src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
@@ -1,5 +1,7 @@
 import warnings
 
+import torch
+
 from transformers.models.opt.modeling_opt import OPTAttention
 
 from brevitas.graph import ModuleToModuleByClass
--- a/src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
+++ b/src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
@@ -21,3 +21,17 @@ def replace_mha_with_quantizable_layers(model, dtype):
     for rewriter in rewriters:
         model = rewriter.apply(model)
     return model
+
+
+@torch.no_grad()
+def add_zero_bias_to_linear(model: torch.nn.Module) -> torch.nn.Module:
+    for name, module in model.named_modules():
+        if type(module) == torch.nn.Linear:
+            if module.bias is None:
+                module.register_parameter(
+                    "bias",
+                    torch.nn.Parameter(
+                        torch.zeros((module.weight.shape[0],), device=module.weight.device, dtype=module.weight.dtype)
+                    ),
+                )
+    return model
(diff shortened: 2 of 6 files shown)


nickfraser:
@Giuseppe5, I've removed the quant_embedding support. It is currently broken since channelwise scaling is not supported for `QuantEmbedding` layers, but our weight quantizers all use channelwise scaling for the generative examples.

Quantizing the embedding seems to have limited utility anyway because:
 - If "input quantization" is enabled, the linear layers in the first decoder layer are quantized anyway
   - It only adds quantization of the first residual path on the first attention layer
 - If input/embedding quantization is enabled, re-quantisation may occur at the first attention layer
 - The storage benefits from quantizing the embedded lookup are usually minimal

src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
```diff
@@ -21,3 +22,18 @@ def replace_mha_with_quantizable_layers(model, dtype):
     for rewriter in rewriters:
         model = rewriter.apply(model)
     return model
+
+
+@torch.no_grad()
+def add_zero_bias_to_linear(model: torch.nn.Module) -> torch.nn.Module:
```
Giuseppe5, line 28: Is this here for loading checkpoint + bias correction?
We have a context manager for that now (load_quant_model in graph/calibrate)

Giuseppe5: Ah no it's for accelerate compatibility, nevermind.

src/brevitas_examples/llm/llm_quant/bias_corr.py
```diff
@@ -4,12 +4,13 @@
...
+from tqdm import tqdm
 
 from brevitas.graph.calibrate import bias_correction_mode
 
 
 @torch.no_grad()
 def apply_bias_correction(model, dataloader):
     with bias_correction_mode(model):
-        for inps in dataloader:
+        for inps in tqdm(dataloader):
```
Giuseppe5, line 15: I think we need to add tqdm as required dependency (in examples requirements but at this point maybe everywhere)

src/brevitas_examples/llm/main.py
```diff
@@ -174,6 +182,15 @@
         'sharded_torchmlir_group_weight',
         'sharded_packed_torchmlir_group_weight'],
     help='Model export.')
+parser.add_argument(
+    '--checkpoint-name',
+    type=str,
+    default=None,
+    help="Filename to save checkpoint. If `None`, no checkpoint is saved (default: %(default)s)")
+add_bool_arg(
+    parser, 'use-ocp', default=False, help='Use OCP format for float quantization. Default: False')
```
Giuseppe5, line 191: Let's merge this and then I'll update the entrypoint here same style as the stable diffusion one in #971

Giuseppe5 (approved):

src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
nickfraser: No, this is to make bias correction work with accelerate properly. Can this also be handled by the context manager you mentioned?

nickfraser:
There currently is no requirements file for the LLM example. I'm adding one in #1002.

Issue #986, opened 2024-07-15
BrevitasQuantProxyHandler args passing may be sometimes wrong


andrei-stoian-zama:
I have a QuantLinear that I instantiate like so:

```
            layer = qnn.QuantLinear(
                in_features,
                out_features,
                True,
                weight_bit_width=n_w_bits,
                bias_quant=None,
                weight_narrow_range=False,
                narrow_range=False,
                signed=True,
                weight_quant=Int8WeightPerTensorFloat
            )
```

I export the model with `BrevitasOnnxManager`:

```

        BrevitasONNXManager.export(
            self.base_module,
            input_shape=X[[0], ::].shape,
            export_path=str(output_onnx_file_path),
            keep_initializers_as_inputs=False,
            opset_version=OPSET_VERSION_FOR_ONNX_EXPORT,
        )
```

when the weights are exported, the `BrevitasQuantProxyHandler` is called:

```
class BrevitasQuantProxyHandler(ONNXBaseHandler, ABC):
    def symbolic_execution(self, x: Tensor):
        scale = self.symbolic_kwargs['scale']
        zero_point = self.symbolic_kwargs['zero_point']
        bit_width = self.symbolic_kwargs['bit_width']
        if bit_width == 1:
            x = BrevitasBinaryQuantFn.apply(x, *self.symbolic_kwargs.values())
        else:
            x = BrevitasQuantFn.apply(x, *self.symbolic_kwargs.values())
        return x, scale, zero_point, bit_width
```

this in turn calls 

```
class BrevitasQuantFn(Function):

    @staticmethod
    def forward(ctx, x, scale, zero_point, bit_width, narrow_range, signed, rounding_mode):
        float_to_int_impl = solve_float_to_int_impl_from_enum(rounding_mode)
        quant = IntQuant(
            float_to_int_impl=float_to_int_impl(),
            tensor_clamp_impl=TensorClamp(),
            narrow_range=narrow_range,
            signed=signed)
        y = quant(scale, zero_point, bit_width, x)
        return y
```

In `BrevitasQuantFn` the arguments to forward are positional, but these arguments are given as `*self.symbolic_kwargs.values()` where `symbolic_kwargs` is a dictionary. 

I think there is an issue with the order of arguments as sometimes, while I set `signed=True/narrow=False` everywhere, the exported onnx shows the opposite: 

![image](https://github.com/user-attachments/assets/874330c3-e09e-47ac-897b-c181651e087d)

[...]

PR #988, opened 2024-07-16
Fix (export/qonnx): Fixed symbolic kwargs order.

nickfraser:
Fixed the order of `narrow_range` and `signed` in QONNX export. Addresses #986.


Giuseppe5 (approved):

andrei-stoian-zama (commented):
this seems o fix the bug!!

PR #1001, opened 2024-08-19
Fix (example/llm): Fix API for newer `torch_mlir` releases.

nickfraser:
Still doesn't work, but gets around some API issues.


Giuseppe5:
Probably we need to deprecate all that part tbh

PR #1002, opened 2024-08-19
Test (example/llm): Refactor and add basic tests for the LLM entry-point

nickfraser:
Create a tests which fully test the end-to-end examples, since the stability of these examples is becoming increasingly important. The idea is to put into a place a structure that can be applied to other entry-points in `brevitas_examples`. ~Incorporates #987~ - Now rebased on `dev`.

TODO:
 - [x] Move argument parser to a separate function, to allow calling from tests
 - [x] Update main function to take parsed arguments and an input, return useful output that we want to test as output
 - [x] Create run tests for tiny versions of important models:
   - [x] Llama
   - [x] OPT
   - [x] Mistral
 - [x] Create run tests on all pre-quantization and PTQ techniques on all selected tiny models:
   - [x] `--ln-affine-merge`
   - [x] `--weight-equalization`
   - [x] `--act-equalization layerwise`
   - [x] `--act-equalization fx`
   - [x] `--bias-corr`
   - [x] `--act-calibration`
   - [x] `--gptq`
 - [x] Test `--replace-mha` on OPT
 - [x] Test some important datatypes:
   - [x] Weight-only (int4)
   - [x] Int8
   - [x] FP8
   - [x] MXFP8 **(fails)**
 - [x] Create accuracy tests for tiny versions of important models (should cover all models, datatypes and PTQ techniques):
   - [x] Llama
   - [x] OPT
   - [x] Mistral
 - [x] Create tests for export flows
   - [x] ONNX QCDQ
   - [x] Torchscript QCDQ
   - [ ] ~torchmlir~ fails and won't fix
   - [ ] ~torchmlir packed weights~ fails and won't fix
 - [x] Added test to github actions


noxfile.py
```diff
@@ -105,7 +124,25 @@ def tests_brevitas_examples_cpu(session, pytorch, jit_status):
     install_pytorch(pytorch, session)
     install_torchvision(pytorch, session)  # For CV eval scripts
     session.install('--upgrade', '.[test, tts, stt, vision]')
-    session.run('pytest', '-n', 'logical', 'tests/brevitas_examples')
+    session.run(
+        'pytest',
+        '-n',
+        'logical',
+        '--ignore-glob',
```
Giuseppe5, line 131: what's this ignore-glob?

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+            name="hf-internal-testing/tiny-random-LlamaForCausalLM",
+            float_ppl=None,
+            supports_fx=True,
+        ),
+        ModelAndPpl(
+            name="hf-internal-testing/tiny-random-MistralForCausalLM",
+            float_ppl=None,
+            supports_fx=False,
+        ),
+        #ModelAndPpl(
```
Giuseppe5, line 82: Why is this (and later) commented?

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+    #args.checkpoint = ptid2pathname(request.node.nodeid) + ".pth" # Example filename which won't clash
+    args.export_prefix = ptid2pathname(request.node.nodeid)
+    args.weight_bit_width = 8
+    args.weight_quant_granularity = "per_channel"  # "per_tensor", "per_channel", "per_group".
+    args.input_bit_width = 8
+    args.act_calibration = True
+    return args
+
+
+@pytest.fixture(
```
Giuseppe5, line 109: I am making an effort to use only pytest_cases fixtures (generally more powerful), but not too big of a problem, in most cases they are interchangeable.

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+    params=[
+        {},
+        {
+            "bias_corr": True},
```
Giuseppe5, line 113: The indentation is confusing to me

Giuseppe5: It could be worth giving names to all the parameters otherwise they appear as 
`toggle_run_args0`, `toggle_run_args1`, which makes debugging the tests complicated

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+            name="hf-internal-testing/tiny-random-OPTForCausalLM",
+            float_ppl=None,
+            supports_fx=True,
+        ),])
+def small_models_with_ppl_pt_ge_2_4(request):
+    yield request.param
+
+
+@pytest.mark.llm
+@requires_pt_ge('2.4')
```
Giuseppe5, line 156: Add a comment about why this is needed

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+    float_ppl = run_dict["float_ppl"]
+    quant_ppl = run_dict["quant_ppl"]
+    del run_dict["float_ppl"]
+    del run_dict["quant_ppl"]
+    args.update(**run_dict)
+    yield args, float_ppl, quant_ppl
+
+
+@pytest.mark.llm
+@requires_pt_ge('2.2')
```
Giuseppe5, line 195: As before, why only 2.2 and up?

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+@requires_pt_ge('2.4')
+def test_small_models_toggle_run_args_pt_ge_2_4(
+        caplog, toggle_run_args, small_models_with_ppl_pt_ge_2_4):
+    caplog.set_level(logging.INFO)
+    args = toggle_run_args
+    args.model = small_models_with_ppl_pt_ge_2_4.name
+    exp_float_ppl = small_models_with_ppl_pt_ge_2_4.float_ppl
+    use_fx = requires_fx(args)
+    if use_fx and not small_models_with_ppl_pt_ge_2_4.supports_fx:
+        pytest.xfail(f"{small_models_with_ppl.name} does not support FX")
```
Giuseppe5, line 165: This will never xfail currently, correct? Can we add a test case to stress this or just remove it?

tests/brevitas_examples/test_llm.py
```diff
@@ -0,0 +1,442 @@
...
+
+@pytest.mark.llm
+@requires_pt_ge('2.2')
+def test_small_models_acc(caplog, acc_args_and_acc):
+    caplog.set_level(logging.INFO)
+    args, exp_float_ppl, exp_quant_ppl = acc_args_and_acc
+    float_ppl, quant_ppl, model = main(args)
+    float_ppl = float_ppl.detach().cpu().numpy()
+    quant_ppl = quant_ppl.detach().cpu().numpy()
+    if config.JIT_ENABLED:
```
Giuseppe5, line 202: Why only with JIT_ON?

noxfile.py
nickfraser: I'm using it to ignore `test_llm.py`. Otherwise the file gets loaded for the other tests and as a result `optimum-amd`, `transformers` and other packages become a dependency for all `brevitas_examples` tests.

Open to other options too, but this seemed like the best at the time.

tests/brevitas_examples/test_llm.py
nickfraser: At the time, the JIT changes the accuracy enough, but I didn't want to loosen the accuracy bounds. I will double-check this, because I had to loosen them for other PyTorch versions anyway...

tests/brevitas_examples/test_llm.py
nickfraser: `optimum-amd[brevitas]` requires PyTorch>=2.2

tests/brevitas_examples/test_llm.py
nickfraser: I'll refactor this to share the same code with the `2.2` one. It makes more sense with that context.

tests/brevitas_examples/test_llm.py
nickfraser: > The indentation is confusing to me
What you see is what I was able to get to pass our `pre-commit`. I'm open to alternative suggestions.

tests/brevitas_examples/test_llm.py
nickfraser: I haven't used `pytest_cases` - I'll look into it.

tests/brevitas_examples/test_llm.py
nickfraser: Test case ready for when/if we expand to MoE support in the LLM entry-point.

tests/brevitas_examples/test_llm.py
nickfraser: Removed this. Thanks!

PR #1005, opened 2024-08-22
Fix (mx): input view during quantization

Giuseppe5:


src/brevitas/core/quant/int.py
```diff
@@ -153,6 +155,7 @@ def forward(self, x: Tensor) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
         int_threshold = self.int_scaling_impl(bit_width)
         scale = threshold / int_threshold
         zero_point = self.zero_point_impl(x, scale, bit_width)
+        x = self.input_view_impl(x)
         y = self.int_quant(scale, zero_point, bit_width, x)
         return y, scale, zero_point, bit_width
```
nickfraser, line 160: Maybe I'm missing something, but shouldn't `y` be returned to the original shape before returning?

Giuseppe5: No the idea is that The groupwise quant tensors will hold the reshaped version in memory, and generate the "original" version when calling `.value`

nickfraser:
LGTM!

PR #1006, opened 2024-08-23
Alternative approach to support torch.compile

Giuseppe5:
This works by assuming that most of the quantization process has already taken place, and it is no longer needed to propagate QuantTensors.

Typical usage:
```python
with inference_mode(model):
     model(input)
     compile_model = torch.compile(model, fullgraph=True)

```


src/brevitas/proxy/runtime_quant.py
```diff
@@ -173,20 +167,19 @@ def forward(self, x: Union[Tensor, QuantTensor]) -> Union[Tensor, IntQuantTensor
...
-                    y = y[0]
-                if isinstance(x, IntQuantTensor):
-                    out = IntQuantTensor(
-                        y, x.scale, x.zero_point, x.bit_width, x.signed, self.training)
-                else:
-                    out = y
+            if torch.compiler.is_compiling():
+                out = y[0]
             else:
-                if isinstance(y, tuple):
```
Giuseppe5, line 187: Restore

nickfraser (changes requested):
This needs tests!

src/brevitas/core/function_wrapper/clamp.py
```diff
@@ -113,40 +113,51 @@ def __init__(
...
+            x[inf_mask] = torch.tensor(float('nan'))
+        else:
+            raise RuntimeError(
+                "Clamping is not saturating, but neither `inf_values` nor `nan_values` is specified"
+            )
+        return x
+
+    def saturating_clamp(self, x, max_value, min_value):
+        return self.tensor_clamp_impl(x, min_val=min_value, max_val=max_value)
+
```
nickfraser, line 138: This is purely a bugfix for our float implementation, right? Not related to `torch.compile`

src/brevitas/export/inference/handler.py
```diff
@@ -0,0 +1,153 @@
...
+    def prepare_for_export(self, module):
+        if module.is_quant_enabled:
+            self.scale = module.scale()
+            self.zero_point = module.zero_point().to(self.scale.device)
+            self.bit_width = module.bit_width()
+            self.min_clamp = min_int(module.is_signed, module.is_narrow_range, self.bit_width)
+            self.max_clamp = max_int(module.is_signed, module.is_narrow_range, self.bit_width)
+
+    def quantize(self, x):
+        return torch.clamp(
```
nickfraser, line 55: It looks like these won't work with Groupwise quantization, correct? So `inference_mode` + MX won't work?

src/brevitas/proxy/float_runtime_quant.py
```diff
@@ -27,7 +28,7 @@ def mantissa_bit_width(self, force_eval=True):
     def exponent_bias(self, force_eval=True):
         return self.retrieve_attribute('exponent_bias', force_eval)
 
-    def saturating(self, force_eval=True):
+    def is_saturating(self, force_eval=True):
```
nickfraser, line 31: Name change just to match convention?

tests/brevitas_end_to_end/test_torchvision_models.py
```diff
@@ -113,15 +114,37 @@ def torchvision_model(model_name, quantize_fn):
...
+        prehook_non_compiled_out = torchvision_model(inp)
+        post_hook_non_compiled_out = torchvision_model(inp)
+        assert torch.allclose(prehook_non_compiled_out, post_hook_non_compiled_out)
+
+        if enable_compile:
+            compiled_model = torch.compile(torchvision_model, fullgraph=True)
+            compiled_out = compiled_model(inp)
+
+        # This fails! Compile might needs more small-scoped tests for accuracy evaluation
+        # assert torch.allclose(post_hook_non_compiled_out, compiled_out)
```
nickfraser, line 145: How badly does it fail? I'd prefer to keep this check with looser bounds if the errors are reasonable...

src/brevitas/core/function_wrapper/clamp.py
Giuseppe5: Not bug fix, more of an utility so that I can independently call these functions in a standalone way (as we do now during inference mode)

src/brevitas/export/inference/handler.py
Giuseppe5: You're right, I forgot to add the export handler for MX INT and MX Float

src/brevitas/proxy/float_runtime_quant.py
Giuseppe5: Yup

tests/brevitas_end_to_end/test_torchvision_models.py
Giuseppe5: ~0.2 absolute error

src/brevitas/export/inference/handler.py
Giuseppe5: Postponed to another update

tests/brevitas_end_to_end/test_torchvision_models.py
Giuseppe5: I will add whatever is needed to make the tests pass

Giuseppe5: I added the tolerance and tested locally, but the tests will still not run on the CI because `torch.compile` works only on torch 2.2

nickfraser:
LGTM!

PR #1007, opened 2024-08-23
Feat (mx): adding padding and transposed support

Giuseppe5:
This depends on #1005 
It does padding left as per standard of MX implementation.

### DIFF
--- a/src/brevitas/quant/solver/parameter.py
+++ b/src/brevitas/quant/solver/parameter.py
@@ -111,16 +111,17 @@ class SolveParameterScalingImplFromEnum(SolveAffineRescalingFromEnum):
 class SolveParameterScalingShape(ExtendedInjector):
 
     @value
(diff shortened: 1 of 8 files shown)


src/brevitas/quant/solver/parameter.py
```diff
@@ -111,36 +111,44 @@ def scaling_impl(scaling_impl_type):
...
+        size.insert(group_dim + 1, group_size)
         return size
 
+    @value
+    def padding(module, group_dim, group_size):
+        padding = [0, 0] * len(module.weight.shape)
+        size = list(module.weight.shape)
+        if size[group_dim] % group_size != 0:
+            padding[2 * group_dim] = group_size - size[group_dim] % group_size
+        padding = list(reversed(padding))
```
Giuseppe5, line 145: Add comment

PR #1009, opened 2024-08-26
Feat (gptq): optimizing CPU to GPU memory transfer

i-colbert:


nickfraser:
I'm slightly worried about this messing with our interop with HuggingFace `accelerate`. Would you test this in a multi-GPU setup with accelerate? Easiest way is to use this: https://github.com/huggingface/optimum-amd/tree/main/examples/quantization/brevitas

nickfraser:
I ran a small multi-GPU test with `accelerate` and this seems to work.

PR #1011, opened 2024-08-28
Fix (proxy): clean-up

Giuseppe5:


src/brevitas/proxy/parameter_quant.py
```diff
@@ -92,25 +110,39 @@ def tracked_parameter_list(self):
...
         return False
 
+    @abstractmethod
+    def create_quant_tensor(self, qt_args: Tuple[Any]) -> Union[Tensor, QuantTensor]:
+        raise NotImplementedError
+
+    def forward(self, x: torch.Tensor) -> Union[Tensor, QuantTensor]:
+        if self.is_quant_enabled:
+            if self._cached_weight is not None and not self.cache_inference_quant_weight_metadata_only:
+                out = self._cached_weight.quant_tensor
```
nickfraser, line 120: What if we're in export mode?

nickfraser (changes requested):
Fix export mode for weights. I might have a think about the act proxy forward pass logic.

Otherwise, LGTM!

src/brevitas/proxy/runtime_quant.py
```diff
@@ -160,21 +168,24 @@ def forward(self, x: Union[Tensor, QuantTensor]) -> Union[Tensor, QuantTensor]:
...
-            y = y[0]
+            # preserve quant_metadata
+            if isinstance(y, tuple):
+                y = y[0]
             out = self.create_quant_tensor(y, x=x)
         else:
-            out = y[0]
+            if isinstance(y, tuple):
+                y = y[0]
+            out = y
```
nickfraser, line 188: I'm struggling with these `tuple` checks. Under what conditions can `y` be an "empty `QuantTensor`"? As far as I can tell, it must be generated by one of the following function calls above:

```python
# Line 166-167
y = self.fused_activation_quant_proxy.activation_impl(y) # Either returns `Tensor` or `QuantTensor`?
y = self.export_handler(y) # returns `Tensor` or `tuple`? Can it be an "empty `QuantTensor`"?
# Line 171
y = self.fused_activation_quant_proxy.activation_impl(y) # returns `Tensor` or `QuantTensor`?
# Line 173
y = self.fused_activation_quant_proxy(y) # returns non-empty `tuple`?
```

If `y` cannot be an empty `tuple`, then these checks should be removed. If `y` can be an empty `QuantTensor`, then I prefer the previous code.

nickfraser (commented):
See note about the `tuple` checks.

src/brevitas/proxy/groupwise_int_parameter_quant.py
```diff
@@ -14,7 +22,7 @@ def group_dim(self):
     def group_size(self):
         return self.quant_injector.group_size
 
-    def create_quant_tensor(self, qt_args: List[Any]) -> GroupwiseIntQuantTensor:
+    def create_quant_tensor(self, qt_args: Tuple[Any]) -> GroupwiseIntQuantTensor:
```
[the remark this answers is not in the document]
Giuseppe5: No in the case of parameter, the input is always only a tuple

src/brevitas/proxy/runtime_quant.py
Giuseppe5: Let's define `full` a tuple where there are no None, and `empty` a tuple where only the first value is not None.
In all functions above, the input is always a Tensor, the output is either a Tensor or a Tuple, never a QuantTensor (which is created only later).

```python
y = self.fused_activation_quant_proxy.activation_impl(y)
```
Here, the output is always a Tensor since activation_impl is a normal pytorch function

```python
y = self.export_handler(y)
```
Output is always a full tuple

```python
y = self.fused_activation_quant_proxy(y)
```
Output can be a full or empty tuple, based on whether the quantizer passed to ActQuantProxy is None or not (e.g., QuantReLU(act_quant=None)

So we can have all cases (full tuple, empty tuple, single tensor) when we reach the if/else part

PR #1012, opened 2024-08-28
 Feat (mx): automatic group_dim in layerwise quant

Giuseppe5:
Depends on #1011

PR #1013, opened 2024-08-28
Feat (mx): gptq compatibility and quant tests

Giuseppe5:
Depends on #1011 and #1012

Added tests for MX and FloatQuant, with some restrictions;
- Neither support external bias quantization at the moment
- Compatibility with MHA is limited due to how the various quant options interact with each-other
- MX quantizers need to be paired to account for the possibility of padding (to be revisited post-release)
- MX quantizers are not JIT compatible

### DIFF
--- a/src/brevitas/core/function_wrapper/shape.py
+++ b/src/brevitas/core/function_wrapper/shape.py
@@ -16,6 +16,7 @@ from brevitas.function.shape import over_batch_over_tensor
 from brevitas.function.shape import over_output_channels
 from brevitas.function.shape import over_output_features
 from brevitas.function.shape import over_tensor
+from brevitas.utils.torch_utils import padding
 
 
 class PermuteDims(brevitas.jit.ScriptModule):
@@ -162,17 +163,10 @@ class OverSubChannelBlockView(brevitas.jit.ScriptModule):
         self.group_dim = group_dim
         self.group_size = group_size
 
-    def padding(self, x):
-        padding = [0, 0] * len(x.shape)
-        size = x.shape
-        if size[self.group_dim] % self.group_size != 0:
-            padding[2 * self.group_dim] = self.group_size - size[self.group_dim] % self.group_size
-        padding = list(reversed(padding))
-        return padding
-
     @brevitas.jit.script_method
     def forward(self, x: torch.Tensor):
-        y = torch.nn.functional.pad(x, self.padding(x), mode='constant', value=0)
+        y = torch.nn.functional.pad(
+            x, padding(x, self.group_size, self.group_dim), mode='constant', value=0)
         y = y.view(self.expanded_groupwise_shape)
         return y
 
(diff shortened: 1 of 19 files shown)


src/brevitas/core/function_wrapper/shape.py
```diff
@@ -154,16 +154,25 @@ def forward(self, x: torch.Tensor):
...
 
-    def __init__(self, expanded_scaling_shape, padding) -> None:
+    def __init__(self, expanded_scaling_shape, group_size, group_dim) -> None:
         super(OverSubChannelBlockView, self).__init__()
         self.expanded_scaling_shape = expanded_scaling_shape
-        self.padding = padding
+        self.group_dim = group_dim
+        self.group_size = group_size
+
+    def padding(self, x):
```
nickfraser, line 165: Let's share this with `DynamicOverSubChannelBlockView`

src/brevitas/quant/solver/parameter.py
```diff
@@ -111,30 +111,39 @@ def scaling_impl(scaling_impl_type):
...
+        size = list(weight_shape)
         size[group_dim] = (size[group_dim] + group_size - 1) // group_size
         size.insert(group_dim + 1, group_size)
-        return size
+        return tuple(size)
+
+    @value
+    def weight_shape(tracked_parameter_list):
+        module = tracked_parameter_list[0]
+        return tuple(module.shape)
```
nickfraser, line 146: Why only use the first weight in the tracked list.

Giuseppe5: This has been reworked so that everything is less implicit. There are still a few None that are sent propagated during dep inj but sooner or later there is an assert to make sure they are defined if they are needed

src/brevitas/proxy/groupwise_int_parameter_quant.py
```diff
@@ -14,6 +14,11 @@ def group_dim(self):
     def group_size(self):
         return self.quant_injector.group_size
 
+    def apply_input_view(self, x):
```
nickfraser, line 17: Worth making a GroupwiseQuantProxyMixin which provides these functions? Looks like that a lot is shared between `GroupwiseWeightQuantProxyFromInjector` & `GroupwiseActQuantProxyFromInjector`...

src/brevitas/graph/gpxq.py
```diff
@@ -232,7 +238,8 @@ def process_input(self, inp):
         if isinstance(inp, IntQuantTensor):
             if is_quant_enabled and self.quant_metadata is None:
                 self.quant_metadata = _CachedIO(inp, metadata_only=True)
-            inp = inp.value
+            if isinstance(inp, QuantTensor):
```
nickfraser, line 241: When could this not be `True`? I see you've added this check, but it's not obvious to me. Can `inp` get modified in-place by `_CachedIO`?

nickfraser (changes requested):
Two small comments, otherwise LGTM!

nickfraser:
Rebased after #1012 was merged.

src/brevitas/graph/gpxq.py
nickfraser: I think the answer to my question is "no", so I've removed that check in [`1f0933f`](https://github.com/Xilinx/brevitas/pull/1013/commits/1f0933f472480659e2b1ece2908728da969a998d).

src/brevitas/proxy/groupwise_int_parameter_quant.py
nickfraser: After further review, I seem to recall there was some issue with Proxy Mixins (perhaps with dependency injection) results in the ExportMixin needing to be in a certain location. I'll leave this for now - we can revisit at another time.

nickfraser:
The failing tests seem to be a github actions issue rather than a real issue. I can run the failing test suite (`nox -s "tests_brevitas_cpu-3.9(jit_disabled, pytorch_2.1.0)"`) locally on a windows machine.

Merging, and will open another issue if this becomes a regular thing.

PR #1014, opened 2024-08-29
fix (nn/avg_pool): Fix for trunc quant not being applied

nickfraser:
Trunc quant wasn't being applied for TruncAvgPool2d since #945. Also removes unnecessary check in `TruncAvgPool2d` & `TruncAdaptiveAvgPool2d`.


Giuseppe5:
Let's open an issue to add this missing test, otherwise it looks good!

Giuseppe5 (approved):

PR #1025, opened 2024-09-12
Decoupled PerChannel/PerTensor quantization

Giuseppe5:
To have a decouple per channel pre-scale, and per tensor post scale:
```python
class Int8WeightNormL2PerChannelPerTensorFixedPoint(Int8WeightNormL2PerChannelFixedPoint):
    scaling_per_output_type = ScalingPerOutputType.TENSOR

```

It currently works for all quantizers that inherit from `WeightNormPerChannelFloatDecoupled`

It can be extended to other decoupled quantization quantizers if needed

### DIFF
--- a/tests/brevitas/export/quant_module_fixture.py
+++ b/tests/brevitas/export/quant_module_fixture.py
@@ -7,6 +7,7 @@ from pytest_cases import set_case_id
 import torch
 from torch import nn
 
+from brevitas.inject.enum import ScalingPerOutputType
 from brevitas.nn import QuantConv1d
 from brevitas.nn import QuantConv2d
 from brevitas.nn import QuantConv3d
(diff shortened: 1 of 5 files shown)


i-colbert (approved):

tests/brevitas/export/quant_module_fixture.py
nickfraser, line 1: This is a code style question, any reason why we don't use the standard `pytest` way of sharing fixtures across multiple files (i.e., [`conftest.py`](https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files))?

nickfraser (approved):
Single question about test organisation (specifically, fixtures), otherwise LGTM!

tests/brevitas/export/quant_module_fixture.py
Giuseppe5: We have a lot of duplication for various reasons and in general tests would need to be restructured a bit.
All this to say that I wouldn't be sure at his point of the cleanest way to do it myself, but if you have a suggestion, all ears.

Issue #1027, opened 2024-09-19
Bug in float quant proxy properties is_fnuz and is_ocp


alexredd99:
The following lines should be checking for `mantissa_bit_width = 2`  and `exponent_bit_width = 5` when determining if an FP8E5M2 minifloat is OCP or FNUZ.

https://github.com/Xilinx/brevitas/blob/3a9bcc69b446964719ed0d19657e4e1d99476e64/src/brevitas/proxy/float_parameter_quant.py#L69
https://github.com/Xilinx/brevitas/blob/3a9bcc69b446964719ed0d19657e4e1d99476e64/src/brevitas/proxy/float_parameter_quant.py#L81
https://github.com/Xilinx/brevitas/blob/3a9bcc69b446964719ed0d19657e4e1d99476e64/src/brevitas/proxy/float_runtime_quant.py#L44
https://github.com/Xilinx/brevitas/blob/3a9bcc69b446964719ed0d19657e4e1d99476e64/src/brevitas/proxy/float_runtime_quant.py#L56

PR #1028, opened 2024-09-19
Fix (proxy): fix for float quant properties is_ocp and is_fnuz

alexredd99:
Fixes issue #1027 and float quant proxies now check that `mantissa_bit_width = 2` and `exponent_bit_width = 5` when determining if an FP8E5M2 minifloat is OCP or FNUZ.


alexredd99:
Should we add tests for these properties? If so, where would be a good place to add them?

Giuseppe5:
We could add them with the export tests in ORT since these properties are _mostly_ useful for that

Giuseppe5:
Thanks for spotting and fixing!

PR #1035, opened 2024-09-27
Fix (llm): small fixes to LLM

Giuseppe5:
- Improve HQO implementation
- Support for MSE with groupwise quantization
- Add possibility to specify groupdim for groupwise weight quantization

### DIFF
--- a/src/brevitas/core/function_wrapper/shape.py
+++ b/src/brevitas/core/function_wrapper/shape.py
@@ -165,6 +165,11 @@ class OverSubChannelBlockView(brevitas.jit.ScriptModule):
 
     @brevitas.jit.script_method
     def forward(self, x: torch.Tensor):
+        # This one is a bit tricky but we could end up here:
+        # - If we quantize the zero point, which will already have expanded shape matching the scale (although no padding, but we don't need the padding)
+        # - Groupwise HQO quantization, where weight will already have been padded and expanded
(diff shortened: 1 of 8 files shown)


src/brevitas/core/function_wrapper/shape.py
```diff
@@ -165,6 +165,11 @@ def __init__(self, expanded_groupwise_shape, group_size, group_dim) -> None:
 
     @brevitas.jit.script_method
     def forward(self, x: torch.Tensor):
+        # This one is a bit tricky but we could end up here:
+        # - If we quantize the zero point, which will already have expanded shape matching the scale (although no padding, but we don't need the padding)
+        # - Groupwise HQO quantization, where weight will already have been padded and expanded
+        if len(x.shape) == len(self.expanded_groupwise_shape):
+            return x
```
nickfraser, line 172: Weird stuff / comments like this make me wonder if we need to re-think our implementation.

nickfraser (approved):
LGTM!

src/brevitas/core/function_wrapper/shape.py
nickfraser: (but let's not block this release)

Giuseppe5: Agreed

PR #1037, opened 2024-10-01
Feat (core): use runtime parameter for scale

Giuseppe5:
It should help with the implementation of quantized scales since we can rely on the runtime value to compute scale rather than used the tracked parameter.

A bit more forward looking, this is a more """functional""" implementation of quantization, more similar to what happens with activations.
If we assume that group_dim can also become a runtime parameter, this could help towards having a """functional""" weight quantizer that it is not attached to any specific layer
@nickfraser


nickfraser:
LGTM! Once the conflicts are fixed and the tests pass, this can be merged.

nickfraser (approved):

PR #1042, opened 2024-10-04
Fix/Feat (trunc avg pool): Update truncation and average pool behaviour

nickfraser:
# Proposal: Changes to `TruncQuant` and `TruncAvgPool`

Given that the behaviour of `TruncQuant` and `TruncAvgPool` have existed for many years, and some examples do rely on them, I think it's worth opening this to be a discussion with the community.

Note, as far as I can tell, the only place `TruncQuant` is used with `TruncAvgPool` and `TruncQuantAccumulator`, but I think the issues can be understood through the lens of `TruncAvgPool`.

## Motivation

I find the current implementation of `TruncQuant` and `TruncAvgPool` to be a little troubling and **incorrect** outside of a specific use-case. I'll explain further down in this document.

## Current Implementation / Status

Before explaining my issues with the current implementation, let me first explain how `TruncQuant` and `TruncAvgPool` currently work.

### Current `TruncQuant` Implementation

`TruncQuant` shifts the integer representation of the input left or right (i.e., multiply by a power-of-two) proportional to the difference in the input and output bit width. That new integer value is then reinterpreted using the same scale as the input.

Example:

```python
import torch
from brevitas.core.function_wrapper import RoundSte
from brevitas.core.bit_width import BitWidthConst
from brevitas.core.quant.int import TruncIntQuant

x = torch.tensor([127])
tq = TruncIntQuant(
    bit_width_impl=BitWidthConst(4),
    float_to_int_impl=RoundSte(),
)
y, scale, zp, bw = tq(x, scale=1, zero_point=0, input_bit_width=8)
print(f"({y}, {scale}, {zp}, {bw})")
```

Produced the following output:

```bash
(tensor([8.]), 1, 0, 4.0)
```

I.e., if two's-complement is used, the binary representation of the underlying data has changed from `0x01111111` to `0x1000`, but since the scale factor remains the same, the interpretation of these values changes from `127` -> `8`. Effectively, the operation divides the `value` of the input by `2**(input_bit_width - output_bitwidth)`, while also rounding the result. Also note, the most-significant bits of the input 8-bit type are always kept (possibly with rounding).

Finally, there is no guard against overflowing, if instead `x = torch.tensor([255])` in the above example, the output is:

```bash
(tensor([16.]), 1, 0, 4.0)
```

Which is an invalid `IntQuantTensor`, since the value `16` cannot be represented in 4-bits (if `scale=1`, `zero_point=0`).

### Current `TruncAvgPool` Implementation

This needs to split into 2 sections, since one part is the behaviour of `torch.nn.functional.AvgPool2d` with `IntQuantTensor` input, afterwards, the output of this functional call is modified and passed into a `TruncQuant` module.

#### `IntQuantTensor` and `torch.nn.functional`

When an `IntQuantTensor` is passed to `torch.nn.functional.AvgPool2d`, effectively the `scale`, `zero_point` parameters are ignored. The `bit_width` field is updated to match the total number of bits required to represent the *sum* which underlies the average operation. The `value` field of `IntQuantTensor` is simply passed to `torch.nn.functional.AvgPool2d` and passed onto the result.

For example, the following code:

```python
import torch

import torch.nn.functional as F
from brevitas.quant_tensor import IntQuantTensor
from brevitas.nn import QuantIdentity

torch.manual_seed(0)

k=3
x = torch.rand((1,1,k,k))
qi = QuantIdentity(return_quant_tensor=True, bit_width=8)

qi.train()
qi(x)
qx = qi(x) # Create valid `IntQuantTensor`

y = F.avg_pool2d(qx, kernel_size=k, stride=None, padding=0, ceil_mode=False, count_include_pad=True, divisor_override=None)

print(qx)
print(y)
print(y.value / y.scale) # Should be integers for a valid `IntQuantTensor`
```

produces the following output:

```bash
IntQuantTensor(value=tensor([[[[0.4972, 0.7704, 0.0910],
          [0.1331, 0.3082, 0.6373],
          [0.4902, 0.8894, 0.4552]]]], grad_fn=<MulBackward0>), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, signed_t=True, training_t=True)
IntQuantTensor(value=tensor([[[[0.4747]]]], grad_fn=<AvgPool2DBackward0>), scale=0.00700347451493144, zero_point=0.0, bit_width=12.0, signed_t=True, training_t=True)
tensor([[[[67.7778]]]], grad_fn=<DivBackward0>)
```

Note that the `value` field of the output is no longer an integer multiple of the scale, meaning that this operations does not produce a valid `IntQuantTensor`.

#### Scaling and Passing to `TruncQuant`

Before the output of the `torch.nn.functional.AvgPool2d` call is passed to `TruncQuant` the `value` field of the intermediate `IntQuantTensor` is multiplied by the value of the denominator in the average calculation - effectively turning the intermediate result from a `AvgPool` to a `SumPool`. This set corrects the issue between the `value` and `scale` from the intermediate output and converts this into a valid `IntQuantTensor` again.

This intermediate result is then passed to `TruncQuant` and the truncation / reinterpretation process described in the previous section is performed.

Overall, when the input and output bitwidths to a `TruncAvgPool` operation are the same, something somewhat sane occurs, for example:

```python
import torch

from brevitas.nn import TruncAvgPool2d, QuantIdentity

torch.manual_seed(0)

k=3
x = torch.rand((1,1,k,k))

qi = QuantIdentity(return_quant_tensor=True, bit_width=8)
qap = TruncAvgPool2d(kernel_size=(k,k), bit_width=8, float_to_int_impl_type='round', return_quant_tensor=True)

qi.train()
qi(x)
qi.eval()
qap.eval()
qx = qi(x)

y = qap(qx)

print(qx)
print(y)
print(y.value / y.scale) # Should be integers for a valid `IntQuantTensor`
```

Which produces the following output:

```bash
IntQuantTensor(value=tensor([[[[0.4972, 0.7704, 0.0910],
          [0.1331, 0.3082, 0.6373],
          [0.4902, 0.8894, 0.4552]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, signed_t=True, training_t=False)
IntQuantTensor(value=tensor([[[[0.2661]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, signed_t=True, training_t=False)
tensor([[[[38.]]]])
```

Note, that since the same random seed is used from the previous example, the output here `IntQuantTensor(value=tensor([[[[0.2661]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, ...` is directly comparable to the output in the previous example `IntQuantTensor(value=tensor([[[[0.4747]]]], ..., scale=0.00700347451493144, zero_point=0.0, bit_width=12.0, ...`.
Note, that the `bit_width` has reduced (12 to 8) as expected, while the scale factor remains the same in both. The dequantized value has scaled approximately proportionally to `k**2 / 16`, where `k**2` is the denominator of the average that occurred in the average pool, while `16` is `2**(input_bit_width - output_bitwidth)` as described in the `TruncQuant` section. An argument can be made that this output is a desired one, even though the dequantized value significantly differs from the expected (unquantized) one.

However, note that the input and output bitwidths to the `TruncAvgPool` are identical. If the bitwidth of `TruncAvgPool` is set to 12, the output becomes:

```bash
IntQuantTensor(value=tensor([[[[0.4972, 0.7704, 0.0910],
          [0.1331, 0.3082, 0.6373],
          [0.4902, 0.8894, 0.4552]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, signed_t=True, training_t=False)
IntQuantTensor(value=tensor([[[[4.2721]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=12.0, signed_t=True, training_t=False)
tensor([[[[610.]]]])
```

A side-effect of the above functionality has converted the average pool to a sum pool. Conversely, if the bitwidth of `TruncAvgPool` is set to 4, the output becomes:

```bash
IntQuantTensor(value=tensor([[[[0.4972, 0.7704, 0.0910],
          [0.1331, 0.3082, 0.6373],
          [0.4902, 0.8894, 0.4552]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=8.0, signed_t=True, training_t=False)
IntQuantTensor(value=tensor([[[[0.0140]]]]), scale=0.00700347451493144, zero_point=0.0, bit_width=4.0, signed_t=True, training_t=False)
tensor([[[[2.]]]])
```

Now, the dequantized output is approximately proportional to `k**2 / 256` to the expected (unquantized) value since the `TruncQuant` module is applying a much larger dividing factor.

### Takeaways

When `input_bitwidth==output_bitwidth` in a `TruncAvgPool` layer, we get somewhat sane functionality, but outside of this, I consider this behaviour at best "unintuitive" and at worst "buggy". Also, the sane behaviour for `input_bitwidth==output_bitwidth` seems to break several rules implicit with the Brevitas codebase, specifically:
 - The dequantized output of a quantizer should be close to the input value
 - If a operation produces a `*QuantTensor` that `QuantTensor` should have valid data
 - An average pool should produce something close to an average of its inputs, unless the `divisor_override` parameter is used
 
 ## Proposal
 
 In order to correct these issues, I propose the following:
  - [x] The scale should be adjusted after an `AvgPool` functional call so that a valid `IntQuantTensor` is produced
  - [x] The `TruncQuant` operator should not reinterpret it's output, instead its scale should be adjusted by the amount of truncation that has occurred
  - [x] Add a clamping operator to `TruncQuant` to avoid overflow / underflow
  - [x] The average pool shouldn't manipulate `IntQuantTensor`s - this should be handled by the functional call and `TruncQuant`
  - [x] The previous behaviour should be retained when `input_bitwidth==output_bitwidth` when the `divisor_override` parameter is used as `2**math.ceil(math.log2(k*k))`. Other scenarios require careful manipulation of `divisor_override` to be achieved.
    - Note, `divisor_override` parameter doesn't exist in ONNX (as of opset `onnx.ai=22`, so the exact previous behaviour is no longer attainable in QONNX, because the average pool can not be exported with this parameter set
 - [x] Update QONNX opset during export to be `qonnx.custom_op.general=2`

Furthermore, I question the usefulness in hardware for the current `TruncAvgPool` for the following reasons:
 - Forcing a power-of-two denominator for the average is only beneficial on hardware that only supports power-of-two scaling (otherwise the denominator can be absorbed and streamlined in FINN, for example)
 - Requiring that the most-significant bits after a summation are kept is likely not optimal unless the input values are all very high in magnitude

So finally, I recommend:
 - ~Introducing a `QuantAvgPool` layer, which can be instantiated with any sensible activation quantizer~ delay for future PR
 - [x] Creating a truncation quantizer which only allows power-of-2 shifts in scale W.R.T. the output of the `AvgPool` functional call
 - [x] Updating QONNX to opset to `qonnx.custom_op.general=2` so that the new behaviour can be differentiated from the old. Note, that after the inclusion of this PR, we have no plans to generate the old version of this Op.
 - [x] Update documentation of TruncQuant to reflect the new behaviour
 - [x] Add comments to JIT workarounds
 - [x] Add issue to track IntQuantTensor & JIT interoperation (#1266) 
 - [x] Add issue to track QONNX minimum version (#1267)


nickfraser:
CC @volcacius.

nickfraser:
CC @auphelia.

src/brevitas/core/scaling/int_scaling.py
```diff
@@ -34,3 +36,13 @@ def __init__(self, signed: bool):
     @brevitas.jit.script_method
     def forward(self, bit_width: Tensor) -> Tensor:
```
nickfraser, line 37: Add `signed` as an optional argument

src/brevitas/core/scaling/int_scaling.py
```diff
@@ -34,3 +36,13 @@ def __init__(self, signed: bool):
...
         return max_int(self.signed, False, bit_width) + 1
+
+
+class TruncPowerOfTwoIntScaling(brevitas.jit.ScriptModule):
```
nickfraser, line 41: Remove

tests/brevitas/core/test_trunc_int_quant.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def test_trunc_int_quant_defaults(self, bit_width_const):
+        trunc_int_quant = TruncIntQuant(
+            bit_width_impl=bit_width_const, float_to_int_impl=RoundSte())
+        assert isinstance(trunc_int_quant.tensor_clamp_impl, TensorClamp)
+        assert isinstance(trunc_int_quant.trunc_scaling_impl, TruncMsbScaling)
+        assert trunc_int_quant.narrow_range == False
+
+    # yapf: disable
+    @pytest_cases.fixture(
```
nickfraser, line 38: Switch to `pytest_cases.parametrize`

src/brevitas/quant/solver/trunc.py
```diff
@@ -26,9 +31,44 @@ def tensor_quant(quant_type):
...
+        else:
+            raise RuntimeError(f'trunc_scaling_impl_type={trunc_scaling_impl_type} not recognized.')
+
+
+class SolveTruncIntScalingImplFromEnum(ExtendedInjector):
+
+    @value
+    def trunc_int_scaling_impl(restrict_scaling_type):
+        if restrict_scaling_type == RestrictValueType.POWER_OF_TWO:
+            return TruncPowerOfTwoIntScaling
```
nickfraser, line 51: Switch to `PowerOfTwoIntScaling`

src/brevitas/quant/solver/trunc.py
```diff
@@ -26,9 +31,44 @@ def tensor_quant(quant_type):
...
+        else:
+            raise RuntimeError(f'restrict_scaling_type={restrict_scaling_type} not recognized.')
+
+
 class TruncQuantSolver(SolveBitWidthImplFromEnum,
                        SolveTensorQuantFloatToIntImplFromEnum,
-                       SolveTruncTensorQuantFromEnum):
+                       SolveActScalingImplFromEnum,
```
nickfraser, line 58: ZeroPoint solvers are missing.

src/brevitas/quant_tensor/int_torch_handler.py
```diff
@@ -115,6 +115,7 @@ def avg_pool2d_handler(
         avg_scaling = kernel_size * kernel_size
```
nickfraser, line 115: Should `kernel_size` or `divisor_override`

src/brevitas/quant/solver/trunc.py
nickfraser: ZeroPoint solvers are not required, because TruncQuant doesn't modify the ZeroPoint.

src/brevitas/core/scaling/int_scaling.py
```diff
@@ -21,7 +21,10 @@ def __init__(self, narrow_range: bool, signed: Optional[bool] = None):
     @brevitas.jit.script_method
     def forward(self, bit_width: Tensor, signed: Optional[Union[bool, Tensor]] = None) -> Tensor:
         is_signed = signed if signed is not None else self.signed
-        assert is_signed is not None, f"signed is not defined, signed={is_signed}"
+        if is_signed is None:
+            assert is_signed is not None, f"signed is not defined, signed={is_signed}"
```
Giuseppe5, line 25: This asserts seems unecessary?

src/brevitas/core/scaling/int_scaling.py
```diff
@@ -21,7 +21,10 @@ def __init__(self, narrow_range: bool, signed: Optional[bool] = None):
...
+            is_signed = True
```
Giuseppe5, line 26: When do we end up here?
Might be worth revisiting the quantizers to make sure that `is_signed` is always defined once we're here

nickfraser: This value gets passed down from user-facing APIs, so we can't ensure that always both values of signed are defined.

src/brevitas/function/ops.py
```diff
@@ -186,9 +188,11 @@ def min_int(
         >>> min_int(signed=False, narrow_range=False, bit_width=torch.tensor(8))
         tensor(0)
     """
-    if signed and narrow_range:
+    _signed = bool(signed.item()) if isinstance(signed, Tensor) else signed
```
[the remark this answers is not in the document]
nickfraser: > Was the previous signature a mistake?

No

> it's not clear to me where the Tensor might be coming from.

From the `IntQuantTensor` that is passed to the `TruncQuant`

src/brevitas/core/scaling/int_scaling.py
Giuseppe5: We can make signed a required argument for all quantizers?

nickfraser: > Might be worth revisiting the quantizers to make sure that is_signed is always defined once we're here

It's partially for backwards compatibility as well.

src/brevitas/core/scaling/int_scaling.py
nickfraser: Switch with `Raise`

src/brevitas/function/ops.py
```diff
@@ -149,22 +153,26 @@ def max_int(signed: bool, narrow_range: bool, bit_width: Tensor) -> Tensor:
         >>> max_int(signed=False, narrow_range=False, bit_width=torch.tensor(8))
         tensor(255)
     """
-    if not signed and not narrow_range:
+    _signed = bool(signed.item()) if isinstance(signed, Tensor) else signed
```
nickfraser, line 156: Comment that this is workaround for a JIT issue.

src/brevitas/core/scaling/int_scaling.py
```diff
@@ -11,26 +13,36 @@
...
         super(PowerOfTwoIntScaling, self).__init__()
         self.signed = signed
 
     @brevitas.jit.script_method
-    def forward(self, bit_width: Tensor) -> Tensor:
-        return max_int(self.signed, False, bit_width) + 1
+    def forward(self, bit_width: Tensor, signed: Optional[Union[bool, Tensor]] = None) -> Tensor:
+        is_signed = signed if signed is not None else self.signed
+        if is_signed is None:
+            assert is_signed is not None, f"signed is not defined, signed={is_signed}"
```
nickfraser, line 45: Replace with `raise`

src/brevitas/core/quant/int.py
```diff
@@ -199,30 +203,98 @@ def forward(self, x: Tensor) -> Tuple[Tensor, Tensor, Tensor, Tensor, Tensor, Te
...
+        >>> zero_point
+        tensor(0.)
+        >>> bit_width
+        tensor(4.)
+
+    Note:
+        Maps to quant_type == QuantType.INT == 'INT' == 'int' in higher-level APIs.
+
+    Note:
+        Set env variable BREVITAS_JIT=1 to enable TorchScript compilation of this module.
```
Giuseppe5, line 245: Not sure I was fully aware of this, looks nice

src/brevitas/quant/solver/trunc.py
```diff
@@ -26,9 +31,44 @@ def tensor_quant(quant_type):
...
+        if restrict_scaling_type == RestrictValueType.POWER_OF_TWO:
+            return PowerOfTwoIntScaling
+        else:
+            raise RuntimeError(f'restrict_scaling_type={restrict_scaling_type} not recognized.')
+
+
 class TruncQuantSolver(SolveBitWidthImplFromEnum,
                        SolveTensorQuantFloatToIntImplFromEnum,
-                       SolveTruncTensorQuantFromEnum):
+                       SolveActScalingImplFromEnum,
```
Giuseppe5, line 58: It has become a fully grown-up quantizer now

tests/brevitas_finn/brevitas/test_brevitas_avg_pool_export.py
```diff
@@ -63,6 +80,11 @@ def test_brevitas_avg_pool_export(
     odict = oxe.execute_onnx(model, idict, True)
     finn_output = odict[model.graph.output[0].name]
     # compare outputs
-    assert np.isclose(ref_output_array, finn_output).all()
+    #if restrict_scaling_type == "power_of_two" and kernel_size == 2:
```
Giuseppe5, line 83: Remove comments

tests/brevitas_finn/brevitas_examples/test_mobilenet_finn_export.py
```diff
@@ -42,6 +47,7 @@ def test_mobilenet_v1_4b(pretrained):
     # do forward pass in PyTorch/Brevitas
     expected = mobilenet(torch_tensor).detach().numpy()
     export_qonnx(mobilenet, input_shape=INPUT_SIZE, export_path=finn_onnx)
+    #output_scale = mobilenet.output.bias_quant.scale()  # Scale at the output
```
Giuseppe5, line 50: remove these?

src/brevitas/export/common/handler/qcdq.py
```diff
@@ -773,38 +773,44 @@ class QCDQCastTruncQuantProxyHandlerMixin(QuantAxisMixin,
...
 
     def symbolic_execution(
             self, x: Tensor, scale: Tensor, zero_point: Tensor, input_bit_width: Tensor,
             signed: Tensor):
         assert self.symbolic_kwargs is not None, 'Symbolic execution requires quant to be enabled'
         output_bit_width = self.symbolic_kwargs['output_bit_width']
+        narrow_range = self.symbolic_kwargs['narrow_range']
         dtype = self.int8_dtype() if signed else self.uint8_dtype()
-        trunc_scale = 2.0 ** (input_bit_width - output_bit_width)
+        scale = self.symbolic_kwargs['output_scale']  # Input scale is ignored now
```
Giuseppe5, line 795: Looking at the future, it might be weird to look at this code and trying to understand what this `now` means here.

Giuseppe5 (approved):
Small comments, otherwise LGTM.

tests/brevitas_finn/brevitas_examples/test_mobilenet_finn_export.py
nickfraser: I also prefer to keep

Issue #1036, opened 2024-09-30
LLM entrypoint with wikitext 2 enters an infinite loop


Giuseppe5:
As title says, using wikitext 2 with LLM entrypoint results in infinite loop.

nickfraser:
Solved in #1044.

PR #1044, opened 2024-10-08
Fix (examples/llm): Fix infinite loop in LLM entrypoint with WikiText2

pablomlago:
Fix for issue #1036. When fuse_sequences is False, and the seqlen is greater than 842 (maximum sequence length in WikiText2), the condition in line 131 of optimum/amd/brevitas/data_utils.py always evaluates to True, resulting in an infinite loop (see commit d0d06070225804dc6fecde00911ae38b53080123 in optimum-amd).

Consequently, we opted for enabling passing fuse_sequences as an argument and warning the user, when fuse_sequences=False, than an overly large seqlen can result in the infinite loop indicated in issue #1036.

### DIFF
--- a/src/brevitas_examples/llm/README.md
+++ b/src/brevitas_examples/llm/README.md
@@ -14,31 +14,19 @@
 Set the env variable BREVITAS_JIT=1 to speed up the quantization process. Currently unsupported whenever export is also toggled or with MSE based scales/zero-points.
 
 ```bash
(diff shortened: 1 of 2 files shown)


pablomlago:
The distribution of the sequence lengths when fuse_sequences is set to False in the following picture:
length_distribution_wikitext2_train
Note that, for the default value seqlen=2048, the infinite loop is entered, as the maximum sequence length is 842.

Giuseppe5:
Can you update the readme?
Generate the usage description with 
`python main.py -h` 
And copy and past it in the readme

PR #1070, opened 2024-10-23
Feat (export): qonnx minifloat export

Giuseppe5:
## Reason for this PR
Minifloat export to QONNX

## Changes Made in this PR

Extended QONNX export to support Weight/Activations export of generic minifloat

## Testing Summary

Simple test to generate the ONNX file. Since we don't have the corresponding mathematical implementation, we can't verify the math.

## Risk Highlight
 
We can only rely on visual inspection of the ONNX.

## Checklist

- [x] Initial CI/CD passing.

@maltanar

### DIFF
--- a/src/brevitas/export/onnx/qonnx/function.py
+++ b/src/brevitas/export/onnx/qonnx/function.py
@@ -71,6 +71,7 @@ class BrevitasFloatQuantFn(Function):
             exponent_bias,
             has_inf,
             has_nan,
+            saturating,
             has_subnormal,
             rounding_mode,
             max_val):
@@ -81,11 +82,12 @@ class BrevitasFloatQuantFn(Function):
             exponent_bit_width,
             mantissa_bit_width,
             exponent_bias,
+            max_val,
             has_inf_i=int(has_inf),
             has_nan_i=int(has_nan),
             has_subnormal_i=int(has_subnormal),
             rounding_mode_s=rounding_mode,
-            max_val_f=max_val)
+            saturation_i=saturating)
         ret.setType(x.type())
         return ret
 
@@ -99,6 +101,7 @@ class BrevitasFloatQuantFn(Function):
             exponent_bias,
             has_inf,
             has_nan,
+            saturating,
             has_subnormal,
             rounding_mode,
             max_val):
--- a/src/brevitas/export/onnx/qonnx/function.py
+++ b/src/brevitas/export/onnx/qonnx/function.py
@@ -59,6 +59,52 @@ class BrevitasQuantFn(Function):
         return y
 
 
+class BrevitasFloatQuantFn(Function):
+
+    @staticmethod
+    def symbolic(
+            g,
+            x,
+            scale,
+            exponent_bit_width,
+            mantissa_bit_width,
+            exponent_bias,
+            has_inf,
+            has_nan,
+            has_subnormal,
+            rounding_mode,
+            max_val):
+        ret = g.op(
+            f'{DOMAIN_STRING}::FloatQuant',
+            x,
+            scale,
+            exponent_bit_width,
+            mantissa_bit_width,
+            exponent_bias,
+            has_inf_i=int(has_inf),
+            has_nan_i=int(has_nan),
+            has_subnormal_i=int(has_subnormal),
+            rounding_mode_s=rounding_mode,
+            max_val_f=max_val)
+        ret.setType(x.type())
+        return ret
+
+    @staticmethod
+    def forward(
+            g,
+            x,
+            scale,
+            exponent_bit_width,
+            mantissa_bit_width,
+            exponent_bias,
+            has_inf,
+            has_nan,
+            has_subnormal,
+            rounding_mode,
+            max_val):
+        return x
+
+
 class BrevitasTruncFn(Function):
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


src/brevitas/export/onnx/qonnx/handler.py
```diff
@@ -14,14 +14,95 @@
...
+                'zero_point': torch.zeros_like(module.scale()),
+                'exponent_bit_width': module.exponent_bit_width(),
+                'mantissa_bit_width': module.mantissa_bit_width(),
+                'exponent_bias': module.exponent_bias(),
+                'saturating': module.is_saturating(),
+                'inf_values': module.inf_values(),
+                'nan_values': module.nan_values(),}
+
+    def symbolic_execution(self, x: Tensor):
+        xx = tuple(self.symbolic_kwargs.values())
```
maltanar, line 57: if the return values (xx) aren't used here and assuming the call to `.values()` has some useful side effect, suggest to amend this to `_ = tuple(self.symbolic_kwargs.values())`. otherwise this line could be removed completely

src/brevitas/export/onnx/qonnx/handler.py
```diff
@@ -14,14 +14,95 @@
...
+from .function import BrevitasFloatQuantFn
 from .function import BrevitasQuantFn
 from .function import BrevitasQuantLSTMCellFn
 from .function import BrevitasTruncFn
 
 
+class BrevitasFloatQuantProxyHandler(ONNXBaseHandler, ABC):
+
+    def validate(self, module):
+        assert not module.is_groupwise, "Export with Per Group quantization not supported"
```
maltanar, line 31: is this a fundamental limitation or just not implemented yet? the `FloatQuant` custom op does support broadcastable shapes for the scale, bitwidth.. inputs

src/brevitas/export/onnx/qonnx/handler.py
```diff
@@ -14,14 +14,95 @@
...
+                'mantissa_bit_width': module.mantissa_bit_width(),
+                'exponent_bias': module.exponent_bias(),
+                'saturating': module.is_saturating(),
+                'inf_values': module.inf_values(),
+                'nan_values': module.nan_values(),}
+
+    def symbolic_execution(self, x: Tensor):
+        xx = tuple(self.symbolic_kwargs.values())
+        scale = self.symbolic_kwargs['scale']
+        print(self.symbolic_kwargs.values())
```
maltanar, line 59: stray `print` for debugging?

src/brevitas/export/onnx/qonnx/handler.py
```diff
@@ -14,14 +14,95 @@
...
+
+class BrevitasWeightFloatQuantProxyHandler(BrevitasFloatQuantProxyHandler):
+    handled_layer = WeightFloatQuantProxyFromInjector
+
+    def __init__(self):
+        super().__init__()
+        self.quant_weights = None
+
+    def validate(self, zero_point):
+        assert zero_point == 0, "Zero-point not supported for binary quant."
```
maltanar, line 72: binary quant -> FloatQuant?

src/brevitas/export/onnx/qonnx/handler.py
```diff
@@ -14,14 +14,95 @@
...
+                'exponent_bit_width': module.exponent_bit_width(),
+                'mantissa_bit_width': module.mantissa_bit_width(),
+                'exponent_bias': module.exponent_bias(),
+                'saturating': module.is_saturating(),
+                'inf_values': module.inf_values(),
+                'nan_values': module.nan_values(),}
+
+    def symbolic_execution(self, x: Tensor):
+        xx = tuple(self.symbolic_kwargs.values())
+        scale = self.symbolic_kwargs['scale']
```
maltanar, line 58: same comment as above - if this is here for side effects assign to `_` otherwise could be removed

src/brevitas/export/onnx/qonnx/function.py
```diff
@@ -59,6 +59,52 @@ def forward(ctx, x, scale, zero_point, bit_width, narrow_range, signed, rounding
...
+            x,
+            scale,
+            exponent_bit_width,
+            mantissa_bit_width,
+            exponent_bias,
+            has_inf_i=int(has_inf),
+            has_nan_i=int(has_nan),
+            has_subnormal_i=int(has_subnormal),
+            rounding_mode_s=rounding_mode,
+            max_val_f=max_val)
```
maltanar, line 88: `FloatQuant`spec was updated 20241-10-30 to move `max_val` to be a compulsory, broadcastable input and no longer an attribute so this should be adjusted. ref:
https://github.com/fastmachinelearning/qonnx/blob/feature/float_quant/docs/qonnx-custom-ops/floatquant_op.md

maltanar (changes requested):
image

Generally looking good, requesting changes to comply with the updated spec plus suggested a few minor improvements

src/brevitas/export/onnx/qonnx/handler.py
Giuseppe5: debug leftover

src/brevitas/export/onnx/qonnx/handler.py
Giuseppe5: I think it's a residual of when groupwise didn't have its own proxy, and it requires some special care.

In principle there are limitations for groupwise export but not for the QONNX. However, we would need a separate class to handle groupwise, mostly because shaping of tensors/scales is special compared to per_tensor/per_channel case.

I would do it in a separate PR, but let me know if it's "urgent"

Giuseppe5:
Everything should be addressed, let me know if this matches QONNX specs

src/brevitas/export/onnx/qonnx/handler.py
maltanar: Got it, thanks! I don't think this is urgent yet.

maltanar (approved):
LGTM!

PR #1080, opened 2024-10-29
Fix (data): updating wikitext2 data utility

i-colbert:
## Reason for this PR

Update the wikitext2 data utility to work within the latest LLM quantization entry point. This version of the wikitext2 data loader uses the whole test dataset without random subsampling, which affords us more consistent benchmarking.

## Changes Made in this PR

Integrated the data pre-processing into the `get_wikitext2` function.

## Testing Summary

N/A 

## Risk Highlight

N/A

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/data.py
+++ b/src/brevitas_examples/llm/llm_quant/data.py
@@ -17,10 +17,11 @@ limitations under the License.
 """
 
 import random
+from typing import Any
 
 from datasets import load_dataset
 import torch
-from transformers import AutoTokenizer
+from tqdm import tqdm
 
 
 def get_c4(nsamples, seed, seqlen, tokenizer, split='train', nvalsamples=0):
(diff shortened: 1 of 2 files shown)


src/brevitas_examples/llm/llm_quant/data.py
```diff
@@ -17,10 +17,11 @@
 """
 
 import random
+from typing import Any
 
 from datasets import load_dataset
 import torch
-from transformers import AutoTokenizer
+from tqdm import tqdm
```
Giuseppe5, line 24: Not sure if we have tqdm listed as depency

Giuseppe5:
Is this version of wikitext2 used in some paper?
Is there anything we're missing from the old version of wikitext2?
Out of curiosity, if we were to compare with other popular implementation of this code (e.g., AutoGPTQ I guess), where do we land?

i-colbert:
> Is this version of wikitext2 used in some paper? Is there anything we're missing from the old version of wikitext2? Out of curiosity, if we were to compare with other popular implementation of this code (e.g., AutoGPTQ I guess), where do we land?

Yes, this version is modified from the [original GPTQ codebase](https://github.com/IST-DASLab/gptq/blob/main/datautils.py), as attributed in the file header, and is likely used by many works to collect results for their papers. The version in `optimum` uses random subsampling with replacement, which is useful for prototyping, but does not actually calculate the likelihood over the whole test dataset as sequences can be repeated or not even represented.

PR #1083, opened 2024-10-31
Feat (proxy): flag to enable/disable QT return

Giuseppe5:
## Reason for this PR

Considering we cache quant metadata during export, we do not need to propagate QuantTensors during export.
This will come in handy for future features, including lazy dequantization which assumes a different paradigm in inference vs export.

## Changes Made in this PR

Return a normal tensor from proxies during export.
Set all `return_quant_tensor` to False during tracing, and restore their states immediately after.

## Testing Summary

NA

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/nn/quant_avg_pool.py
+++ b/src/brevitas/nn/quant_avg_pool.py
@@ -15,6 +15,7 @@ from brevitas.function.ops_ste import ceil_ste
 from brevitas.inject.defaults import RoundTo8bit
 from brevitas.quant_tensor import _unpack_quant_tensor
 from brevitas.quant_tensor import QuantTensor
+from brevitas.utils.quant_utils import _CachedIO
 
 from .mixin.acc import AccQuantType
 from .mixin.acc import TruncMixin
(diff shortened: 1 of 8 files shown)


src/brevitas/nn/quant_avg_pool.py
```diff
@@ -38,6 +39,9 @@ def __init__(
         AvgPool2d.__init__(self, kernel_size=kernel_size, stride=stride)
         QuantLayerMixin.__init__(self, return_quant_tensor)
         TruncMixin.__init__(self, trunc_quant=trunc_quant, **kwargs)
+        self.cache_inference_quant_act = False
+        self.cache_quant_io_metadata_only = True
+        self.cache_class = None
```
nickfraser, line 44: Worth considering whether this should be its own class? Or should some of these be attributes of `_CachedIO`?

Any views on that?

nickfraser (approved):
See the comment. If you think this is the best approach then approved - please merge.

Issue #1073, opened 2024-10-25
Optimized rotation matrix


Giuseppe5:
**Is your feature request related to a problem? Please describe.**
Current rotation equalization method is based on orthogonal and hadamard matrices. We can extend this to learned rotational matrices

**Describe the solution you'd like**
We need to have an extra rotation class in `equalized_layers.py`, that holds the two learnable matrices.
After the rotation is applied and the layers are wrapped with this new equalized layer, we can apply the learning process, e.g., the one described in SpinQuant.

Post learning, the rotation matrices can be fused in their respective wrapped layers and remove the wrapped layers.

**Additional context**
The infrastructure to pick up the layers to rotate is already in place. The most effort is to put together the training process.

@i-colbert

Giuseppe5:
#1127

PR #1086, opened 2024-11-03
Add rotation matrix to equalized layers

vishwamartur:
Related to #1073

Add `RotationEqualizedLayer` class to `equalized_layer.py` to hold two learnable matrices for rotation.

* Implement `forward` method in `RotationEqualizedLayer` to apply rotation and call the wrapped layer.
* Add method to fuse rotation matrices into the wrapped layer.
* Add unit tests for `RotationEqualizedLayer` class in `test_equalized_layer.py`.
* Test forward pass with rotation matrices.
* Test fusing rotation matrices into the wrapped layer.

### DIFF
--- a/src/brevitas/nn/equalized_layer.py
+++ b/src/brevitas/nn/equalized_layer.py
@@ -41,3 +41,49 @@ class EqualizedModule(torch.nn.Module):
         # We convert everything to args so that hooks can work correctly
         out = self.layer(*kwargs.values())
         return out
+
+
+class RotationEqualizedLayer(torch.nn.Module):
+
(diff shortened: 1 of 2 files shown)


Giuseppe5:
Hi @vishwamartur,

I took a look at this PR and unfortunately I might have to close it, since it is very far away from the solution I had envisioned for this problem, and I am not even sure this solution would allow us to reach our goal, which is to replicate the setup used in some recent works like SpinQuant.

As mentioned in the other PR, it is always recommended to use the issues to discuss what solution you would like to propose.

In the next days I will update the relevant issue with a broad description of the solution I had planned for this problem.

Apologies for the miss-understanding and if you're still willing to work on this, stay tuned for the updates on the issue!

Issue #1076, opened 2024-10-25
New Transformers version has broken LLM tests


Giuseppe5:
**Describe the bug**
New transformers versions means we need to update our tests

**Reproducibility**
- [x] Can be reproduced consistently.
- [ ] Difficult to reproduce.
- [ ] Unable to reproduce.

**To Reproduce**

Run LLM tests

PR #1088, opened 2024-11-05
Fix LLM tests

pablomlago:
## Reason for this PR

See https://github.com/Xilinx/brevitas/issues/1076.

## Changes Made in this PR

Since transformers 4.46.0 (https://github.com/huggingface/transformers/releases/tag/v4.46.0), OPT attention mechanism relies, by default, on OPTSdpaAttention, thus resulting in the method replace_mha_with_quantizable_layers (https://github.com/pablomlago/brevitas/blob/master/src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py#L11) not injecting the Brevitas layers.

Additionally, the signature of the forward function in OPTAttention was updated in transformers 4.46.0, thus requiring to update the signature of QuantMultiHeadAttention analogously, even if this attribute is not actually used.

## Testing Summary

TODO

## Risk Highlight

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/tests/brevitas_examples/test_llm.py
+++ b/tests/brevitas_examples/test_llm.py
@@ -220,7 +220,7 @@ def test_small_models_toggle_run_args_pt_ge_2_4(
             "act_equalization": "layerwise",
             "gptq": True,
             "float_ppl": 31056.0 if transformers_version_ge('4.46.0') else 31274.05078125,
-            "quant_ppl": 33056.0 if transformers_version_ge('4.46.0') else 31278.166015625},])
+            "quant_ppl": 33056.0 if transformers_version_ge('4.46.0') else 33139.23046875},])
 def acc_args_and_acc(default_run_args, request):
     args = default_run_args
     run_dict = request.param


requirements/requirements-llm.txt
```diff
@@ -1,3 +1,3 @@
 # optimum-amd[brevitas] @ git+https://github.com/huggingface/optimum-amd.git@main
 tqdm
-transformers[sentencepiece]==4.45.2
+transformers[sentencepiece]>=4.46.0
```
nickfraser, line 3: Why have you bumped the version here? Does this no longer work with `transformers<4.46.0`?

tests/brevitas_examples/test_llm.py
```diff
@@ -200,18 +200,18 @@ def test_small_models_toggle_run_args_pt_ge_2_4(
...
-            "float_ppl": 33239.5,
-            "quant_ppl": 33283.75390625},])
+            "float_ppl": 33312.0,  # 33239.5,
+            "quant_ppl": 33056.0},  # 33283.75390625},
+        {
+            "model": "hf-internal-testing/tiny-random-MistralForCausalLM",
+            "act_equalization": "layerwise",
+            "gptq": True,
+            "float_ppl": 31056.0,  # 31274.05078125
+            "quant_ppl": 33056.0},])  # 33139.23046875},])
```
nickfraser, line 214: What do the numbers in the comments mean?

src/brevitas_examples/llm/llm_quant/prepare_for_quantize.py
```diff
@@ -2,11 +2,16 @@
...
 
 from brevitas.graph import ModuleToModuleByClass
 from brevitas_examples.llm.llm_quant.mha_layers import QuantizableOPTAttention
 
-QUANTIZABLE_MHA_MAP = {OPTAttention: (QuantizableOPTAttention, {'batch_first': True})}
+QUANTIZABLE_MHA_MAP = {
+    OPTAttention: (QuantizableOPTAttention, {
+        'batch_first': True}),
+    OPTSdpaAttention: (QuantizableOPTAttention, {
+        'batch_first': True}),}
```
nickfraser, line 14: An alternative is to add the argument `attn_implementation="eager"` to the `from_pretrained` call. @Giuseppe5, do you have a preference here? We would need to add a flag to main I guess...

src/brevitas/nn/quant_mha.py
```diff
@@ -602,7 +602,8 @@ def forward(
             key_padding_mask: Optional[Tensor] = None,
             need_weights: bool = True,
             attn_mask: Optional[Tensor] = None,
-            average_attn_weights: bool = True) -> Tuple[Tensor, Optional[Tensor]]:
+            average_attn_weights: bool = True,
+            position_ids: Optional[Tensor] = None) -> Tuple[Tensor, Optional[Tensor]]:
```
nickfraser, line 606: Please do not modify this forward function. `position_ids` is not in [PyTorch's MHA layer's forward function](https://pytorch.org/docs/stable/_modules/torch/nn/modules/activation.html#MultiheadAttention.forward). If this is argument is ignored, then `QuantizeOPTAttention` should handle that.

nickfraser (commented):
Open question about new pattern for layer replacement.

requirements/requirements-llm.txt
pablomlago: The test would be failing.

tests/brevitas_examples/test_llm.py
pablomlago: Those were the numbers for transformers < 4.46.0. They need to be removed, but I wanted to get confirmation on the preferred way to go regarding the transformers version.

nickfraser: Thanks - can you add that to the comment?

Giuseppe5: Maybe we can have a dict that pulls out different reference accuracy based on transformer version? It might get annoying in the long run to keep track of everything, but it could also help catch regression that we might have initially attributed to "numbers have changed"

nickfraser:
I think only @Giuseppe5's comment needs to be addressed, otherwise this can be merged IMO.

Giuseppe5:
There's a bunch of test failing on the extended test suite @pablomlago

PR #1097, opened 2024-11-18
Fix (scaling)!: clamp to avoid inf/nan in forward/backward

Giuseppe5:
## Reason for this PR

Restrict scaling with log op could cause inf/nan in the forward pass and/or in the gradients.

## Changes Made in this PR

Applying a restriction before (new behaviour) and after (existing behaviour) helps avoiding that.

## Testing Summary

Add tests for standalone and runtime scaling to check inf/nan in forward/backwards.

## Risk Highlight

This is a breaking change. Old checkpoints will work, new QAT/PTQ runs could produce different results from before.

- [x] This PR contains API-breaking changes.

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/core/scaling/standalone.py
+++ b/src/brevitas/core/scaling/standalone.py
@@ -251,9 +251,7 @@ class ParameterFromStatsFromParameterScaling(brevitas.jit.ScriptModule):
         if self.init_done:
             threshold = self.stats_scaling_impl.restrict_clamp_threshold(
                 self.restrict_threshold_pre(threshold))
-            # Clamping avoids eventual log(0) with restrict_val
-            value = self.clamp_scaling(self.value)
-            value = abs_binary_sign_grad(self.stats_scaling_impl.restrict_clamp_scaling(value))
+            value = abs_binary_sign_grad(self.stats_scaling_impl.restrict_clamp_scaling(self.value))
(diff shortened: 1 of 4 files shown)


src/brevitas/core/scaling/standalone.py
```diff
@@ -412,12 +417,14 @@ def forward(self, stats_input: Tensor, threshold: Optional[Tensor] = None) -> Te
         else:
             if self.counter <= self.collect_stats_steps:
                 out = self.buffer
+                # No clamping is necessary since statistics are already clamped in training_forward
                 out = self.restrict_scaling_pre(out)
             else:
                 out = self.value
             threshold = self.restrict_threshold(self.restrict_threshold_pre(threshold))
-            out = self.clamp_scaling(self.restrict_scaling(out))
```
Giuseppe5, line 419: I believe `self.clamp_scaling` is not necessary here because we're already in log domain, and `self.restrict_scaling` is responsible only for rounding and potential conversion from log to real domain (i.e., exponential).

Applying clamping before restrict_val means we can't get to negative values, thus scaling values between 0 and 1.
@nickfraser

nickfraser: Sounds right to me!

PR #1099, opened 2024-11-21
Fix (scaling/standalone): better switch from runtime stats to param

Giuseppe5:
## Reason for this PR

Currently, if we switch from training to eval before `stats_collection_steps` is done, we never update the `value` parameter to store the buffer value. This has a few side effects:
- When applying learned round, we might keep the model in eval model but still accumulate gradients. If the value parameter is not being used, no gradients are accumulated
- When exporting state_dict, value is not exported
- When doing PTQ calibration, current setup is such that the buffer is never converted to its corresponding parameter `value`, causing some of the issues mentioned above.

## Changes Made in this PR

~~At eval time, during the first iteration the buffer is always converted to param.
The side effect of this happens in the case the user would want to switch multiple times between training/evaluation mode very early on in the training process. Although it is common to switch between training/eval to check loss on the validation set, it is usually done after enough iteration that the buffer has already been converted to parameter anyway.
I'd admit that it could be marked as breaking change for this edge cases.~~

~~This has been removed in a more recent commit. I believe there are no more breaking changes at this point.~~

All fixed, no more breaking changes.
After calibration, we forcefully convert the buffer to parameters.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/core/scaling/standalone.py
+++ b/src/brevitas/core/scaling/standalone.py
@@ -374,6 +374,7 @@ class ParameterFromRuntimeStatsScaling(brevitas.jit.ScriptModule):
         self.restrict_inplace_preprocess = restrict_scaling_impl.restrict_init_inplace_module()
         self.restrict_scaling_pre = restrict_scaling_impl.restrict_init_module()
         self.restrict_threshold_pre = restrict_threshold_impl.restrict_init_module()
+        self.init_done: bool = brevitas.jit.Attribute(False, bool)
 
     @brevitas.jit.script_method
     def training_forward(self, stats_input: Tensor, threshold: Tensor) -> Tensor:
@@ -396,6 +397,7 @@ class ParameterFromRuntimeStatsScaling(brevitas.jit.ScriptModule):
             self.counter = new_counter
             return abs_binary_sign_grad(clamped_stats / threshold)
         elif self.counter == self.collect_stats_steps:
+            self.init_done = True
             self.restrict_inplace_preprocess(self.buffer)
             inplace_tensor_mul(self.value.detach(), self.buffer)
             threshold = self.restrict_threshold(self.restrict_threshold_pre(threshold))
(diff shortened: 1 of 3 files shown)


nickfraser (approved):
> The side effect of this happens in the case the user would want to switch multiple times between training/evaluation mode very early on in the training process.
Could the accuracy difference in the LLM tests be caused by this? I think the tests run at some very small `seqlen` (2?)

Otherwise, LGTM!

src/brevitas/core/scaling/standalone.py
```diff
@@ -374,6 +374,12 @@ def __init__(
         self.restrict_inplace_preprocess = restrict_scaling_impl.restrict_init_inplace_module()
         self.restrict_scaling_pre = restrict_scaling_impl.restrict_init_module()
         self.restrict_threshold_pre = restrict_threshold_impl.restrict_init_module()
+        self.init_done: bool = brevitas.jit.Attribute(False, bool)
+
+    def init_scale(self):
+        self.restrict_inplace_preprocess(self.buffer)
+        inplace_tensor_mul(self.value.detach(), self.buffer)
+        self.counter = self.counter + 1
```
nickfraser, line 382: Does this need to set `self.init_done = True`?

nickfraser (approved):
Check comment about `self.init_done`, otherwise LGTM!

src/brevitas/core/scaling/standalone.py
Giuseppe5: Init done is not needed. I added it in this PR but it's redundant with the last changes

src/brevitas/core/scaling/standalone.py
```diff
@@ -375,6 +375,12 @@ def __init__(
         self.restrict_scaling_pre = restrict_scaling_impl.restrict_init_module()
         self.restrict_threshold_pre = restrict_threshold_impl.restrict_init_module()
 
+    def init_scale(self):
+        if self.counter <= self.collect_stats_steps:
```
Giuseppe5, line 379: This avoids double re-init, since we're modifying in place the value of value.

PR #1116, opened 2024-12-05
Feat (examples/llm): Specify experiments via YAML files

nickfraser:
Add ability to specify the configuration of an experiment using the LLM example as a YAML file, which would easier to share with colleagues. Adding `--config <yaml_file>` will attempt to override the `main.py` defaults from the user provided YAML file. The user may still override these by adding furthering key/word arguments via the commandline.


Giuseppe5:
Would it be possible to use the more structured yaml to group together params that make sense to group together?
E.g., all rotation equalization stuff.

The list of args will only grow, but maybe this could help us keeping it somewhat organized.

nickfraser:
> Would it be possible to use the more structured yaml to group together params that make sense to group together?

I think this is a good idea, but my intuition is to save this for a future PR.

Giuseppe5 (approved):

PR #1134, opened 2024-12-18
Feat (mx): unpadding during dequantization

Giuseppe5:
## Reason for this PR

Groupwise quantization requires padding when the input channel shape is not divisible by groupsize.

Padding works well until it doesn't, and there are important edge cases that were not covered by the previous implementation.
 (e.g., weight only quantization where padding was required. Until now, we also had to force activation quantization because otherwise we had shape mismatch).

## Changes Made in this PR
With the current implementation, we un-pad when dequantizing, taking care of all the edge cases

Few todos:
- [x] Consolidate dequantization for groupwise tensors + inference export
- [x] Fix typing in runtime activations
- [x] Testing

## Risk Highlight

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/proxy/float_runtime_quant.py
+++ b/src/brevitas/proxy/float_runtime_quant.py
@@ -77,7 +77,7 @@ class ActFloatQuantProxyFromInjector(ActFloatQuantProxyFromInjectorBase):
             self,
             qt_args: Union[torch.Tensor, Tuple[Any]],
             x: Optional[FloatQuantTensor] = None) -> FloatQuantTensor:
-        if x is None:
+        if isinstance(qt_args, tuple):
             out = FloatQuantTensor(*qt_args, signed=self.is_signed, training=self.training)
         else:
             out = FloatQuantTensor(
--- a/src/brevitas/proxy/groupwise_float_parameter_quant.py
+++ b/src/brevitas/proxy/groupwise_float_parameter_quant.py
@@ -34,6 +34,7 @@ class GroupwiseWeightFloatQuantProxyFromInjector(WeightFloatQuantProxyFromInject
         return x.flatten(start_dim, start_dim + 1)
 
     def create_quant_tensor(self, qt_args: Tuple[Any]) -> GroupwiseFloatQuantTensor:
+        shape = self.tracked_parameter_list[0].shape
         out, scale, zero_point, exponent_bit_width, mantissa_bit_width, exponent_bias, saturating, inf_values, nan_values = qt_args
         return GroupwiseFloatQuantTensor(
             out,
@@ -48,4 +49,5 @@ class GroupwiseWeightFloatQuantProxyFromInjector(WeightFloatQuantProxyFromInject
             inf_values,
             nan_values,
             self.is_signed,
-            self.training)
+            self.training,
+            shape)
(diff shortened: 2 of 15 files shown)


src/brevitas/proxy/groupwise_float_parameter_quant.py
```diff
@@ -28,6 +28,7 @@ def apply_input_view(self, x):
         return x.flatten(start_dim, start_dim + 1)
 
     def create_quant_tensor(self, qt_args: Tuple[Any]) -> GroupwiseFloatQuantTensor:
+        shape = self.tracked_parameter_list[0].shape
```
Giuseppe5, line 31: We don't support weight quant sharing for groupwise anyway, so this is safe, but it is ugly.

src/brevitas/quant_tensor/groupwise_float_quant_tensor.py
```diff
@@ -89,19 +91,9 @@ def __torch_function__(self, func, types, args=(), kwargs=None):
...
-            new_scale = self.scale_.expand(curr_shape).flatten(start_dim, start_dim + 1)
-        else:
-            new_scale = self.scale_
-        if self.zero_point_.shape != ():
-            new_zp = self.zero_point_.expand(curr_shape).flatten(start_dim, start_dim + 1)
-        else:
-            new_zp = self.zero_point_
-
-        return new_value, new_scale, new_zp
+        from brevitas.utils.quant_utils import groupwise_dequant
```
Giuseppe5, line 94: Also ugly, maybe the function should live somewhere else?

src/brevitas/proxy/groupwise_float_runtime_quant.py
```diff
@@ -45,7 +45,8 @@ def create_quant_tensor(
                 inf_values,
                 nan_values,
                 self.is_signed,
-                self.training)
+                self.training,
+                dequant_shape=x.shape)
```
nickfraser, line 49: Do we need to add to the clause that `x is not None`? I.e., won't this fail if `x=None`?

src/brevitas/proxy/groupwise_float_runtime_quant.py
```diff
@@ -60,5 +61,6 @@ def create_quant_tensor(
                 x.inf_values,
                 x.nan_values,
                 x.signed,
-                self.training)
+                self.training,
+                x.shape)
```
nickfraser, line 65: Same here: won't this fail if `x=None` & `qt_args` is a `Tensor`?

src/brevitas/proxy/float_runtime_quant.py
```diff
@@ -77,7 +77,7 @@ def create_quant_tensor(
             self,
             qt_args: Union[torch.Tensor, Tuple[Any]],
             x: Optional[FloatQuantTensor] = None) -> FloatQuantTensor:
-        if x is None:
+        if isinstance(qt_args, tuple):
             out = FloatQuantTensor(*qt_args, signed=self.is_signed, training=self.training)
         else:
             out = FloatQuantTensor(
```
nickfraser, line 83: Same here: won't this fail if `x=None` & `qt_args` is a `Tensor`?

nickfraser (changes requested):
Please check guards for the optional argument `x`. I think this can crash under certain circumstances. If there are preconditions that means that this can never occur, maybe add a comment about this.

src/brevitas/proxy/groupwise_float_runtime_quant.py
Giuseppe5: X is never None anymore. File runtime_quant.py, line 195 of the new file

nickfraser: Can we update the function signature to `create_quant_tensor` then? `x: GroupwiseFloatQuantTensor)` Or will this cause some other issue that I'm not seeing?

nickfraser (approved):
Approve once the function signature of `create_quant_tensor` is updated.

PR #1172, opened 2025-02-02
Feat (gpfq): optimizing with lower diagonal matrix formulation

i-colbert:
## Reason for this PR

Currently, there are two GPFQ implementations: (1) `GPFQ`, which is faster for smaller models but runs into memory issues with larger models, and (2) `GPFQv2` which avoid memory issues with larger models. See links below for these two formulations.

There is another formulation of GPFQ that is both more memory efficient and compute efficient, which this PR implements.

**References for original GPFQ formulation:**
- [A Greedy Algorithm for Quantizing Neural Networks](https://www.jmlr.org/papers/v22/20-1233.html)
- [Post-training Quantization for Neural Networks with Provable Guarantees](https://epubs.siam.org/doi/abs/10.1137/22M1511709)

**Reference for memory-efficient GPFQ:**
- [Accumulator-Aware Post-Training Quantization](https://arxiv.org/abs/2409.17092)

## Changes Made in this PR

Improved GPFQ implementation and removal of previous GPFQ implementations.

## Testing Summary

Using the existing tests for GPxQ.

## Risk Highlight

- [x] This PR includes code from another work (please detail).

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/graph/gpfq.py
+++ b/src/brevitas/graph/gpfq.py
@@ -206,7 +206,7 @@ class GPFQ(GPxQ):
             Dg[group_index].copy_(self.G[group_index].diag())
             Dh[group_index].copy_(self.H[group_index].diag())
         # if either norms are 0, the weight is effectively pruned
-        Ds = torch.where(Dg * Dh != 0, Dg / Dh, 0.)  # \hat{D}_tt / D_tt
+        Ds = torch.where(Dg * Dh != 0, Dg / Dh, torch.zeros_like(Ds))  # \hat{D}_tt / D_tt
 
         Lg: Tensor = torch.zeros((self.groups, self.columns, self.columns),
                                  device=dev,
--- a/src/brevitas/graph/gpfq.py
+++ b/src/brevitas/graph/gpfq.py
@@ -206,7 +206,7 @@ class GPFQ(GPxQ):
             Dg[group_index].copy_(self.G[group_index].diag())
             Dh[group_index].copy_(self.H[group_index].diag())
         # if either norms are 0, the weight is effectively pruned
-        Ds = torch.where(Dg * Dh != 0, Dg / Dh, torch.zeros_like(Ds))  # \hat{D}_tt / D_tt
+        Ds = torch.where(Dg * Dh != 0, Dg / Dh, torch.zeros_like(Dg))  # \hat{D}_tt / D_tt
 
         Lg: Tensor = torch.zeros((self.groups, self.columns, self.columns),
                                  device=dev,
--- a/src/brevitas/graph/gpfq.py
+++ b/src/brevitas/graph/gpfq.py
@@ -125,26 +125,6 @@ class GPFQ(GPxQ):
             current_layer.forward_count = 0
             raise StopFwdException
 
-    def get_quant_weights(self, g, t, permutation_list) -> Tensor:
-        # TODO: we would like to implement slicing to minimize the amount
-        # of quantization, but right now slicing is done for a single channel
-        # and we would need to implement it for an arbitary slice of channels
-        if isinstance(self.layer, qnn.QuantLinear):
-            # No slicing, not optimized
-            q = self.layer.quant_weight(quant_input=self.quant_metadata)
-            q = q.value.unsqueeze(0)  # [1, OC, IC]
-            q = q[g, :, permutation_list[g][:t]]  # [groups, OC/groups, t]
-        elif isinstance(self.layer, SUPPORTED_CONV_OP):
-            # No slicing, not optimized
-            q = self.layer.quant_weight(quant_input=self.quant_metadata)
-            q = q.value
-            if isinstance(self.layer, (qnn.QuantConvTranspose1d, qnn.QuantConvTranspose2d)):
-                q = q.transpose(1, 0)  # This performs a view
-            q = q.flatten(1)
-            q = q.view(self.groups, -1, q.shape[-1])
-            q = q[g, :, permutation_list[g][:t]]  # [groups, OC/groups, t]
-        return q
-
     def single_layer_update(self):
         assert not self.layer.weight_quant.requires_quant_input, \
             "Error: GPFQ does not support weight quantizers that require metadata from input quantizers."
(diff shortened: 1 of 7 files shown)


src/brevitas/graph/gpfq.py
```diff
@@ -280,103 +125,124 @@ def update_batch(self, module, input, current_layer):
...
-                perm = perm.to(weight.device)
-                permutation_list.append(perm)
-        else:
-            for group_index in range(self.groups):
-                # No permutation, permutation tensor is a ordered index
-                perm = torch.tensor(range(weight.shape[-1]), device=weight.device)
-                permutation_list.append(perm)
-        return permutation_list
+    def get_quant_weights(self, g, t, permutation_list) -> Tensor:
+        # TODO: we would like to implement slicing to minimize the amount
```
Giuseppe5, line 129: I can't think of a reason why slicing can't be done for a slice rather than a single value. We might need to change the shared function signature a bit, but I'd rather do that than having a completely new function

src/brevitas/graph/gpfq.py
```diff
@@ -280,103 +125,124 @@ def update_batch(self, module, input, current_layer):
...
+            q = q[g, :, permutation_list[g][:t]]  # [groups, OC/groups, t]
+        elif isinstance(self.layer, SUPPORTED_CONV_OP):
+            # No slicing, not optimized
+            q = self.layer.quant_weight(quant_input=self.quant_metadata)
+            q = q.value
+            if isinstance(self.layer, (qnn.QuantConvTranspose1d, qnn.QuantConvTranspose2d)):
+                q = q.transpose(1, 0)  # This performs a view
+            q = q.flatten(1)
+            q = q.view(self.groups, -1, q.shape[-1])
+            q = q[g, :, permutation_list[g][:t]]  # [groups, OC/groups, t]
```
Giuseppe5, line 145: For example, this doesn't handle groupwise convolutions/transposed convolutions

src/brevitas/graph/gpfq.py
```diff
@@ -280,103 +125,124 @@ def update_batch(self, module, input, current_layer):
...
+            "Error: GPFQ does not support weight quantizers that require metadata from input quantizers."
+
+        if hasattr(self.layer, 'allocate_params'):
             self.layer.allocate_params(self.layer)
-        weight = self.layer.weight.data
+        del self.B  # free up memory by deleting the buffer
+
+        weight: Tensor = self.layer.weight.data
+        if self.create_weight_orig:
+            assert hasattr(self.layer, 'weight_orig'), \
```
Giuseppe5, line 158: I'm confused by this assert, isn't it redundant? Once the flag is set, then weight_orig is definitely present

src/brevitas/graph/gpfq.py
```diff
@@ -280,103 +125,124 @@ def update_batch(self, module, input, current_layer):
...
+                "Error: GPFQ requires the original weights to be stored, see `create_weight_orig`."
+            weight_orig: Tensor = self.layer.weight_orig.data
+        else:
+            warnings.warn("Warning: GPFQ will perform better with `create_weight_orig=True`.")
+            weight_orig: Tensor = weight.clone()
```
Giuseppe5, line 163: Why are we still creating weight_orig and cloning?

src/brevitas/graph/gpfq.py
```diff
@@ -280,103 +125,124 @@ def update_batch(self, module, input, current_layer):
...
+                # higher magnitude activations are quantized first
+                self.G[group_index] = self.G[group_index, perm, :][:, perm]
+                self.H[group_index] = self.H[group_index, perm, :][:, perm]
+            else:
+                # No permutation, permutation tensor is a ordered index
+                perm = torch.tensor(range(self.H.shape[-1]), device=dev)
+            perm = perm.to(weight.device)
+            permutation_list.append(perm)
+
+        Dg: Tensor = torch.zeros((self.groups, self.columns), dtype=torch.float32)
```
Giuseppe5, line 203: Not sure if I'm a fan of typing inline. We never do it and I'd like to avoid inconsistencies.

Giuseppe5: Also, it's very clear what's going on without the typing

Giuseppe5 (changes requested):

src/brevitas/graph/gpfq.py
i-colbert: Leaving the option for a user to disable `create_weight_orig` in case of memory issues. In such a scenario, we still need the track the original floating-point weights for the update rules, but the floating-point activations aren't the true floating-point activations. This should save memory still since the duplications won't be stored.

Here, `weights` are being altered over time while `weight_orig` is not. Note that the main difference from the intended use of GPFQ is how we calculate the floating-point activations.

src/brevitas/graph/gpfq.py
i-colbert: Leaving the option for a user to disable `create_weight_orig` in case of memory issues. In such a scenario, we still need the track the original floating-point weights for the update rules, but the floating-point activations aren't the true floating-point activations. This should save memory still since the duplications won't be stored.

Here, `weights` are being altered over time while `weight_orig` is not. Note that the main difference from the intended use of GPFQ is how we calculate the floating-point activations.

src/brevitas/graph/gpfq.py
i-colbert: Agreed, and resolved.

src/brevitas/graph/gpfq.py
```diff
@@ -4,199 +4,40 @@
 from copy import deepcopy
 import math
 from typing import List, Optional
+import warnings
```
Giuseppe5, line 7: This is ok, but from now on I want to stop using warning and move everything to logging, so we have a bit more control

src/brevitas_examples/common/axe.py
```diff
@@ -7,16 +7,20 @@
...
     LinAlgError = RuntimeError
 
-from brevitas.graph.gpfq import GPFQv2
+from brevitas.function.ops import max_int
+from brevitas.function.ops import min_int
+from brevitas.graph.gpfq import GPFQ
 from brevitas.graph.gptq import GPTQ
 from brevitas.graph.gpxq import SUPPORTED_CONV_OP
 from brevitas.graph.gpxq import SUPPORTED_TCONV_OP
+from brevitas.utils.quant_utils import _CachedIO
```
Giuseppe5, line 23: This is only for IntQuantTensor, but I guess it's fine

i-colbert: The AXE implementation is currently still experimental. Will work to extend support in future PR.

PR #1178, opened 2025-02-11
Feat (hadamard): support region expansion

Giuseppe5:
## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/config.py
+++ b/src/brevitas/config.py
@@ -20,6 +20,7 @@ IGNORE_MISSING_KEYS = env_to_bool('BREVITAS_IGNORE_MISSING_KEYS', False)
 JIT_ENABLED = env_to_bool('BREVITAS_JIT', False) and _enabled
 NATIVE_STE_BACKEND_ENABLED = env_to_bool('BREVITAS_NATIVE_STE_BACKEND', False)
 VERBOSE = env_to_bool('BREVITAS_VERBOSE', False)
+LOGGING_LEVEL = os.environ.get('BREVITAS_LOGGING', 'CRITICAL')
 
 # Internal global variables
 _FULL_STATE_DICT = False
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1695,7 +1695,7 @@ class GraphRotationEqualization(RotationEqualization):
             id_list = [id(r.name_to_module[sink_name]) for sink_name in r.sinks_names]
             eq_layers.update(id_list)
 
-        if self.orphan_sink:
+        if len(orphan_regions) > 0:
             for o_r in orphan_regions:
                 # Layerwise have only a single sink named 'sinks0'
                 id_sink = id(o_r.get_module_from_name('sinks0'))
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1685,8 +1685,9 @@ class GraphRotationEqualization(RotationEqualization):
         eq_layers = set()
         orphan_regions = []
 
-        blacklist_orphan_layers = self.blacklist_layers + self.layers_to_expand
-        self.find_module(graph_model, orphan_regions, blacklist_layers=blacklist_orphan_layers)
+        if self.orphan_sink:
+            blacklist_orphan_layers = self.blacklist_layers + self.layers_to_expand
+            self.find_module(graph_model, orphan_regions, blacklist_layers=blacklist_orphan_layers)
 
         if len(expanded_regions) > 0:
             parameter_number_pre = 0
@@ -1702,7 +1703,7 @@ class GraphRotationEqualization(RotationEqualization):
             id_list = [id(r.name_to_module[sink_name]) for sink_name in r.sinks_names]
             eq_layers.update(id_list)
 
-        if self.orphan_sink or len(expanded_regions) > 0:
+        if len(orphan_regions) > 0:
             for o_r in orphan_regions:
                 # Layerwise have only a single sink named 'sinks0'
                 id_sink = id(o_r.get_module_from_name('sinks0'))
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -28,6 +28,7 @@ from brevitas.graph.base import ModuleInstanceToModuleInstance
 from brevitas.graph.base import ModuleInstanceTransformTensor
 from brevitas.graph.base import ModuleInstanceWrapModule
 from brevitas.graph.base import Transform
+from brevitas.graph.hadamard import find_closest_hadamard_number
 from brevitas.graph.hadamard import get_hadK
 from brevitas.graph.hadamard import matmul_hadU
 from brevitas.graph.hadamard import matmul_hadU_cuda
@@ -155,6 +156,7 @@ class Region:
     sinks: Dict = field(default_factory=dict)
     acts: Tuple = field(default_factory=tuple)
     name_to_module: Dict = field(default_factory=dict)
+    expand_region: bool = False
 
     @property
     def srcs_names(self):
@@ -1342,7 +1344,10 @@ def _apply_rotate(
         else:
             try:
                 # Build hadamard rotation matrix
-                rot_mat, K = get_hadK(hidden_dim)
+                original_rot_mat, original_k = rot_mat, K = get_hadK(hidden_dim)
+
+                hidden_dim = find_closest_hadamard_number(hidden_dim, hidden_dim + 1).cpu().item()
+                rot_mat, K = get_hadK(int(hidden_dim))
                 rot_func = _apply_had_device
             except AssertionError as e:
                 print(f"Incomptible shapes {hidden_dim}")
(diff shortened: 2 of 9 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -1519,7 +1555,25 @@ def find_module(self, model, regions: List, prefix=''):
...
                 full_name = prefix + '.' + name if prefix != '' else name
-                self.find_module(module, regions, full_name)
+                self.find_module(module, regions, full_name, blacklist_layers)
+
+    def find_module_by_name(self, model: nn.Module, regions: List[Region], prefix: str = ''):
+        """
+        Iterate through the model looking at immediate children of every module to look for named modules.
+        This allows us to stop the search when we meet a top-level module that is supported.
+        """
+        if prefix in self.layers_to_expand:
```
Giuseppe5, line 1565: Implicit assumption is that we use this function to layers_to_expand.

Fair assumption as we use the other function (`find_module`) for the supported layerwise modules.
It can also become a new args for the function I guess, but not much value in that at the moment

src/brevitas/graph/equalize.py
```diff
@@ -155,6 +156,7 @@ class Region:
     sinks: Dict = field(default_factory=dict)
     acts: Tuple = field(default_factory=tuple)
     name_to_module: Dict = field(default_factory=dict)
+    expand_region: bool = False
```
Giuseppe5, line 159: Not sure if I like putting this here.
The class is used for other things and we use this flag in a very specific way.

I guess this class was also created to account for such issues, so why not? Maybe in the future this flag will be used more

src/brevitas/graph/equalize.py
```diff
@@ -1342,7 +1344,10 @@ def _apply_rotate(
         else:
             try:
                 # Build hadamard rotation matrix
-                rot_mat, K = get_hadK(hidden_dim)
+                original_rot_mat, original_k = rot_mat, K = get_hadK(hidden_dim)
```
Giuseppe5, line 1347: There could be a better way to do this. In general, we generate this "too early", so the idea is to generate both

src/brevitas/graph/equalize.py
```diff
@@ -1384,27 +1389,44 @@ def _apply_rotate(
 
         for name, indexes in region.sinks.items():
             module = region.get_module_from_name(name)
+            weight_axis = _get_input_axis(module)
+
             # Only "weight" is rotated
-            tensor_names_axis = [("weight", _get_input_axis(module))]
+            if region.expand_region:
+                assert isinstance(module, nn.Linear), "Currently only Linear layers support expanded hadamard"
```
Giuseppe5, line 1396: The code just below can support any layer type, the issue is more on the activation side since we need to define also the dimension along which the input needs padding. It is not complicated to do tbh, but not necessary atm

src/brevitas/graph/equalize.py
nickfraser: Could this also be used in the channel-splitting case?

src/brevitas/graph/equalize.py
```diff
@@ -1501,16 +1523,30 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
...
+            self.layers_to_expand = []
+        self.supported_sinks = ()
 
-    def find_module(self, model, regions: List, prefix=''):
+    def find_module(
+            self,
+            model: nn.Module,
+            regions: List[Region],
+            prefix: str = '',
+            blacklist_layers: Optional[List[str]] = None):
```
Giuseppe5, line 1543: The exclusion list needs to be passed since we want to exclude all the layers that might be expanded

src/brevitas/graph/hadamard.py
```diff
@@ -73,6 +73,27 @@ def get_hadK(n, transpose=False):
     return hadK, K
 
 
+def find_closest_hadamard_number(orig, n):
+    import math
+
+    values_to_check = [172, 156, 140, 108, 60, 52, 40, 36, 28, 20, 12]
```
Giuseppe5, line 79: Hardcoded values as the ones below. Not pretty, not ugly

src/brevitas/nn/equalized_layer.py
```diff
@@ -53,24 +54,40 @@ def forward(self, *args, **kwargs):
...
         super().__init__()
         if had_mat is not None:
             self.had_mat = torch.nn.Parameter(had_mat).cpu()
         else:
             self.had_mat = None
         self.layer = layer
         self.k = k
+        self.expand = expand
+
+    def input_pad(self, inp):
```
Giuseppe5, line 67: Not very general. Not pretty, not ugly

src/brevitas/graph/equalize.py
```diff
@@ -1342,7 +1344,10 @@ def _apply_rotate(
         else:
             try:
                 # Build hadamard rotation matrix
-                rot_mat, K = get_hadK(hidden_dim)
+                original_rot_mat, original_k = rot_mat, K = get_hadK(hidden_dim)
+
+                hidden_dim = find_closest_hadamard_number(hidden_dim, hidden_dim + 1).cpu().item()
```
nickfraser, line 1349: `+ 1` hard coded? Shouldn't this be a parameter? E.g., what if I want to expand by _more_ than the next highest Hadamard?

src/brevitas/graph/equalize.py
nickfraser, line 1396: Save convs for a new PR?

src/brevitas_examples/llm/main.py
```diff
@@ -994,6 +1002,14 @@ def parse_args(args, override_defaults={}):
...
             args += [f"--{key}", str(override_defaults[key])]
             del override_defaults[key]
+=======
+    parser.add_argument(
+        '--rotation-layers-to-expand',
+        type=str,
+        default=[],
+        nargs='*',
+        help='A list of module names to expand with hadamard rotation. Default: %(default)s')
+>>>>>>> Feat (hadamard): support region expansion
```
Giuseppe5, line 1012: Ops

src/brevitas/graph/equalize.py
```diff
@@ -1384,27 +1389,44 @@ def _apply_rotate(
...
         for name, indexes in region.sinks.items():
             module = region.get_module_from_name(name)
+            weight_axis = _get_input_axis(module)
+
             # Only "weight" is rotated
-            tensor_names_axis = [("weight", _get_input_axis(module))]
+            if region.expand_region:
+                assert isinstance(module, nn.Linear), "Currently only Linear layers support expanded hadamard"
+                hidden_dim = module.weight.shape[1]
+                new_hidden = find_closest_hadamard_number(hidden_dim, hidden_dim + 1)
```
nickfraser, line 1398: `+1` here? Maybe the previous comment should live here?

src/brevitas/nn/equalized_layer.py
```diff
@@ -53,24 +54,40 @@ def forward(self, *args, **kwargs):
...
+        # TODO: This only works for Linear layers. We have an assert in equalize.py to check for this
+        hidden_dim = inp.shape[-1]
+        new_hidden = find_closest_hadamard_number(hidden_dim, hidden_dim + 1)
```
nickfraser, line 70: Similar here - the `+1` hardcoding...

src/brevitas/nn/equalized_layer.py
nickfraser, line 67: Can this share logic with the block in the `Region` class?

src/brevitas_examples/llm/main.py
```diff
@@ -981,6 +988,7 @@ def parse_args(args, override_defaults={}):
         type=str,
         nargs='*',
         help='A list of tasks for zero_shot evaluation. Default: %(default)s')
+<<<<<<< HEAD
```
nickfraser: Merge conflict not resolved

src/brevitas_examples/llm/main.py
```diff
@@ -994,6 +1002,14 @@ def parse_args(args, override_defaults={}):
         for key in extra_args_keys:
             args += [f"--{key}", str(override_defaults[key])]
             del override_defaults[key]
+=======
```
nickfraser: Merge conflict not resolved

nickfraser (changes requested):

src/brevitas/graph/equalize.py
Giuseppe5: You would need to do a "+1" expansion twice. We can change the function to accept a "step" argument, to decide how many steps to move forward.
Even then, not sure I want to expose this new `step` argument right now

src/brevitas/graph/equalize.py
Giuseppe5: Yes and no. When doing channel splitting you basically assume that all the regions are going to be expanded anyway.

src/brevitas/nn/equalized_layer.py
Giuseppe5: We should, I'll look into that

src/brevitas/graph/equalize.py
Giuseppe5: I think so, yeah. I'll open an issue to keep track of that as good first issue

src/brevitas/utils/logging.py
```diff
@@ -12,6 +13,25 @@
 from brevitas.utils.quant_utils import *
 
 
+def setup_logger(name):
```
nickfraser, line 16: :O!

src/brevitas/graph/equalize.py
```diff
@@ -1629,17 +1680,29 @@ def apply(self,
...
-        self.find_module(graph_model, orphan_regions)
+
+        blacklist_orphan_layers = self.blacklist_layers + self.layers_to_expand
+        self.find_module(graph_model, orphan_regions, blacklist_layers=blacklist_orphan_layers)
+
+        if len(expanded_regions) > 0:
+            parameter_number_pre = 0
+            for m in graph_model.parameters():
+                parameter_number_pre += m.numel()
+            logging.info(f"{len(expanded_regions)} layers will be expanded during rotation")
```
nickfraser, line 1695: Personally I think all logging should be protected with a `verbose` flag. Sometimes we will want to turn these things off (for some specific experiments)

nickfraser (approved):
Only one note about verbosity flags. I guess this can also be achieved by setting the log level higher globally.

src/brevitas/utils/logging.py
```diff
@@ -12,6 +13,25 @@
...
+
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.DEBUG)
```
nickfraser, line 19: Maybe have the logging level as an optional parameter? You might want to turn off all 'INFO' messages at some point.

src/brevitas/graph/equalize.py
Giuseppe5: I will go the pytorch way and have an env variable to set the level.
Default will be "INFO" and user can select DEBUG with BREVITAS_LOGGING=DEBUG

src/brevitas/utils/logging.py
Giuseppe5: See the comment above. I think this will change a lot as we figure out what to do/how to do it, I want to have something reasonable to start and I believe your comment/my proposal above should accomplish just that

src/brevitas/config.py
```diff
@@ -20,6 +20,7 @@ def env_to_bool(name, default):
 JIT_ENABLED = env_to_bool('BREVITAS_JIT', False) and _enabled
 NATIVE_STE_BACKEND_ENABLED = env_to_bool('BREVITAS_NATIVE_STE_BACKEND', False)
 VERBOSE = env_to_bool('BREVITAS_VERBOSE', False)
+LOGGING_LEVEL = os.environ.get('BREVITAS_LOGGING', 'CRITICAL')
```
nickfraser, line 23: Love it! <3

nickfraser (approved):
LGTM!

PR #1189, opened 2025-02-14
fix (examples/benchmark): Fix when `run_results.yaml` does not exist.

nickfraser:
Occasionally, `quantize_llm` can fail in a way that also kills `run_args_bucket_process` (e.g., it goes OOM). If this is the case, `run_results.yaml` will never be created and the `"crashed"` status will never be set.

This PR introduces a workaround for this case.


pablomlago:
LGTM, I was wondering if, in a future PR, we would like to add a number of per-process retries. to handle errors that kill the process, e.g. having threads that call `.join` on the processes, and check the exitcode once they return. Therefore, in case of a failure, each process can be potentially restarted, but the offending configuration will not be run again.

nickfraser:
Good idea, but I don't think we need to prioritise this ATM.

PR #1201, opened 2025-02-21
Feat (ex/llm): Add possibility for minifloat quantized scales, with constant scaling

nickfraser:
Includes #1198.


nickfraser:
This is now superseeded by #1198.

PR #1206, opened 2025-02-28
Local compile support

Giuseppe5:
## Reason for this PR

Expands our support for torch.compile

## Changes Made in this PR

Starting with torch 2.0, it is now possible to `compile` your code to get on-the-fly speed-up compared
to eager execution. With the most recent versions of torch, this functionality is greatly improved
with support to more operations and patterns.

Currently, there are three main ways to leverage torch.compile with Brevitas, each with its own pros and cons.

The first two of these approaches rely on newly introduced quant_inference_mode.
This mode should be used once quantization is finished, and the idea is to trade away some flexibility, which is not
needed anymore at inference time, in exchange for slightly faster execution times.

### Full model compile + quant_inference_mode

Pros: Best speeds-up
Cons: High compile time, high chance of failure

``` python
model = MyQuantModel()
# Quantization code goes here
...
# Inference
with torch.no_grad(), with quant_inference_mode(model):
    # We need one forward pass to cache quantization-related hyper-params
    model(**example_input)
    model = torch.compile(model)
```

### Quantizers compile + quant_inference_mode

Pros: Fast compile time, good speed-up (float8), less chance of errors
Cons: Sometimes speed-up are minimal or absent

``` python
model = MyQuantModel()
# Quantization code goes here
...
# Inference
with torch.no_grad(), with quant_inference_mode(model, compile=True):
    # We need one forward pass to cache quantization-related hyper-params
    model(**example_input)
```

### Quantizers compile + PTQ

Pros: Good PTQ speeds-up in some cases, can be combined with inference_mode + eval
Cons: Possibilities of errors when combined with various PTQ/various models

``` python
model = MyQuantModel()
for m in model.modules():
    if hasattr(m, 'compile_quant'):
        m.compile_quant()

# Quantization code goes here
...
# Inference
with torch.no_grad(), with quant_inference_mode(model, compile=True):
    # We need one forward pass to cache quantization-related hyper-params
    model(**example_input)
```

## Some Results
These are some results showing inference speed-up with approach number 2 (quantizers compile + quant_inference_mode)

### Sana 1.6B, with per-group fp8 quantization

|                           | Eager Inference Time (500 samples) | Compile Inference Time (500 samples) |
|---------------------------|------------------------------------|--------------------------------------|
| Float                     | 25m                                | Not Measured                         |
| Weight-only Quantization  | 1h14m                              | 26m                                  |
| Act + Weight Quantization | 2h10m                              | 1h15m                                |

### Llama 3.2 1B, with per-group fp8 quantization

|                           | Eager Inference Time (WikiText2) | Compile Inference Time (WikiText2) |
|---------------------------|----------------------------------|------------------------------------|
| Float                     | 12s                              | Not Measured                       |
| Weight-only Quantization  | 40s                              | 18s                                |
| Act + Weight Quantization | 1m                               | 40s                                |

## Testing Summary

Added test to check numerical correctness between various modes

## Checklist

- [x] Initial CI/CD passing.


src/brevitas/proxy/parameter_quant.py
```diff
@@ -97,6 +97,10 @@ def __init__(self, quant_layer: nn.Module, quant_injector: Injector) -> None:
         self.quant_tensor_class = None  # To be redefined by each class
         self.skip_create_quant_tensor = False
 
+    def compile_quant(self):
+        if self.tensor_quant is not None:
+            self.tensor_quant = torch.compile(self.tensor_quant, dynamic=True, fullgraph=True)
```
pablomlago, line 102: Do we need dynamic here? I would assume that shapes are static.

Giuseppe5: Best case scenario, it does nothing, worst case scenario, it helps with recompilation.

pablomlago: Maybe if it's possible to set it to False, we should do that: " if you say torch.compile(dynamic=True), we will try to make everything as dynamic as possible. This is mostly useful for small operators; if you try it on a big model it will (1) probably crash PT2 and (2) run slow for no good reason."](https://pytorch.org/docs/stable/torch.compiler_dynamic_shapes.html)

Giuseppe5: I would consider tensor quantization as small operator. I will try to do some benchmarking with this True/False and see what happens

pablomlago: Yeah, the operator itself should be "small", but it's present in every layer, so a small overhead in every operator might be significant model-wise.

nickfraser (approved):
LGTM!

nickfraser (approved):
LGTM!

PR #1214, opened 2025-03-08
Feat (magr): initial implementation of MagR

i-colbert:
## Reason for this PR

Initial implementation of [MagR: Weight Magnitude Reduction for Enhancing Post-Training Quantization](https://arxiv.org/abs/2406.00800), adapted from https://github.com/AozhongZhang/MagR.

The [default config](https://github.com/Xilinx/brevitas/blob/0b763213d5c7d891bc5de17ef2bb6177e94630e5/src/brevitas_examples/llm/config/default_template.yml) gives the following INT4 weight-only quantization results

|  | OPT-125M | Llama3 1B |
|-------------|----------|-----------|
| Float16     | 23.77    | 8.77      |
| RTN | 27.78    | 11.05     |
| MagR        | 27.67    | 11.01     |
| GPTQ        | 26.72    | 10.84     |
| GPTQ+MagR   | 26.34    | 10.30     |

via
```shell
python main.py --config=config/default_template.yml --model=meta-llama/Llama-3.2-1B --weight-bit-width=4 --eval --gptq --magr
```
where RTN = round-to-nearest (i.e., no `--gptq` or `--magr`)

#### Notable differences

The paper claims that $\eta \alpha$ is used in Line 4 of Algorithm 1 to scale $\mathbf{V}^k$ before projection and $\alpha = 0.001$ for most experiments, but a constant $\alpha$ is used in the [official implementation](https://github.com/AozhongZhang/MagR/blob/ed40ef0edf58072b210258ab5b0d21adbb13b55e/MagR.py#L92). In our implementation, we scale the projection by $\alpha \cdot \Vert H \Vert_2 / \Vert H \Vert_1$, which effectively increases the $\ell_\infty$ penalty when $H$ is estimated to be lower rank. In this case, we may be able to hide more error in the null space when updating the weights. Experiments suggest this change is a net improvement.

## Changes Made in this PR

- Creating new `MagR` class with `magr_mode` using `GPxQ` and `gpxq_mode` backbones
- Adding new `apply_magr` function to Huggingface entry point in `brevitas_examples`
- Migrating transposed convolution checks to use same `is_conv_transpose` utility
- Extending `test_gpxq.py` to also test new APIs

## Testing Summary

Testing basic `apply_magr` function across all equalization fixtures.

## Risk Highlight

- [x] This PR includes code from another work (please detail).

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/graph/base.py
+++ b/src/brevitas/graph/base.py
@@ -193,7 +193,7 @@ class ModuleToModule(GraphTransform, ABC):
         if load_state_dict:
             # work-around since weight_orig is from a previous quantization algorithm
             if hasattr(old_module, 'weight_orig'):
-                new_module.register_buffer('weight_orig', old_module.weight_orig.data)
+                new_module.register_buffer('weight_orig', old_module.weight_orig.data.cpu())
             if not is_parametrized(old_module):
                 new_module.load_state_dict(old_module.state_dict())
             else:
--- a/src/brevitas/graph/base.py
+++ b/src/brevitas/graph/base.py
@@ -191,6 +191,9 @@ class ModuleToModule(GraphTransform, ABC):
     def _replace_old_module(self, model, old_module, new_module, load_state_dict=True):
         replace_module(model, old_module, new_module)
         if load_state_dict:
+            # work-around since weight_orig is from a previous quantization algorithm
+            if hasattr(old_module, 'weight_orig'):
+                new_module.register_buffer('weight_orig', old_module.weight_orig.data)
             if not is_parametrized(old_module):
                 new_module.load_state_dict(old_module.state_dict())
             else:
--- a/src/brevitas/graph/gpxq.py
+++ b/src/brevitas/graph/gpxq.py
@@ -20,21 +20,12 @@ from brevitas.graph.calibrate import disable_return_quant_tensor
 from brevitas.graph.calibrate import DisableEnableQuantization
 from brevitas.graph.calibrate import restore_return_quant_tensor
 from brevitas.graph.utils import is_conv_transposed
+from brevitas.graph.utils import is_quant_module
 import brevitas.nn as qnn
 from brevitas.quant_tensor import _unpack_quant_tensor
-from brevitas.nn.quant_layer import QuantWeightBiasInputOutputLayer as QuantWBIOL
 from brevitas.quant_tensor import QuantTensor
 
-# supported base torch modules
-SUPPORTED_TCONV_TORCH_MODULE = (nn.ConvTranspose1d, nn.ConvTranspose2d, nn.ConvTranspose3d)
-SUPPORTED_CONV_TORCH_MODULE = (nn.Conv1d, nn.Conv2d, nn.Conv3d, *SUPPORTED_TCONV_TORCH_MODULE)
-
-# supported Brevitas quant modules
-SUPPORTED_TCONV_QUANT_MODULE = (
-    qnn.QuantConvTranspose1d, qnn.QuantConvTranspose2d, qnn.QuantConvTranspose3d)
-SUPPORTED_CONV_QUANT_MODULE = (
-    qnn.QuantConv1d, qnn.QuantConv2d, qnn.QuantConv3d, *SUPPORTED_TCONV_QUANT_MODULE)
-
+SUPPORTED_CONV_OP = (nn.Conv1d, nn.Conv2d, nn.Conv3d, nn.ConvTranspose1d, nn.ConvTranspose2d, nn.ConvTranspose3d)
 
 @dataclass
 class LayerHandler:
@@ -107,10 +98,12 @@ class gpxq_mode(ABC):
             self.model.forward = self.catch_stopfwd
 
     def _is_module_supported(self, module):
-        if isinstance(module, SUPPORTED_CONV_QUANT_MODULE):
-            return True
-        elif isinstance(module, qnn.QuantLinear):
-            return True
+        if is_quant_module(module):
+            is_quant_enabled = module.weight_quant.is_quant_enabled
+        else:
+            is_quant_enabled = False
+        if isinstance(module, (nn.Linear, *SUPPORTED_CONV_OP)):
+            return is_quant_enabled
         else:
             return False
 
@@ -142,7 +135,7 @@ class gpxq_mode(ABC):
                         f'Behaviour might deviate from what expected.')
 
                 # Attach hooks for GPTQ
-                if self._is_module_supported(module) and module.weight_quant.is_quant_enabled:
+                if self._is_module_supported(module):
                     gpxq_module_optimizer = self.initialize_module_optimizer(
                         module,
                         name,
@@ -203,7 +196,7 @@ class GPxQ(ABC):
 
         # By default, use groups = 1
         self.groups = 1
-        if isinstance(self.layer, SUPPORTED_CONV_TORCH_MODULE):
+        if isinstance(self.layer, SUPPORTED_CONV_OP):
             if is_conv_transposed(self.layer):
                 weight_shape[1], weight_shape[0] = weight_shape[0], weight_shape[1]
             self.groups = self.layer.groups
@@ -221,7 +214,7 @@ class GPxQ(ABC):
     def process_input(self, inp):
         # Input is a tuple, so we take first element
         inp = inp[0]
-        if isinstance(self.layer, QuantWBIOL):
+        if is_quant_module(self.layer):
             inp = self.layer.input_quant(inp)
             is_quant_enabled = self.layer.weight_quant.is_quant_enabled
         else:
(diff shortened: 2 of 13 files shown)


src/brevitas/graph/gpxq.py
```diff
@@ -13,18 +13,26 @@
...
 from brevitas.quant_tensor import QuantTensor
 
-SUPPORTED_TCONV_OP = (qnn.QuantConvTranspose1d, qnn.QuantConvTranspose2d, qnn.QuantConvTranspose3d)
+# supported base torch modules
+SUPPORTED_TCONV_TORCH_MODULE = (nn.ConvTranspose1d, nn.ConvTranspose2d, nn.ConvTranspose3d)
+SUPPORTED_CONV_TORCH_MODULE = (nn.Conv1d, nn.Conv2d, nn.Conv3d, *SUPPORTED_TCONV_TORCH_MODULE)
 
-SUPPORTED_CONV_OP = (qnn.QuantConv1d, qnn.QuantConv2d, qnn.QuantConv3d, *SUPPORTED_TCONV_OP)
+# supported Brevitas quant modules
+SUPPORTED_TCONV_QUANT_MODULE = (
```
Giuseppe5, line 32: Why can't we use the base torch modules?

i-colbert: We do use the base torch modules for this implementation, but the current `SUPPORTED_CONV_OP` uses only the Brevitas quant modules, so I separate them here to "TORCH" and "QUANT" modules so I can use the base torch module while still maintaining the current functionality of using only the Brevitas quant modules.

Giuseppe5: Since all Brevitas models are also instances of the base torch ones, we could get rid of the Brevitas' ones. I guess it doesn't hurt and maybe it helps with clarity but I'm not convinced

src/brevitas/graph/gpxq.py
```diff
@@ -138,7 +145,6 @@ def __enter__(self):
                     gpxq_module_optimizer = self.initialize_module_optimizer(
                         module,
                         name,
-                        act_order=self.act_order,
```
Giuseppe5, line 141: Seems unnecessary to remove this but ok

src/brevitas/graph/magr.py
```diff
@@ -0,0 +1,272 @@
...
+                             device='cpu',
+                             dtype=torch.float32,
+                             pin_memory=torch.cuda.is_available())
+        self.B = torch.zeros((self.groups, self.columns, self.columns),
+                             device='cpu',
+                             dtype=torch.float32,
+                             pin_memory=torch.cuda.is_available())
+        self.nsamples = 0
+
+    def update_batch(self, module, input, current_layer):
```
Giuseppe5, line 101: Is this different from GPTQ?

src/brevitas/graph/gpxq.py
i-colbert: From what I can tell, we need to keep the Brevitas modules to maintain functionality for `__is_module_supported(...)`. If we leave some layers unquantized (e.g., last layers in a model), then it changes the functionality. I don't have data to support a change in functionality like this.

Giuseppe5:
Could you post some results with int3? I am curious if that can highlight a bit more the advantages of MagR, since int4 looks like we're almost reaching a plateau

src/brevitas/graph/gpxq.py
i-colbert: It is removed here since the MagR implementation has no concept of `act_order`. Maybe you can think of another way to implement this, but the alternative I can think of is to pass unused arguments to MagR. Which would you prefer?

src/brevitas/graph/magr.py
i-colbert: Only by checking the base torch modules. If you prefer that we switch to base torch modules, we can use the GPTQ function here instead of this duplication.

src/brevitas/graph/base.py
```diff
@@ -191,6 +191,9 @@ def _init_new_module(self, old_module: Module, name=None):
     def _replace_old_module(self, model, old_module, new_module, load_state_dict=True):
         replace_module(model, old_module, new_module)
         if load_state_dict:
+            # work-around since weight_orig is from a previous quantization algorithm
+            if hasattr(old_module, 'weight_orig'):
+                new_module.register_buffer('weight_orig', old_module.weight_orig.data)
```
Giuseppe5, line 196: I am playing with weight orig for other stuff and I believe it makes sense to move this to cpu and then move it to cuda on demand, otherwise we have total model duplication and it's not good.

src/brevitas/graph/gpxq.py
```diff
@@ -213,9 +220,11 @@ def __init__(
     def process_input(self, inp):
         # Input is a tuple, so we take first element
         inp = inp[0]
-        inp = self.layer.input_quant(inp)
-
-        is_quant_enabled = self.layer.weight_quant.is_quant_enabled
+        if isinstance(self.layer, QuantWBIOL):
+            inp = self.layer.input_quant(inp)
+            is_quant_enabled = self.layer.weight_quant.is_quant_enabled
```
Giuseppe5, line 225: Where are we using this?

src/brevitas/graph/magr.py
Giuseppe5: Doesn't this also need quantized modules?
If I have a conv that is not quantized, shall I also apply MagR?

src/brevitas_examples/llm/llm_args.py
```diff
@@ -197,6 +197,9 @@ def create_llm_args_parser():
         '--quantize-input-zero-point', action='store_true', help='Quantize input zero-point.')
     parser.add_argument(
         '--quantize-last-layer', action='store_true', help='Quantize last nn.Linear layer.')
+    parser.add_argument('--magr', action='store_true', help='Apply MagR.')
+    parser.add_argument('--magr-alpha', type=float, default=1e-3, help='Alpha for MagR.')
+    parser.add_argument('--magr-num-steps', type=int, default=200, help='Number of steps for MagR.')
```
Giuseppe5, line 202: If any of this can be hard-coded without too big of an impact, I'd be happier

src/brevitas/graph/gpxq.py
i-colbert: We use it in line 232 of the target branch (copy-pasted below) to cache quantized metadata. I only increase indentation here and protect it with an if statement since the two lines here require the layer to be quantized.

```python
        # If using quantized activations, inp could be QuantTensor. In
        # this case, we overwrite the metadata.
        if isinstance(inp, QuantTensor):
            if is_quant_enabled and self.quant_metadata is None:
                self.quant_metadata = self.layer.input_quant.cache_class(inp, metadata_only=True)
            inp = inp.value
```

src/brevitas_examples/llm/llm_args.py
i-colbert: Sure, I'll hard-code the number of steps. From my experiments, the value for $alpha$ is sensitive to the model and its rank/nullspace. The paper didn't mention this much, but they only tested Llama2.

src/brevitas/graph/magr.py
i-colbert: MagR is applied as a pre-processing step before quantization, so all modules need to be unquantized. I didn't implement a way to specifically exclude layers. Maybe that can be in a later PR?

i-colbert: Note that this function is different from the GPTQ implementation in that it doesn't raise an exception after executing the layer. We collect all covariances for MagR at the beginning, then update layer-by-layer. You can see this difference in the `apply_magr()` function. There is also the trivial difference that MagR checks for the base torch module (`nn.Linear`) instead of the quantized Brevitas module (`qnn.QuantLinear`).  If you would prefer that we reduce code duplication, I can separate out the core update function into an `update_batch_inner()` for then GPTQ class and then inherent `GPTQ` instead of `GPxQ` for `MagR`.

src/brevitas/graph/gpxq.py
i-colbert: Changed to use the base torch modules.

tests/brevitas/graph/test_gpxq.py
```diff
@@ -78,3 +79,46 @@ def test_toymodels(toy_quant_model, act_order, use_quant_activations, apply_gpxq
...
+    if 'mha' in test_id:
+        inp = torch.randn(32, *IN_SIZE_LINEAR[1:])
+    else:
+        inp = torch.randn(32, *IN_SIZE_CONV_SMALL[1:])
+    model.eval()
+    model(inp)  # test forward pass and collect scaling factors
+    dataset = TensorDataset(inp, inp)
+    dataloader = DataLoader(dataset, batch_size=16, num_workers=0, pin_memory=True, shuffle=True)
+
+    apply_magr(model, dataloader)
```
Giuseppe5, line 124: Isn't any properties that we should check except for running?

i-colbert: Are there any properties that you would want to have included? We can always check that the L-infinity norm is in fact decreasing, but I am not confident that there is a theoretical guarantee that it should always decrease, so a failure might not mean a code bug. For reference, we don't do any other property checks except for running for the other GPxQ tests, likely for similar reasons.

Issue #1229, opened 2025-03-19
Add BOS token to calibration dataset in LLMs


Giuseppe5:
**Is your feature request related to a problem? Please describe.**
Through a fortuitous debug, we realized that adding BOS token to calibration dataset could help, at least in terms of perplexity, in data-dependent PTQ algorithms.
Pending a verification on the effect on few-shot accuracy, we might decide to add this as pre-processing step.

This would break reproducibility of past results.

**Describe the solution you'd like**
Option 1:
We have a breaking change, assuming this is a unequivocal improvement

Option 2:
We have yet another flag, default set to preserve retro-compatibility 

Option 3:
Same as above, but the default corresponds to the new behaviour, and a specific flag is needed to reproduce past results

Personally, I could go for Option 2 (since we haven't had a release in a while, it could be worth having a snapshot of the current behaviour), maybe with a warning that in the next release we will move to option 3 (assuming the change is always an improvement).

nickfraser:
Note, PR #1230 plans to eventually address this issue.

nickfraser:
Suggestion / proposal:

Perhaps we should consider modifying this:

>```
>    batch = data['input_ids'][:, (i * seqlen):((i + 1) * seqlen)]
>```

to break sentences at specific tokens (i.e., EOS, '.', '\n'), this would make it so that the prepending of BOS tokens into subsequent strings make sense. Of course, we would still need to crop the subsequences to match the target `seqlen`. It would add some extra annoyance, but at least it might make the whole pipeline make more sense.

nickfraser:
> Consequently, when partioning according to the seqlen of the model, this token will only be present in the first position of the first batch (i.e. testloader[0]["input_ids][0, 0]). Note that a similar preprocessing is used in [SpinQuant](https://github.com/facebookresearch/SpinQuant/tree/main) and [GPTQ](https://github.com/IST-DASLab/gptq).

Interestingly, the cropping for calibration with these tools selects from [random substrings](https://github.com/IST-DASLab/gptq/blob/669e51587e93a1cc04015119e26b1b00e48e77c9/datautils.py#L20-L30) within each sequence.

pablomlago:
[...]

PR #1240, opened 2025-04-01
Feat (brevitas_examples/llm): BOS preprocessing for calibration data

Giuseppe5:
## Reason for this PR
Follow up to #1229 

## Changes Made in this PR

The final results of the discussion is to add the possibility to pre-prepocess the calibration datasets, adding BOS token whenever needed, before fusing and splitting the datasets into chunks of seqlength length.

We also preserve the possibility to not do this, and thus have a single BOS token at the beginning of the entire calibration set, which doesn't make a whole lot of sense logically but it's a widespread setup

## Testing Summary

Check issue above for results

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -45,7 +45,7 @@ def create_llm_args_parser():
     parser.add_argument(
         '--dataset',
         type=str,
-        choices=['wikitext2', 'c4', 'wikitext2_clm', 'c4_clm', 'pile_clm'],
+        choices=['wikitext2', 'c4', 'pile'],
         default='wikitext2',
         help='Dataset to use for quantization (default: %(default)s)')
     parser.add_argument(
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -45,7 +45,7 @@ def create_llm_args_parser():
     parser.add_argument(
         '--dataset',
         type=str,
-        choices=['wikitext2', 'c4'],
+        choices=['wikitext2', 'c4', 'wikitext2_clm', 'c4_clm', 'pile_clm'],
         default='wikitext2',
         help='Dataset to use for quantization (default: %(default)s)')
     parser.add_argument(
@@ -368,6 +368,8 @@ def create_llm_args_parser():
         '--few-shot-zeroshot',
         action="store_true",
         help='Whether to do zero or few shot eval. Default %(default)s)')
+    parser.add_argument(
+        '--add-bos-token', action="store_true", help='Add BOS token. Default %(default)s)')
     parser.add_argument(
         '--few-shot-limit', type=int, default=None, help='Few shot limit. Default %(default)s)')
     parser.add_argument(
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -369,7 +369,9 @@ def create_llm_args_parser():
         action="store_true",
         help='Whether to do zero or few shot eval. Default %(default)s)')
     parser.add_argument(
-        '--add-bos-token', action="store_true", help='Add BOS token. Default %(default)s)')
+        '--no-bos-preprocessing',
+        action="store_true",
+        help='Do not add BOS token during pre-processing. Default %(default)s)')
     parser.add_argument(
         '--few-shot-limit', type=int, default=None, help='Few shot limit. Default %(default)s)')
     parser.add_argument(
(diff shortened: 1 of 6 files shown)


Giuseppe5:
Re-posting here from @nickfraser 

> 
> Sanity-checking results with this change - I'm seeing for several Llama-3.x models @ FP4, that perplexity doesn't change much, but it's not always an improvement. See results below:
> 
>  model | config | float_ppl | quant_ppl(dev) | quant_ppl(PR) |
> | --- | --- | --- | --- | --- |
> | meta-llama/Llama-3.2-1B | 1 | 8.78 | 10.20 | 10.19 |
> | meta-llama/Llama-3.2-1B | 2 | 8.78 | 10.79 | 10.35 |
> | meta-llama/Llama-3.2-1B | 3 | 8.78 | 10.28 | 10.43 |
> | meta-llama/Llama-3.2-1B | 4 | 8.78 | 10.94 | 10.66 |
> | meta-llama/Llama-3.2-1B | 5 | 8.78 | 10.53 | 10.45 |
> | meta-llama/Llama-3.2-1B | 6 | 8.78 | 11.06 | 10.70 |
> | meta-llama/Llama-3.2-1B | 7 | 8.78 | 10.30 | 10.40 |
> | meta-llama/Llama-3.2-1B | 8 | 8.78 | 10.89 | 10.70 |
> | meta-llama/Llama-3.2-3B | 9 | 6.99 | 8.09 | 7.76 |
> | meta-llama/Llama-3.2-3B | 10 | 6.99 | 8.32 | 7.90 |
> | meta-llama/Llama-3.2-3B | 11 | 6.99 | 8.20 | 7.89 |
> | meta-llama/Llama-3.2-3B | 12 | 6.99 | 8.29 | 7.85 |
> 
> More results to come...

src/brevitas_examples/llm/llm_args.py
```diff
@@ -45,7 +45,7 @@ def create_llm_args_parser():
     parser.add_argument(
         '--dataset',
         type=str,
-        choices=['wikitext2', 'c4'],
+        choices=['wikitext2', 'c4', 'wikitext2_clm', 'c4_clm', 'pile_clm'],
```
nickfraser, line 48: How does the `_clm` differentiate from the other versions? Looks like the preprocessing is handled by the `--no-bos-preprocessing` argument.

Giuseppe5: This is a mistake, I'll remove this

nickfraser (approved):
LGTM! Does this produce different values in our accuracy tests? These might need to be updated too in the full test suite.

src/brevitas_examples/llm/llm_quant/data.py
```diff
@@ -63,54 +83,134 @@ def get_c4(
...
-                # Add BOS token.
-                if tokenizer.eos_token_id is not None:
-                    input_ids[:, 0] = tokenizer.eos_token_id
+    return dataset
 
-                dataset.append({"input_ids": input_ids, "attention_mask": attention_mask})
-                pbar.update(1)
 
-    return dataset
+def group_texts(examples: Dict[str, List[np.ndarray]],
```
pablomlago, line 90: Add comment indicating that the code is adapted from this [file](https://github.com/huggingface/nanotron/blob/main/src/nanotron/data/processing.py).

src/brevitas_examples/llm/llm_quant/data.py
```diff
@@ -63,54 +83,134 @@ def get_c4(
...
+            _tokenize_and_group_texts,
+            tokenizer=tokenizer,
+            sequence_length=seqlen,
+            filter_empty_sequences=filter_empty_sequences),
+        input_columns=text_column_name,
+        remove_columns=data.column_names,
+        features=Features({"input_ids": Sequence(feature=Value(dtype="int64"), length=seqlen)}),
+        batched=True,
+        num_proc=dataset_processing_num_proc_per_process,
+        load_from_cache_file=False,
```
pablomlago, line 174: `load_from_cache_file` was disabled while experimenting, but ideally it should be `True` for efficiency. As in nanotron,  I would a flag `dataset_overwrite_cache`  to `get_dataset_clm`, and then do `load_from_cache_file = not dataset_overwrite_cache`.

Giuseppe5: Not sure if I want to add another flag. I'm going to hard-code it to True after the tests pass. If someone wants to overwrite their dataset, they might have to delete them manually

src/brevitas_examples/llm/llm_args.py
```diff
@@ -45,7 +45,7 @@ def create_llm_args_parser():
     parser.add_argument(
```
pablomlago, line 45: Do you think it is worth having separate arguments for passing the calibration/validation dataset? Almost always, we want to evaluate on wikitext2 for comparison with other works, which forces to set `---dataset wikitext2`, even though any other calibration dataset could be used.

src/brevitas_examples/llm/main.py
```diff
@@ -543,7 +542,8 @@ def quantize_llm(args, extra_args=None):
             with torch.no_grad(), quant_inference_mode(model, compile=args.compile_eval):
                 model(**calibration_loader[0])
 
-                wrapped_model = HFLM(pretrained=model)  # need to wrap for LLM eval
+                wrapped_model = HFLM(
```
pablomlago, line 545: I think `add_bos_token` is not currently exposed in `llm_args.py`. Maybe we should leave it as it was before, since the increment in performance was not very noticeable.

Giuseppe5: It got misshandled during review, my bad.

Anyway, I don't have a strong opinion about whether to have a separate flag for this or just stick with the old behaviour. @nickfraser what do you think?

src/brevitas_examples/llm/llm_args.py
Giuseppe5: Let's keep that for another PR

PR #1253, opened 2025-04-11
test (ex/llm): Minor fixes to tests. Add rotation tests.

nickfraser:
Depends on #1252.

 - [x] Add rotation run tests
 - [x] Better handling of PT version guards
 - [x] Guard FX-based rotation tests to PT>2.4
 - [ ] ~Disable or enable BOS preprocessing (probably disable?)~ Addressed in #1272
 - [ ] ~Decrease test parallelism? If so, add issue to track~ no longer required (thanks to #1277)
 - [ ] ~Decrease validation set size during tests (disable `--eval` during "run" tests?)~ Addressed in #1277
   - [ ] ~Test that some prior run tests would have failed with this change~ Addressed in #1277
 - [x] Update PPL scores for accuracy tests


nickfraser:
#1277 contains a fix for the dataset issue - perhaps we should re-enabled parallelism.

tests/brevitas_examples/test_llm.py
```diff
@@ -86,7 +98,7 @@ def update(self, **kwargs):
 
 
 def requires_fx(args):
-    return args.act_equalization == "fx" or args.weight_equalization or args.ln_affine_merge
+    return fx_required(args)
```
Giuseppe5, line 101: Shouldn't we replace the call to `requires_fx` with one to `fx_required`?

tests/brevitas_examples/test_llm.py
```diff
@@ -130,14 +146,22 @@ def default_run_args(request):
     args.seqlen = 2
     args.model = "hf-internal-testing/tiny-random-MistralForCausalLM"
     args.dataset = "c4"
-    args.eval = True
+    args.eval = False
```
pablomlago, line 149: The bottleneck in these tests was the dataset loading/preprocessing, so I feel with the mocking of `load_raw_dataset`, we could leave this flag as it was, since the whole suite takes ~ 3 min.

nickfraser: Thanks - agreed, I prefer this!

tests/brevitas_examples/test_llm.py
```diff
@@ -155,12 +164,13 @@ def default_run_args(request):
     args.input_bit_width = 8
     args.act_calibration = True
     args.dtype = "float32"
+    args.no_bos_preprocessing = True
```
pablomlago, line 167: This flag was replaced by `bos_preprocessing`

tests/brevitas_examples/test_llm.py
```diff
@@ -274,52 +269,22 @@ def test_small_models_toggle_run_args_pt_ge_2_4(
...
+            "quant_ppl": 36910.191},
         {
             "model": "hf-internal-testing/tiny-random-OPTForCausalLM",  # Requires PT>=2.4 to run
             "weight_equalization": True,
             "ln_affine_merge": True,
             "replace_mha": True,
-            "float_ppl": 51649.797,
-            "quant_ppl": 51694.785},
+            "float_ppl": 52994.4296875,
+            "quant_ppl": 52123.94140625},
```
nickfraser: Revert these PPL scores.

tests/brevitas_examples/test_llm.py
```diff
@@ -274,52 +269,22 @@ def test_small_models_toggle_run_args_pt_ge_2_4(
...
         {
             "model": "hf-internal-testing/tiny-random-OPTForCausalLM",  # Requires PT>=2.4 to run
             "weight_equalization": True,
             "ln_affine_merge": True,
             "quant_sdpa": True,
-            "float_ppl": 51649.797,
-            "quant_ppl": 51688.922},])
-def acc_args_and_acc_pt_ge_2_4(default_run_args, request):
+            "float_ppl": 52994.4296875,
+            "quant_ppl": 52116.3359375},])
```
nickfraser: Revert these PPL scores.

nickfraser (commented):
Fix PPL scores

PR #1257, opened 2025-04-14
Feat (graph/calibrate): refactor DisableEnableQuantization

pablomlago:
## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.


src/brevitas/graph/calibrate.py
```diff
@@ -47,67 +50,36 @@
...
 
 
-def disable_return_quant_tensor(model):
-    previous_state = {}
-    for module in model.modules():
-        if hasattr(module, 'return_quant_tensor'):
-            previous_state[module] = module.return_quant_tensor
-            module.return_quant_tensor = False
-    return previous_state
+def unpack_input(inp: Union[tuple, QuantTensor]) -> torch.Tensor:
```
Giuseppe5, line 53: We already have this in brevitas.quant_tensor, it's called _unpack_quant_tensor

If that doesn't work with tuple, it might be worth extending it (with the assumption that a tuple is always a decayed QT)

src/brevitas/graph/calibrate.py
```diff
@@ -191,74 +225,467 @@ def disable_act_quant_hook(self, module, inp, output):
...
+            self.weight_quant_state = DisableEnableQuantization.disable_weight_quantization(
+                model=self.model,
+                is_training=self.is_training,
+            )
+        if self.disable_bias_quant:
+            self.bias_quant_state = DisableEnableQuantization.disable_bias_quantization(
+                model=self.model,
+                is_training=self.is_training,
+            )
+        if self.disable_return_quant_tensor:
```
Giuseppe5, line 614: This is not a self flag.
If either act quant or weight quant is disabled, then this is True otherwise you have unwanted errors, and it might be tricky for the user to understand why

src/brevitas/graph/calibrate.py
```diff
@@ -341,15 +768,16 @@ def forward_hook_wbiol(self, module, inp, output, name):
         We do not return the original quant output, but the float one, to avoid error accumulation
         """
         # Compute float reference
-        self.disable_act_quantization(module, is_training=False)
-        self.disable_param_quantization(module, is_training=False)
+        DisableEnableQuantization.disable_act_quantization(module, is_training=False)
```
Giuseppe5, line 771: Why can't this be a context manager case?

src/brevitas/graph/gpxq.py
```diff
@@ -35,7 +33,7 @@ class LayerHandler:
     forward_count: int = 0
 
 
-class gpxq_mode(ABC):
+class gpxq_mode(disable_enable_quantization):
```
Giuseppe5, line 36: Maybe I would go for composability rather than inheritance.
Functionally might be the same, but intuitively I don't see why GPxQ should inherit from that

src/brevitas/graph/gpxq.py
```diff
@@ -110,6 +113,8 @@ def _is_module_supported(self, module):
             return False
 
     def __enter__(self):
+        # Disable quantization selectively
+        super().__enter__()
```
Giuseppe5, line 117: Ah ignore the previous comments, I guess with composibility we couldn't do this right?

src/brevitas/graph/calibrate.py
```diff
@@ -164,101 +137,352 @@ def apply(self, model):
         return model
 
 
-class DisableEnableQuantization(Transform):
-
-    def __init__(self, call_act_quantizer_impl=False):
-        super(DisableEnableQuantization, self).__init__()
-        self.disable_act_quant_hooks = []
-        self.call_act_quantizer_impl = call_act_quantizer_impl
+class DisableEnableQuantization:
```
Giuseppe5, line 140: Let's rename to QuantizationStatusManager or something like that

This would be a breaking change.

src/brevitas/graph/calibrate.py
```diff
@@ -164,101 +137,352 @@ def apply(self, model):
...
+    def __init__(
+            self,
+            model: nn.Module,
+            is_training: Optional[bool] = None,
+            call_act_quantizer_impl: bool = False,
+            disable_act_quant: bool = True,
+            disable_weight_quant: bool = True,
+            disable_bias_quant: bool = True,
+            disable_return_quant_tensor: bool = True,
+            excluded_modules: Optional[List[nn.Module]] = None):
```
Giuseppe5, line 373: When do we need this?

src/brevitas/graph/calibrate.py
```diff
@@ -191,74 +232,240 @@ def disable_act_quant_hook(self, module, inp, output):
...
+            model: nn.Module, is_training: bool, previous_state: Dict[nn.Module, bool]) -> None:
+        QuantizationStatusManager._restore_quantization_state(
+            model=model,
+            is_training=is_training,
+            previous_state=previous_state,
+            quant_proxies=_BIAS_PROXIES,
+        )
+
+
+class disable_enable_quantization:
```
Giuseppe5, line 350: Also this should be renamed to something else, like quantization_status or something

src/brevitas/graph/calibrate.py
```diff
@@ -191,74 +232,240 @@ def disable_act_quant_hook(self, module, inp, output):
...
+            activation stats, for example during activation calibration in PTQ
+        disable_quant_act (bool): whether to disable activation quantization
+        disable_weight_quant (bool): whether to disable weight quantization
+        disable_bias_quant (bool): whether to disable bias quantization
+    """
+
+    def __init__(
+            self,
+            model: nn.Module,
+            is_training: Optional[bool] = None,
```
Giuseppe5, line 371: Do we need a separate arg for this?

src/brevitas/graph/calibrate.py
```diff
@@ -341,15 +548,15 @@ def forward_hook_wbiol(self, module, inp, output, name):
         We do not return the original quant output, but the float one, to avoid error accumulation
         """
         # Compute float reference
-        self.disable_act_quantization(module, is_training=False)
-        self.disable_param_quantization(module, is_training=False)
-
+        QuantizationStatusManager.disable_act_quantization(module, is_training=False)
```
Giuseppe5, line 551: Can't we use the context manager here?

src/brevitas/graph/calibrate.py
pablomlago: It was a fake commit to run the test pipeline and check if a test kept failing, I have already reverted that. change

src/brevitas/graph/calibrate.py
pablomlago: It was needed for AWQ, I have refactored the context manager to enable this functionality without relying on the extra logic from excluded_modules.

src/brevitas/graph/calibrate.py
```diff
@@ -164,101 +137,352 @@ def apply(self, model):
...
+
+    def __init__(
+            self,
+            model: nn.Module,
+            is_training: Optional[bool] = None,
+            call_act_quantizer_impl: bool = False,
+            disable_act_quant: bool = True,
+            disable_weight_quant: bool = True,
+            disable_bias_quant: bool = True,
+            disable_return_quant_tensor: bool = True,
```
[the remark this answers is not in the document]
pablomlago: Setting `disable_return_quant_tensor = disable_act_quant or disable_weight_quant` relieves the user from setting this flag manually, thus preventing "QuantLayer is not correctly configured" errors, when running the model's forward within the context manager. Note that there are situations in which `disable_return_quant_tensor` could be False, even if weight quantization is disabled (e.g. the module has output_quant enabled), but this level of granularity is not needed for the use-cases we currently have, so we opted for the coarse-grained solution that does not require knowledge of the specific quantizers the modules have.

src/brevitas/graph/calibrate.py
pablomlago: Would `quantization_status_manager` work?

Giuseppe5: Ok

src/brevitas/graph/calibrate.py
pablomlago: Yes, we need to enable the user to set a custom training state within the context manager. This is leveraged in calibration_mode, which sets the quantizers to train mode.

PR #1248, opened 2025-04-09
Feat (ex/benchmark): Add optional start/end indices

nickfraser:
Add an option to crop the number of experiments should be run from the "benchmark" infrastructure. Useful if one wants to break down a large grid search in multiple smaller experiments and distribute across multiple nodes, e.g., using slurm or similar.

A user can now do something like this:

Default (run all 4 experiments, i.e., retain previous behaviour):

```bash
$ python llm_benchmark.py --config benchmark_config.yaml --dry-run
Num. experiments: 4
Benchmark args.:
        config: benchmark_config.yaml
        results_folder: ./
        gpus: 0
        num_gpus_per_process: 1
        max_num_retries: 1
        dry_run: True
        start_index: 0
        end_index: -1
Non-default args.:
        --weight-bit-width: [4, 5, 6, 7]
```

Run experiments 0 and 1:

```bash
$ python llm_benchmark.py --config benchmark_config.yaml --dry-run --start-index 0 --end-index 2
Num. experiments: 2
Benchmark args.:
        config: benchmark_config.yaml
        results_folder: ./
        gpus: 0
        num_gpus_per_process: 1
        max_num_retries: 1
        dry_run: True
        start_index: 0
        end_index: 2
Non-default args.:
        --weight-bit-width: [4, 5]
```

Run experiments 2-3:

```bash
$ python llm_benchmark.py --config benchmark_config.yaml --dry-run --start-index 2 --end-index 4
Num. experiments: 2
Benchmark args.:
        config: benchmark_config.yaml
        results_folder: ./
        gpus: 0
        num_gpus_per_process: 1
        max_num_retries: 1
        dry_run: True
        start_index: 2
        end_index: 4
Non-default args.:
        --weight-bit-width: [6, 7]
```

PR #1268, opened 2025-04-16
feat (ex/benchmark): Added option to shuffle order of benchmark processes

nickfraser:
Update to #1248. Also allow the option to shuffle the processes before launching the queue.

### DIFF
--- a/src/brevitas_examples/common/benchmark/utils.py
+++ b/src/brevitas_examples/common/benchmark/utils.py
@@ -13,6 +13,7 @@ import itertools
 import multiprocessing
 from multiprocessing import Queue
 import os
+import random
 import sys
 import time
 import traceback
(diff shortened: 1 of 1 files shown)


src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -245,6 +246,13 @@ def parse_config_args(args: List[str]) -> Namespace:
         help=
         'Index from which to end current run. A negative value runs all jobs from `--start-index` (default: %(default)s).'
```
pablomlago, line 247: Maybe indicate that this index is exclusive, e.g. if start-index=1, end-index=3, the jobs that will be run are [1,2], not [1,2,3]. From the code it is clear, but maybe it is worth being explicit in the help.

PR #1254, opened 2025-04-14
Fix [ex/llm] Fix per-row quant_sdpa broadcastable shape

nickfraser:
## Reason for this PR

Fix issue when using `--quant-sdpa` and `per_row` quantization. Previously this combination caused the following issue:

```bash
  File "/scratch/users/nfraser/workspace/brevitas/src/brevitas_examples/common/generative/quant_blocks.py", line 51, in forward                                                                                                                
    x = self.dynamic_scaling_broadcastable_fn(x, shape)                                                                                                                                                                                        
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                        
  File "/scratch/users/nfraser/workspace/brevitas/src/brevitas_examples/common/generative/quantize.py", line 389, in <lambda>                                                                                                                  
    kv_broadcastable_shape_lambda = lambda x, shape: x.view(shape[0], 1, shape[-2], shape[-1])                                                                                                                                                 
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                 
RuntimeError: shape '[1, 1, 64, 2048]' is invalid for input of size 65536
```

PR #1270, opened 2025-04-17
Fix (graph/hadamard): `.view` can fail with functional QuantSDPA

nickfraser:
Depends on #1254. When running the LLM example with the following (minimal?) config it fails as follows:

config:

```
functional_sdpa_quant: true
input_bit_width: 4
input_quant_granularity: per_row
input_scale_type: dynamic
model: hf-internal-testing/tiny-random-LlamaForCausalLM
replace_rmsnorm: true
rotation: fused_no_fx
rotation_sdpa_regions: true
```

```
  File "/scratch/users/nfraser/workspace/brevitas/src/brevitas/nn/quant_sdpa.py", line 215, in forward
    query, key, value = self.pre_process_q(query), self.pre_process_k(key), self.pre_process_v(value)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/users/nfraser/workspace/brevitas/src/brevitas/nn/equalized_layer.py", line 99, in functional_rotate_input
    inp = matmul_hadU(inp)
          ^^^^^^^^^^^^^^^^
  File "/scratch/users/nfraser/workspace/brevitas/src/brevitas/graph/hadamard.py", line 104, in matmul_hadU
    input = X.clone().view(-1, n, 1)
            ^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

Changing view with `reshape` may introduce a memory leak - I have not investigated this yet.

### DIFF
--- a/tests/brevitas_examples/test_llm.py
+++ b/tests/brevitas_examples/test_llm.py
@@ -169,7 +169,7 @@ def run_test_models_run_args(args, model_with_ppl):
         "act_equalization=fx,gptq=True",
         "quant_sdpa_fx_per_row",
         "quant_sdpa_functional_per_row",
-        ],
+        "functional_sdpa_quant=True,rotation=fused_no_fx",],
     params=[
         {},
         {"weight_param_method": "hqo"},
@@ -182,7 +182,14 @@ def run_test_models_run_args(args, model_with_ppl):
         {"ln_affine_merge": True},
         {"act_equalization": "fx", "gptq": True},
         {"quant_sdpa": True, "input_scale_type": "dynamic", "input_quant_granularity": "per_row"},
-        {"functional_sdpa_quant": True, "input_scale_type": "dynamic", "input_quant_granularity": "per_row"},])
+        {"functional_sdpa_quant": True, "input_scale_type": "dynamic", "input_quant_granularity": "per_row"},
+        {
+            "functional_sdpa_quant": True,
+            "rotation": "fused_no_fx",
+            "rotation_sdpa_regions": True,
+            "input_scale_type": "dynamic",
+            "replace_rmsnorm": True
+        },])
 # yapf: enable
 def toggle_run_args(default_run_args, request):
     args = default_run_args


src/brevitas/graph/hadamard.py
```diff
@@ -101,7 +101,7 @@ def find_closest_hadamard_number(starting_dim, steps=1):
 def matmul_hadU(X, transpose=False):
     n = X.shape[-1]
     hadK, K = get_hadK(n, transpose)
-    input = X.clone().view(-1, n, 1)
+    input = X.clone().reshape(-1, n, 1)
```
Giuseppe5, line 104: Reshape is a bit costly.
Having said that, this is the function that gets called on the CI because no GPUs are available.

If you don't get any error when running this with a GPU-capable machine, I would merge this without giving too much thought.

nickfraser: > Reshape is a bit costly.

Only when it's not possible to do `.view`, otherwise it's just does `.view`. See [documentation](https://pytorch.org/docs/stable/generated/torch.reshape.html).

>> When possible, the returned tensor will be a view of input. Otherwise, it will be a copy.

Honestly, it might be best practice to start using `reshape` in favour of `view`.

> Having said that, this is the function that gets called on the CI because no GPUs are available.

If that is the case there is bigger problem. This crashes was caused when there are GPUs available. The `fast_hadamard` package wasn't install though, due to some dependency issue.

Giuseppe5: > Honestly, it might be best practice to start using reshape in favour of view.
I wasn't aware that it does view when possible, in that case, I fully agree and I'll make a mental note out of this.

> The fast_hadamard package
Sorry, I said when GPUs are available but I did actually mean when GPUs are available AND the fast_had package is installed. Otherwise it makes sense to fall back to that.
We would need to test with the fast_had package, let me know if I can help there.

nickfraser: > We would need to test with the fast_had package, let me know if I can help there.

Please do. I don't have an environment with `fast_hadamard` installed ATM.

Giuseppe5: Tested locally, it seems to work as intended

PR #1279, opened 2025-05-06
Fix (export/onnx): Improved ONNX export performance

nickfraser:
ONNX export has been slow since the introduction of `PatchFp8Ops` during export, to allow the export of FP8 operators in PyTorch versions without native PyTorch support. The current solution is simply a workaround. We should work out if we need this any more, since it is likely not needed in new PyTorch versions. Maybe we should throw an assertion error if FP8 export is not supported.


nickfraser (commented):
Approved pending passing tests.

PR #1308, opened 2025-06-04
Feat (core): Remove assumptions on positiveness of scales

pablomlago:
Scales are assumed to be strictly positive throughout our codebase, which hinders the representational capacity of our quantizers (the bit sign of the scales is wasted).

As an example, for integer symmetric quantization with AbsMax, it is verified that:

$$-2^{b-1} \leq \left\lfloor\frac{W}{\alpha}\right\rceil \leq 2^{b-1} \iff 0 \leq \left\lfloor\frac{W}{\alpha}\right\rceil  \leq 2\times 2^{b-1} = 2^{b}$$

Therefore, allowing negative scales ensures that the weight with the maximum absolute value can be represented exactly, irrespective of whether it is positive or negative, as for $`w_{j} = w_{\text{argmax}_{i}|w_{i}|}`$, it is verified that:

$$\left\lfloor \frac{w_{j}}{\alpha} \right\rceil = \left\lfloor \frac{w_{j}}{-\text{sign}(w_{j})|w_{j}|}2^{b-1} \right\rceil = \lfloor -2^{b-1} \rceil = -2^{b-1}$$

This is in contrast with $`\alpha = \text{max}|w_{i}|`$, since:

$$\left\lfloor \frac{w_{j}}{\alpha} \right\rceil = \left\lfloor \frac{w_{j}}{|w_{j}|}2^{b-1} \right\rceil = \lfloor \text{sign}(w_{j})2^{b-1} \rceil = \text{sign}(w_{j})2^{b-1}$$

For this choice, if the abs-max in a group corresponded to a positive weight, its value will not be represented exactly as it is clamped after adding the zero-point. This is due to the fact that, when `narrow_range = False`, values get clamped to be within $`[-2^{b-1},2^{b-1}-1]`$.

### DIFF
--- a/tests/brevitas/function/test_autograd_ste_ops.py
+++ b/tests/brevitas/function/test_autograd_ste_ops.py
@@ -30,7 +30,6 @@ from tests.brevitas.hyp_helper import scalar_float_nz_tensor_st
 from tests.brevitas.hyp_helper import scalar_float_tensor_st
 from tests.brevitas.hyp_helper import two_float_tensor_random_shape_st
 from tests.marker import jit_disabled_for_export
-from tests.marker import requires_pt_ge
 
 # brevitas.ops.autograd_ste_ops. and not brevitas.function.ops.
 # in order to mock where it's used, not where it's defined
@@ -177,7 +176,6 @@ class TestScalarClampMinSte:
         assert_allclose(output, expected_output)
 
     @jit_disabled_for_export()
-    @requires_pt_ge('2.0.1')
     @given(x=scalar_clamp_min_ste_test_st())
     def test_symbolic(self, x):
         """


src/brevitas/core/scaling/runtime.py
```diff
@@ -182,7 +181,6 @@ def __init__(
     @brevitas.jit.script_method
     def forward(self, x):
         out = x * self.affine_weight + self.affine_bias()
-        out = abs_binary_sign_grad(out)
```
Giuseppe5, line 185: This operation happens after we have already clamped.

There are two issues now:
- If we _do_ want to constrain the scale to be positive, then we cannot do it anymore, since the sign can be flipped here
- Not strictly related to this change, but if the weight is zero, and we have already clamped, then we lose that clamping factor.

Problem number 2 is fairly easy to solve, but I'm not sure what to do for problem number one.

Giuseppe5: Actually the solution is simple.
We apply abs_binary_sign_grad to self.affine_weight and that's it

src/brevitas_examples/imagenet_classification/a2q/utils.py
```diff
@@ -290,7 +290,7 @@ def acc_reg_penalty(module: AccumulatorAwareParameterPreScaling, inp, output):
         nonlocal reg_penalty
         (weights, input_bit_width, input_is_signed) = inp
         s = module.scaling_impl(weights)  # s
-        g = abs_binary_sign_grad(module.restrict_clamp_scaling(module.value))  # g
+        g = module.restrict_clamp_scaling(module.value)  # g
```
Giuseppe5, line 293: Not sure I would change this, since it is an example

requirements/requirements-export.txt
```diff
@@ -1,2 +1,3 @@
 onnx==1.17.0
 onnxoptimizer
+qonnx
```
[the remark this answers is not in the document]
pablomlago: `requirements-export.txt` is installed in the environment for testing, and I needed `qonnx` for the test on the symbolic trace of `scalar_clamp_min_ste`.

tests/brevitas/function/test_ops_ste.py
```diff
@@ -182,3 +186,37 @@ def test_scalar_clamp_min_ste_backend(prefix_and_status: Tuple[str, bool], x):
...
+        # min_val must be set to a value greater than zero
+        assume(abs(x[0]) > 0.0)
+        min_val, val, val_grad = x
+        module = ScalarSignedClampMinSte(min_val)
+        val.requires_grad_(True)
+        output = module(val)
+        output.backward(val_grad, retain_graph=True)
+        assert_allclose(val_grad, val.grad)
+
+    @requires_pt_ge('2.0.1')
```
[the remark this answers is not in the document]
pablomlago: I recall that I was seeing failing tests in older pt versions, but I will double-check.

PR #1311, opened 2025-06-06
Feat (qronos): adding initial implementation of Qronos algorithm

i-colbert:
## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/graph/qronos.py
+++ b/src/brevitas/graph/qronos.py
@@ -1,0 +1,289 @@
+# Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import math
+
+import torch
+from torch import Tensor
+
+try:
+    from torch.linalg import LinAlgError
+except:
+    LinAlgError = RuntimeError
+
+import warnings
+
+import unfoldNd
+
+from brevitas.graph.gpfq import GPFQ
+from brevitas.graph.gpxq import SUPPORTED_CONV_OP
+from brevitas.graph.utils import is_conv_transposed
+import brevitas.nn as qnn
+from brevitas.utils.torch_utils import StopFwdException
+
+
+class Qronos(GPFQ):
+    """
+    Implementation of Qronos as proposed in: https://arxiv.org/pdf/2505.11695
+    """
+
+    def __init__(
+            self,
+            layer,
+            name,
+            act_order,
+            len_parallel_layers,
+            create_weight_orig,
+            num_blocks: int = 100) -> None:
+        super().__init__(layer, name, act_order, len_parallel_layers, create_weight_orig)
+        self.blocksize = math.ceil(self.columns / num_blocks)
+
+    def update_batch(self, module, input, current_layer):
+        if self.disable_pre_forward_hook:
+            return input
+
+        # Update reference to current layer
+        current_layer.layer_names.add(self.name)
+        inp = self.process_input(input)
+        batch_size = inp.shape[0]
+
+        is_quant_enabled = module.weight_quant.is_quant_enabled
+
+        # Preprocess the input to compute the covariance
+        if isinstance(self.layer, qnn.QuantLinear):
+            if len(inp.shape) > 2:
+                inp = inp.reshape((-1, sum(inp.shape[2:])))
+            inp = inp.t()
+            # For QuantLinear layer, groups will be 1
+            inp_processed = inp.unsqueeze(0)
+
+        if isinstance(self.layer, SUPPORTED_CONV_OP):
+            # Pick the correct unfoldNd class
+            if is_conv_transposed(self.layer):
+                unfold_impl = unfoldNd.UnfoldTransposeNd
+            else:
+                unfold_impl = unfoldNd.UnfoldNd
+
+            unfold = unfold_impl(
+                self.layer.kernel_size,
+                dilation=self.layer.dilation,
+                padding=self.layer.padding,
+                stride=self.layer.stride)
+
+            # Split input based on how many groups in convolution
+            inp_by_group = torch.chunk(inp, self.groups, 1)
+            inp_processed = []
+            # Preprocess input by group
+            for inp in inp_by_group:
+                inp = unfold(inp)
+                inp = inp.transpose(1, 0)
+                inp = inp.flatten(1)
+                inp_processed.append(inp)
+            inp_processed = torch.stack(inp_processed)
+
+        # Normalizing by the sequence length for numerical stability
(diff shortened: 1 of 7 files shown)


src/brevitas_examples/llm/main.py
```diff
@@ -311,11 +312,14 @@ def quantize_llm(args, extra_args=None):
     if args.magr and not args.load_checkpoint:
         print("Applying MagR...")
         model = offload_model(model)
+        # save original weights for GPFQ or Qronos since they both require optimizing
+        # a mismatched objective function (see https://arxiv.org/abs/2505.11695), or
+        # if explicitly specified in the CLI
+        _create_weight_orig = args.gpfq or args.qronos or args.gpxq_create_weight_orig
```
Giuseppe5, line 318: I don't like this.
The user is fully responsible to set the flag if they are using GPFQ or Qronos.

We should just document somewhere that they are needed (e.g., in the yamls to reproduce the results)

src/brevitas_examples/llm/llm_quant/gpxq.py
```diff
@@ -149,24 +150,22 @@ def apply_gptq(
...
+def _gpfq_callback(
         model,
         dataloader,
         act_order=True,
         group_of_parallel_layers=None,
         block_name=None,
         max_accumulator_bit_width=None,
-        max_accumulator_tile_size=None):
+        max_accumulator_tile_size=None,
+        gpfq_class=GPFQ):
```
Giuseppe5, line 161: Not a massive fan of this, since Qronos and GPFQ are related but not the same thing.
The idea is correct, maybe it will require a change of names. Something like `single_forward_optimization` vs `double_forward_optimization`.

If you don't want to change this now, I guess it can wait.

src/brevitas_examples/imagenet_classification/ptq/ptq_common.py
```diff
@@ -608,38 +609,67 @@ def apply_gptq(
...
-                       gpfq_class=gpfq_class) as gpfq:
-            gpfq_model = gpfq.model
-            for i in tqdm(range(gpfq.num_layers)):
-                for i, (images, target) in enumerate(calib_loader):
-                    images = images.to(device)
-                    images = images.to(dtype)
-                    gpfq_model(images)
-                gpfq.update()
+    # We use the mismatched optimization callback, which uses two forward passes
+    _mismatched_optimization_callback(
```
Giuseppe5, line 651: `mismatched` has a negative connotation in my mind, but I could be totally wrong on this

i-colbert: Updated to `_dual_optimization_callback`, which may be a bit more descriptive? The code comment should help provide intuition though. This is for when we optimize what we call the "mismatched objective function" $$\Vert XW - \tilde{X}W \Vert $$ instead of $$ \Vert XW - XQ \Vert $$, which allows us to not only correct for weight quantization error, but also errors from previously quantized layers. GPFQ has historically done this, but now Qronos does as well.

src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
```diff
@@ -302,7 +303,7 @@ def validate_args(args):
     help='Use quant activations for GPTQ (default: disabled)')
 add_bool_arg(
     parser,
-    'gpxq-create-weight-orig',
+    'gptq-create-weight-orig',
```
Giuseppe5, line 306: Why is this `gptq` and not `gpxq`?

i-colbert: `create_weight_orig` is always true for GPFQ. It's currently hard-coded this way for both the TorchVision and Huggingface example entry points, so I changed the name to reflect that this is only currently applied to GPTQ.

Zooming out - I don't think it makes much sense to apply `create_weight_orig=False` for GPFQ or Qronos anyway. This used to make sense from the original implementation of GPFQ with the error matrix, but this new memory-efficient and compute-efficient implementation requires `create_weight_orig=True`.

tests/brevitas/graph/test_gpxq.py
```diff
@@ -1,34 +1,65 @@
...
 from brevitas.graph.magr import magr_mode
+from brevitas.graph.qronos import Qronos
 
 from .equalization_fixtures import *
 
 
-def apply_gpfq(
-        calib_loader: DataLoader, model: nn.Module, act_order: bool, use_quant_activations: bool):
+@torch.no_grad()
+def _mismatched_optimization_callback(
```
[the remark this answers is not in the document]
i-colbert: Good catch - updated.

src/brevitas/graph/gpfq.py
```diff
@@ -258,7 +225,7 @@ def __init__(
             use_quant_activations: bool = True,
             return_forward_output: bool = False,
             act_order: bool = False,
-            gpfq_class: GPFQ = GPFQ) -> None:
+            algorithm_impl: GPFQ = GPFQ) -> None:
```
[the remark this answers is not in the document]
Giuseppe5: No it's fine

src/brevitas_examples/imagenet_classification/ptq/ptq_evaluate.py
Giuseppe5: > Right now, if create_weight_orig=False, then Qronos throws an error, but GPFQ just clones the original weights to try to still do something useful (see [here](https://github.com/Xilinx/brevitas/blob/f0978b2d6df559dff9097d4e3c8ffbbcfd934b96/src/brevitas/graph/gpfq.py#L136)).

I am confused what happens with GPTQ if the flag is true. That is not really creating a clone, it is just creating a pointer to the weights, but that is not the behaviour I expect. In my mind, if create_weight_orig is equal to True, I would expect the behaviour currently in the `else` branch

src/brevitas/graph/qronos.py
```diff
@@ -0,0 +1,257 @@
...
+
+from brevitas.graph.gpfq import GPFQ
+from brevitas.graph.gpxq import SUPPORTED_CONV_OP
+from brevitas.graph.utils import is_conv_transposed
+from brevitas.utils.torch_utils import StopFwdException
+
+
+class Qronos(GPFQ):
+    """
+    Implementation of Qronos as proposed in: https://arxiv.org/pdf/2505.11695
```
nickfraser, line 24: Could you add a docstring, similar to the GPFQ one?

nickfraser (approved):
Just a docstring would be handy, so that documentation for Qronos will be generate for the next doc build, otherwise LGTM!

PR #1320, opened 2025-06-20
Feat (scaling): rescaled min-max scaling and zero point

i-colbert:
## Reason for this PR

Some of the latest post-training quantization research papers consider a scaled min-max grid, where the scaling factor $$s$$ is defined as $$s = \gamma \frac{\max (w) - \min (w)}{2^b - 1}$$ for weights $$w$$ quantized to bit width $$b$$, and $$\gamma \in (0,1]$$

Reproducing the results in the Qronos paper would require this; for example, Table 1: https://arxiv.org/abs/2505.11695

The goal of this PR is to leverage `_AffineStats` to accomplish this.

## Changes Made in this PR

- Extended `_AffineRescaling` to be initialized to a prescribed scalar value for mul and add
- Added rescaling to the zero points
- Added unit test to verify it works as intended

## Testing Summary

Added testing for custom scaled min-max weight quantizers with per-channel and per-tensor scaling.

## Risk Highlight

## Checklist

- [x] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/core/zero_point.py
+++ b/src/brevitas/core/zero_point.py
@@ -119,7 +119,7 @@ class StatsFromParameterZeroPoint(brevitas.jit.ScriptModule):
         _affine_shift_scale = zero_point_affine_shifting_init is not None
         if _affine_shift_scale and not _affine_rescaling:
             raise RuntimeError(
-                "Enabling shifting of the scale requires enabling affine rescaling first.")
+                "Enabling shifting of the zero point requires enabling affine rescaling first.")
         if _affine_rescaling:
             self.affine_rescaling = _AffineRescaling(
                 zero_point_shape,
@@ -321,7 +321,7 @@ class ParameterFromStatsFromParameterZeroPoint(brevitas.jit.ScriptModule):
         _affine_shift_scale = zero_point_affine_shifting_init is not None
         if _affine_shift_scale and not _affine_rescaling:
             raise RuntimeError(
-                "Enabling shifting of the scale requires enabling affine rescaling first.")
+                "Enabling shifting of the zero point requires enabling affine rescaling first.")
         if _affine_rescaling:
             self.affine_rescaling = _AffineRescaling(
                 zero_point_shape,


src/brevitas/core/scaling/runtime.py
```diff
@@ -148,54 +132,59 @@ def __init__(
...
-        if config.IGNORE_MISSING_KEYS and affine_weight_key in missing_keys:
-            missing_keys.remove(affine_weight_key)
-        if config.IGNORE_MISSING_KEYS and affine_bias_key in missing_keys:
-            missing_keys.remove(affine_bias_key)
+        value_key = prefix + 'value'
+        if config.IGNORE_MISSING_KEYS and value_key in missing_keys:
+            missing_keys.remove(value_key)
+
+
+class ConstAffineRescaling(brevitas.jit.ScriptModule):
```
Giuseppe5, line 168: If we pass `affine_rescaling_init` to `_AffineRescaling`, then we do not need this class

Giuseppe5: This has other disadvantages compared to `_AffineRescaling`. For example, since you saved it as a stateless buffer, it won't be picked up in the state dict, which means that saving/loading a checkpoint with this rescaling will have unexpected behaviours

src/brevitas/core/scaling/runtime.py
```diff
@@ -81,21 +75,12 @@ def __init__(
             self,
             restrict_scaling_impl: Module,
             restrict_threshold_impl: Module,
-            scaling_shape: Tuple[int, ...],
-            scaling_min_val: Optional[float],
-            affine_rescaling: bool,
-            affine_shift_scale: bool,
-            dtype: Optional[torch.dtype],
-            device: Optional[torch.device]) -> None:
+            affine_rescaling_impl: Optional[Module] = None,
```
Giuseppe5, line 78: I think this looks like makes things cleaner, but actually it will only complicate stuff at dependency injection time

src/brevitas/core/scaling/runtime.py
```diff
@@ -81,21 +75,12 @@ def __init__(
             self,
             restrict_scaling_impl: Module,
             restrict_threshold_impl: Module,
-            scaling_shape: Tuple[int, ...],
-            scaling_min_val: Optional[float],
-            affine_rescaling: bool,
```
Giuseppe5, line 86: In my mind, the solution would have been to remove these two boolean flag, and apply a similar trick to the one we use for scaling_min_val, where we assume it's None, and if it isn't, then we instantiate the ClampingModule.

In this case, `affine_rescaling_init_val` would become an optional input set to None, and if it is not None, then we could create `_AffineRescaling` class with the init value, otherwise identity as we already do below.

The init value can either be a scalar, that will be automatically broadcasted to scaling_shape, or something else (still compatible with scaling_shape)

src/brevitas/core/scaling/standalone.py
```diff
@@ -253,9 +248,11 @@ def forward(self, x: Tensor, threshold: Optional[Tensor] = None) -> Tensor:
         if threshold is None:
             threshold = torch.ones(1).type_as(x)
         if self.init_done:
-            threshold = self.stats_scaling_impl.restrict_clamp_threshold(
-                self.restrict_threshold_pre(threshold))
-            value = abs_binary_sign_grad(self.stats_scaling_impl.restrict_clamp_scaling(self.value))
+            threshold = self.stats_scaling_impl.restrict_threshold_pre(threshold)
```
Giuseppe5, line 251: I would change this (and also below) to directly call `self.state_scaling_impl`.

Looks like we are doing all the same steps done within the forward pass, but one by one for no particular reason

src/brevitas/core/scaling/runtime.py
```diff
@@ -36,8 +36,9 @@ def __init__(
             force_parameter: bool = False,
             restrict_scaling_impl: Module = FloatRestrictValue(),
             restrict_threshold_impl: Optional[Module] = None,
-            affine_rescaling: bool = False,
-            affine_shift_scale: bool = False,
+            scaling_affine_rescaling_shape: Tuple[int, ...] = SCALAR_SHAPE,
```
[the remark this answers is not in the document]
i-colbert: `scaling_shape` can be per-channel, and `scaling_affine_rescaling_shape` can be per-tensor. This is what we tested in Qronos, and it's sufficient. Same thing for zero points.

src/brevitas/core/scaling/runtime.py
```diff
@@ -81,19 +82,25 @@ def __init__(
...
-            affine_rescaling: bool,
-            affine_shift_scale: bool,
+            scaling_affine_rescaling_shape: Tuple[int, ...],
+            scaling_affine_rescaling_init: Optional[float],
+            scaling_affine_shifting_init: Optional[float],
             dtype: Optional[torch.dtype],
             device: Optional[torch.device]) -> None:
         super(_StatsScaling, self).__init__()
-        if affine_shift_scale and not affine_rescaling:
+        _affine_rescaling: bool = scaling_affine_rescaling_init is not None
```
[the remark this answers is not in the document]
i-colbert: Ok, do you want me to remove these typing hints?

src/brevitas/core/scaling/runtime.py
```diff
@@ -168,21 +176,22 @@ class _AffineRescaling(brevitas.jit.ScriptModule):
...
-            shift_scale,
+            affine_weight_init: Optional[float],
+            affine_bias_init: Optional[float],
             dtype: Optional[torch.dtype],
             device: Optional[torch.device]):
         super(_AffineRescaling, self).__init__()
-        self.affine_weight = Parameter(torch.ones(scaling_shape, dtype=dtype, device=device))
-        if shift_scale:
+        self.affine_weight = Parameter(
+            affine_weight_init * torch.ones(scaling_shape, dtype=dtype, device=device))
```
Giuseppe5, line 185: This init (and the next) seems a bit convoluted to me.

Since the typing of affine_weight_init is `Optional[float]`, we assume that the user will always pass a single scalar value.
Then we can initialize it as we do it for the parameter scaling:
```python
self.affine_weight = torch.full(scaling_shape, affine_weight_init, dtype=dtype, device=device)
```

src/brevitas/core/scaling/runtime.py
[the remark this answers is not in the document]
Giuseppe5: Yes

src/brevitas/core/scaling/runtime.py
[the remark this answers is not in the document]
i-colbert: The `affine_weight` and `affine_bias` are learnable parameters. This allows to either learn different shaped parameters, the default is per-tensor. The parameter is initialized by a scalar but can learn to have non-uniform values (biases are usually initialized in this way too, typically to 0). Exposing the shape enables the option to learn the affine rescaling per-tensor or per-channel.

src/brevitas/core/zero_point.py
```diff
@@ -101,14 +103,33 @@ def __init__(
...
             zero_point_stats_input_concat_dim,
             tracked_parameter_list)
+        _affine_rescaling: bool = zero_point_affine_rescaling_init is not None
+        _affine_shift_scale: bool = zero_point_affine_shifting_init is not None
+        if _affine_shift_scale and not _affine_rescaling:
+            raise RuntimeError(
+                "Enabling shifting of the scale requires enabling affine rescaling first.")
+        if _affine_rescaling:
+            self.affine_rescaling = _AffineRescaling(
+                zero_point_affine_rescaling_shape,
```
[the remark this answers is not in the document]
i-colbert: Done.

PR #1323, opened 2025-06-23
Feat (ex/common): learned round refactor

pablomlago:
### Description

Refactors the learned round optimization framework from a monolithic, parameter-heavy
`LearnedRoundOptimizer` into a modular, dataclass-configured `LearnedRoundTrainer` with
registry-based component dispatch. This makes the system composable, extensible, and
easier to maintain.

Among the new features:
- **Multi-optimizer support**: each `OptimizerArgs` targets a specific parameter group via
  `TargetParamFn`, enabling joint optimization of rounding and scale parameters with
  separate optimizers/LR schedulers.
- **Multi-loss aggregation**: losses are now independent `BlockLoss` implementations
  (e.g., `MSELoss`, `RoundRegularisationLoss`) registered in `BLOCK_LOSS_REGISTRY` and
  aggregated by the trainer, instead of being coupled in a single class.
- **Registry-driven dispatch**: `BLOCK_LOSS_REGISTRY`, `TARGET_PARAM_FN_REGISTRY`,
  `LEARNED_ROUND_INIT_FN_REGISTRY`, and `TRAINING_HANDLERS_REGISTRY` allow extending the
  system by registering new components — no trainer code changes needed.
- **Cache as Dataset**: `Cache` extends `torch.utils.data.Dataset`, enabling standard
  `DataLoader` integration with batching, random sampling, and collate functions.
- **`fast_update` support for LLM**: `CacheLLM` implements `collate_fn_output_next()` and
  `collate_fn_input_next()` for sequential block caching without full model re-inference.

### Tests

- Added parametrized tests for `convert_str_dict` (bools, ints, floats, scientific
  notation, nested dicts).
- Added tests for optimizer and LR scheduler class parsing.
- Added vision learned round test cases (layerwise and blockwise) for `resnet18`.
- Updated existing learned round tests for the new API and module paths.
- Renamed learned round mode strings (`"linear_round"` -> `"identity"`,
  `"hard_sigmoid_round"` -> `"hard_sigmoid"`).

### DIFF
--- a/tests/brevitas_examples/test_vision.py
+++ b/tests/brevitas_examples/test_vision.py
@@ -25,6 +25,7 @@ from tests.brevitas_examples.common import get_default_args
 from tests.brevitas_examples.common import process_args_and_metrics
 from tests.brevitas_examples.common import UpdatableNamespace
 from tests.conftest import SEED
+from tests.marker import skip_on_macos_nox
 
 # TODO (pml): Use stricter tolerance once a proper image dataset is used.
 # The validation set has 6 images, so ATOL_ACC is set to tolerate a single missclasification
@@ -104,6 +105,7 @@ class ImageNetCases:
     def case_small_models_args_and_metrics(self, run_dict, default_run_args, request):
         yield process_args_and_metrics(default_run_args, run_dict, extra_keys=ImageNetCases.METRICS)
 
+    @skip_on_macos_nox
     @pytest_cases.parametrize(
         "run_dict",
         [


Giuseppe5:
Is there a chance to split this into more manageable chunks?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+from brevitas.core.function_wrapper.learned_round import LearnedRoundSte
+from brevitas.inject.enum import LearnedRoundImplType
+from brevitas.optim.sign_sgd import SignSGD
+from brevitas.proxy.parameter_quant import WeightQuantProxyFromInjectorBase
+from brevitas.utils.python_utils import AutoName
+from brevitas_examples.common.learned_round.learned_round_method import BlockLoss
+from brevitas_examples.common.learned_round.learned_round_method import MSELoss
+from brevitas_examples.common.learned_round.learned_round_method import RegularisedMSELoss
+
+OPTIMIZER_MAP = {
```
Giuseppe5: Maybe we should implement a way to register new and/or custom optimizer to this map?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+    "sign_sgd": SignSGD,}
+LR_SCHEDULER_MAP = {}
```
Giuseppe5: Same as above

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+# TODO (pml): Is it possible to remove this boilerplate?
+class TargetParametrizations(AutoName):
+    SCALES = auto()
+    LEARNED_ROUND = auto()
+
+    @property
+    def param_fn(self) -> Callable[[nn.Module, OrderedDict, str], bool]:
+        return {
```
Giuseppe5: Bit ugly, we can have a separate attribute/property with the dict, and this just return `self.pre_dict[self.value]`

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+            TargetParametrizations.LEARNED_ROUND.value: get_round_parameters}[self.value]
+
+
+class BlockLossType(AutoName):
+    MSE = auto()
+    REGULARISED_MSE = auto()
+
+    @property
+    def loss_class(self) -> Type[BlockLoss]:
+        return {
```
Giuseppe5: As above

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+            BlockLossType.MSE.value: MSELoss,
+            BlockLossType.REGULARISED_MSE.value: RegularisedMSELoss}[self.value]
+
+
+# Both `get_round_parameters` and `get_scale_parameters` are meant to be passed as the argument `get_target`
+# of `_get_target_parameters`, which iterates over the modules of a model in a recursive function.
+# In the case of `get_round_parameters` the return value indicates whether the submodules of a given module
```
Giuseppe5: You say:

> In the case of `get_round_parameters`

as if it is different in `get_scale_parameters` but it is not really the case (plus/minus some very small differences), but overall the comment can be adjusted (and shortened)

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+        for param_name, param in module.named_parameters():
+            if param_name.endswith('scaling_impl.value'):
+                state_dict[f"{prefix}.{param_name}"] = param
+        # Early stoppping
+        return True
+    return False
+
+
+#TODO (pml): Add license from Nanotron
+def _convert_str_dict(passed_value: Dict) -> Dict:
```
Giuseppe5: This should go to some util file, maybe python utils?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+            # Check for digit
+            elif value.isdigit():
+                passed_value[key] = int(value)
+            elif value.replace(".", "", 1).isdigit():
+                passed_value[key] = float(value)
+
+    return passed_value
+
+
+def _parse_dataclass_dicts(data_cls, dict_attributes: List[str]) -> None:
```
Giuseppe5: As above

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+            )
+        # Set the updated value
+        setattr(data_cls, attr, kwargs)
+
+
+def _parse_optimizer_class(optimizer_str: str) -> Type[Optimizer]:
+    if optimizer_str in OPTIMIZER_MAP:
+        optimizer_class = OPTIMIZER_MAP[optimizer_str]
+    else:
+        optimizer_keys = [
```
Giuseppe5: I would remove the else and ask the user to register it as mentioned above

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+            optimizer_class = getattr(torch.optim, optimizer_keys[0])
+
+    return optimizer_class
+
+
+def _parse_lr_scheduler_class(lr_scheduler_str: str) -> Type:
+    if lr_scheduler_str in LR_SCHEDULER_MAP:
+        lr_scheduler_class = LR_SCHEDULER_MAP[lr_scheduler_str]
+    else:
+        lr_scheduler_keys = [
```
Giuseppe5: Same as above

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+        metadata={"help": "Scaling factor for the loss."},
+    )
+    use_best_model: bool = field(
+        default=True,
+        metadata={
+            "help":
+                ("Whether to use the best setting of the learned round found "
+                 "during training.")})
+    use_amp: bool = field(
+        default=True,
```
Giuseppe5, line 163: Are we sure we want this true?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+        metadata={"help": "Whether to train using PyTorch Automatic Mixed Precision."})
+    amp_dtype: Union[str, torch.dtype] = field(
+        default=torch.float16,
```
Giuseppe5, line 166: Are we sure we want this to have a default? I'd have the user specify to make sure they know what they're doing

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+        default="identity",
+        metadata={
+            "help": "Defines the functional form of the learned round parametrization.",
+            "choices": [param.value.lower() for param in LearnedRoundImplType]})
+    learned_round_kwargs: Optional[Union[Dict, str]] = field(
+        default=None,
+        metadata={"help": "Extra keyword arguments for the learned round parametrization."},
+    )
+    fast_update: bool = field(
+        default=True, metadata={"help": ("Whether to use fast update with learned round.")})
```
Giuseppe5: Add a line to say what's the drawback, if any.
If there are none, why would I never not use it?

src/brevitas_examples/common/learned_round/learned_round_trainer.py
```diff
@@ -318,6 +287,25 @@ def __call__(self, module, args, kwargs, output) -> None:
         raise StopFwdException
 
 
+def get_blocks(model: nn.Module, block_check_fn: Callable[[nn.Module, str],
```
Giuseppe5, line 223: I believe we might have something similar to this for quantization

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
```diff
@@ -351,97 +339,264 @@ def save_inputs_output(
...
-                lr_scheduler.step()
 
-    def _step(self, optimizers: List[Optimizer], lr_schedulers: List[Any]) -> None:
-        for optimizer in optimizers:
-            if optimizer:
-                optimizer.step()
-                optimizer.zero_grad()
-        for lr_scheduler in lr_schedulers:
-            if lr_scheduler:
+class LearnedRoundOptimizer:
```
Giuseppe5: This should be a Trainer class of some sorts

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,345 @@
...
+        # Parse in args that could be `dict` sent in from the CLI as a string
+        _parse_dataclass_dicts(self, self._DICT_ATTRIBUTES)
+        # Parse string to learning rate scheduler class if needed
+        self.lr_scheduler_cls = (
+            _parse_lr_scheduler_class(self.lr_scheduler_cls) if isinstance(
+                self.lr_scheduler_cls, str) else self.lr_scheduler_cls)
+
+
+@dataclass
+class OptimizerArgs:
```
Giuseppe5, line 82: Split optimizer/scheduler

src/brevitas_examples/common/learned_round/learned_round_trainer.py
```diff
@@ -351,97 +339,264 @@ def save_inputs_output(
...
+
+    def _get_all_target_parameters(self, model: nn.Module) -> OrderedDict:
+        state_dict = OrderedDict()
+        # Iterate over the target parameters
+        for optimizer_target in self.config.training_args.optimizers_targets:
+            # `get_target_paramets` modifies state_dict in-place
+            state_dict = self._get_target_parameters(model, optimizer_target.param_fn, state_dict)
+        return state_dict
+
+    def _create_optimizer_and_scheduler(
```
Giuseppe5, line 426: split optimizer/scheduler

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
```diff
@@ -351,97 +339,264 @@ def save_inputs_output(
...
+        optimizer = optimizer_args.optimizer_cls(
+            params=params, lr=optimizer_args.lr, **optimizer_args.optimizer_kwargs)
+        # Instantiate learning rate schedulers
+        lr_scheduler_args = optimizer_args.lr_scheduler_args
+        lr_scheduler = (
+            lr_scheduler_args.lr_scheduler_cls(optimizer, **lr_scheduler_args.lr_scheduler_kwargs)
+            if lr_scheduler_args is not None else None)
+        return optimizer, lr_scheduler
+
+    def _create_optimizers_lr_schedulers(
```
Giuseppe5: name of the function is confusingly similar with the one above

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I opted for removing this dict, and having the optimizers retrieved from either `torch.optim` or `brevitas.optim`.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I opted for removing this dict since, at the moment, there is no plan to introduce a custom one. Moreover, this could be registered in `brevitas.optim.lr_scheduler` and use a logic similar as the one for the optimizers.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I created a `Registry` class in `python_utils.py`, and I use that for tracking/retrieving the different implementations.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I polished it a bit to make it clearer that it is an example.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: The drawback is that it requires implementing extra methods in the `Cache` and it is not valid for every topology.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I'm not sure if the case is strong enough for that, a lr scheduler instance is semantically tied to an optimizer, so it is sensible for the dataclass hierarchy to reflect this.

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
pablomlago: I opted for removing it, after the changes to the optimizer instantiation.

src/brevitas_examples/common/learned_round/learned_round_trainer.py
pablomlago: Same as above.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: This was the previous default value, so we might want to leave it as it is for retrocompatibility.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: Same as above.

src/brevitas_examples/common/learned_round/learned_round_method.py
```diff
@@ -36,11 +47,24 @@ def format_loss_components(self, *args) -> str:
         pass
 
 
+# Registries for implementations of learned round components
+BLOCK_LOSS_REGISTRY = Registry[Type[BlockLoss]]('BlockLoss registry')
```
Giuseppe5: I kind of hate the full uppercase stuff, although I understand it since it, but I'm leaning towards all lower case names

src/brevitas_examples/common/learned_round/learned_round_args.py
Giuseppe5: In the context of examples, retrocompatibility can be sacrified if we think it is worth it.
We could consider it for this case since `use_amp=True` stresses me out a bit

src/brevitas_examples/common/learned_round/learned_round_args.py
Giuseppe5: Then I guess the default should be False, since that is the safest option?

src/brevitas/utils/python_utils.py
```diff
@@ -64,3 +75,89 @@ def run(*args, **kwargs):
...
+                f"Value set for attribute {attr} of dataclass {type(data_cls).__name__} cannot be converted into a dictionary."
+            )
+        # Set the updated value
+        setattr(data_cls, attr, kwargs)
+
+
+T = TypeVar("T")
+
+
+class Registry(Generic[T]):
```
Giuseppe5, line 144: Interesting class, I'm assuming we could re-use it also for other stuff like registering ops for __torch_function__ in our quant tensors?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,248 @@
...
+from brevitas import optim
+from brevitas.inject.enum import LearnedRoundImplType
+from brevitas.utils.python_utils import parse_dataclass_dicts
+from brevitas_examples.common.learned_round.learned_round_method import BLOCK_LOSS_REGISTRY
+from brevitas_examples.common.learned_round.learned_round_method import BlockLoss
+from brevitas_examples.common.learned_round.learned_round_method import TARGET_PARAM_FN_REGISTRY
+from brevitas_examples.common.learned_round.learned_round_method import TargetParamFn
+
+OPTIMIZER_NAMESPACES = [torch.optim, optim]
+LR_SCHEDULER_NAMESPACES = [torch.optim.lr_scheduler]
```
Giuseppe5, line 26: Not amazing, not awful.  We could also consider removing this and just check the value when needed

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,248 @@
...
+            if (
+                # Verify that the key stars with the one passed by the user
+                optimizer_key.lower() == optimizer_str.lower() and
+                # Verify that key corresponds to a class
+                isinstance(namespace.__dict__[optimizer_key], type) and
+                # Make sure the abstract class is not used
+                optimizer_key != "Optimizer" and
+                # An optimizer implements zero_grad and step. Check that this
+                # is the case for the class retrieved from torch.optim
+                hasattr(namespace.__dict__[optimizer_key], 'step') and
```
Giuseppe5: I would be also of the opinion that we skip these checks.
If the user is doing something weird, stuff will fail anyway

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,248 @@
...
+            for optimizer_key in namespace.__dict__.keys()
+            # Make sure that only valid optimizer implementations are
+            # retrieved, when matching the string passed by the user
+            if (
+                # Verify that the key stars with the one passed by the user
+                optimizer_key.lower() == optimizer_str.lower() and
+                # Verify that key corresponds to a class
+                isinstance(namespace.__dict__[optimizer_key], type) and
+                # Make sure the abstract class is not used
+                optimizer_key != "Optimizer" and
```
Giuseppe5: An abstract class can't be instantiated on its own, again maybe we're checking too many things

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,248 @@
...
+            (namespace, lr_scheduler_key)
+            for lr_scheduler_key in torch.optim.lr_scheduler.__dict__.keys()
+            # Check for making sure that only valid LRScheduler implementations are
+            # retrived, when matching with the string passed by the user
+            if
+            ((
+                lr_scheduler_key.lower() == lr_scheduler_str.lower() or
+                lr_scheduler_key.lower() == lr_scheduler_str.lower() + "lr") and
+             # Verify that key corresponds to a class
+             isinstance(torch.optim.lr_scheduler.__dict__[lr_scheduler_key], type) and
```
Giuseppe5: Same as above, I'd limit the checks to the bare minimum and let it crash and burn in other cases.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -111,55 +140,94 @@ def get_blocks(model: nn.Module, block_name_attribute: str) -> List[nn.Module]:
...
+    for key, value in kwargs.items():
+        if isinstance(value, list) and len(value) > 0:
+            kwargs[key] = torch.cat(kwargs[key], dim=0)
+    return kwargs
+
+
+# TODO (pml): Transition to `args` being a nested dictionary, which is translated
+# an validated to `Config`` (e.g. using the package dacite)
+def parse_args_to_dataclass(args: Namespace) -> Config:
+    from brevitas_examples.common.learned_round.learned_round_args import LearnedRoundArgs
```
Giuseppe5: Why the local imports?

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -0,0 +1,248 @@
...
+        # Parse in args that could be `dict` sent in from the CLI as a string
+        parse_dataclass_dicts(self, self._DICT_ATTRIBUTES)
+
+        self.learned_round_param = LearnedRoundImplType(
+            self.learned_round_param.upper()) if isinstance(
+                self.learned_round_param, str) else self.learned_round_param
+
+
+@dataclass
+class Config:
```
Giuseppe5: Seems overly specific for just these two elements.
I'd probably keep them separate as a tuple or something like that for now

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -111,55 +140,94 @@ def get_blocks(model: nn.Module, block_name_attribute: str) -> List[nn.Module]:
...
-        lr_scheduler_kwargs=lr_scheduler_kwargs,
-        scale_optimizer_kwargs=scale_optimizer_kwargs,
-        scale_optimizer_class=scale_optimizer_class)
+        model: nn.Module, calibration_dataset: torch.utils.data.Dataset, args: Namespace) -> None:
     cache = CacheLLM()
+    llm_block_check_fn = functools.partial(get_blocks, block_name_attribute=args.gpxq_block_name)
+
+    config = parse_args_to_dataclass(args)
+    learned_round_optimizer = LearnedRoundTrainer(config=config)
+    calibration_loader = DataLoader(
```
Giuseppe5: In theory now we have a dataloader with a batch size that we could use it, correct?

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -111,55 +140,94 @@ def get_blocks(model: nn.Module, block_name_attribute: str) -> List[nn.Module]:
...
-        optimizer_kwargs=optimizer_kwargs,
-        lr_scheduler_kwargs=lr_scheduler_kwargs,
-        scale_optimizer_kwargs=scale_optimizer_kwargs,
-        scale_optimizer_class=scale_optimizer_class)
+        model: nn.Module, calibration_dataset: torch.utils.data.Dataset, args: Namespace) -> None:
     cache = CacheLLM()
+    llm_block_check_fn = functools.partial(get_blocks, block_name_attribute=args.gpxq_block_name)
+
+    config = parse_args_to_dataclass(args)
+    learned_round_optimizer = LearnedRoundTrainer(config=config)
```
Giuseppe5: change name of variable

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -85,18 +114,18 @@ def llm_learned_round_prepare_fn(model: nn.Module) -> None:
     return llm_cache_state
 
 
-def llm_learned_round_finish_fn(model: nn.Module, llm_cache_state: Dict) -> None:
+def llm_learned_round_finish_fn(model: nn.Module, llm_cache_state: bool) -> None:
     model.config.use_cache = llm_cache_state
 
 
-def llm_forward(model: nn.Module, inputs: Any) -> None:
+def llm_forward(model: nn.Module, inputs: Dict[str, Any]) -> None:
```
Giuseppe5, line 117: This could be an inline function within `apply_learned_round`, same for `llm_block_forward`

src/brevitas_examples/common/learned_round/learned_round_trainer.py
Giuseppe5, line 1: Change the name of the file

src/brevitas_examples/common/learned_round/learned_round_trainer.py
```diff
@@ -182,112 +182,82 @@
...
-                    if submodule_name.endswith('scaling_impl.value'):
-                        scale_parameters.append(submodule)
-            else:
-                _get_scale_parameters(module_child)
 
-    # Run recursion from block
-    _get_scale_parameters(block)
-    return scale_parameters
+# Cache, as a subclass of torch.utils.data.Dataset, needs to implement __getitem__ and __len__
+class Cache(Generic[_T_inputs, _T_outputs], Dataset[_T_cache]):
```
Giuseppe5: These should probably go in a new `learned_round_utils.py` file in the same folder.
In a perfect world, the `learned_round_utils` inside ptq or llm_quant would change/specialize stuff present in `common/learned_round/learned_round_utils.py`

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
```diff
@@ -479,168 +632,42 @@ def _populate_cache(
...
-            # Override if the model with the lowest training error is not used
-            best_loss = curr_loss
-            last_best_iter = self.iters
-
-        # Move the block back to CPU
-        block.cpu()
-
-        return init_loss, best_loss, last_best_iter
-
     def apply_learned_round(
```
Giuseppe5: This should be called something like optimize

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -85,18 +114,18 @@ def llm_learned_round_prepare_fn(model: nn.Module) -> None:
     return llm_cache_state
 
 
-def llm_learned_round_finish_fn(model: nn.Module, llm_cache_state: Dict) -> None:
+def llm_learned_round_finish_fn(model: nn.Module, llm_cache_state: bool) -> None:
```
Giuseppe5: Both this and `model_finish_fn` should be called something like `clean_up` or something along these lines

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
```diff
@@ -654,115 +681,102 @@ def apply_learned_round(
...
-            if block_idx + 1 < len(blocks) and fast_update:
+            if block_idx + 1 < len(blocks) and self.config.learned_round_args.fast_update:
                 cache = self.skip_full_execution(block, blocks[block_idx + 1], block_forward, cache)
 
         # The original configuration of the model is restored after finishing the optimization
         if model_finish_fn is not None:
             model_finish_fn(model, model_dict)
 
-    def skip_full_execution(self, block, next_block, block_forward, cache):
+    def skip_full_execution(
```
Giuseppe5: This should probably be called `populate_cache` or `update_cache`

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -1,63 +1,73 @@
...
-from brevitas_examples.common.learned_round.learned_round_parser import \
-    parse_learned_round_loss_class
-from brevitas_examples.common.learned_round.learned_round_parser import parse_lr_scheduler_class
-from brevitas_examples.common.learned_round.learned_round_parser import parse_optimizer_class
+from brevitas_examples.common.learned_round.learned_round_optimizer import LearnedRoundTrainer
 
+_T_args = Tuple[torch.Tensor, ...]
+_T_kwargs = Dict[str, Any]
+_T_inputs = Tuple[_T_args, _T_kwargs]
+_T_outputs = torch.Tensor
```
Giuseppe5: Maybe not needed but I'll allow it for coherence

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: What do you mean by "check the value when needed"?

src/brevitas/utils/python_utils.py
pablomlago: Yes, I do not see any drawback in using this class in those scenarios.

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: `Optimizer` in `torch.optim` is not exactly an abstract class (it does not inherit from `ABC`).

src/brevitas_examples/common/learned_round/learned_round_args.py
pablomlago: I would be up for keeping it as it, since it can easily grow if more optimization targets are added.

src/brevitas_examples/common/learned_round/learned_round_method.py
pablomlago: I think these fall under the [Constants](https://peps.python.org/pep-0008/#constants) naming convention, right? If so, I'd rather keep them as they are.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
pablomlago: I felt that, otherwise, the signatures were too verbose.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
pablomlago: Not needed anymore, it was to delay an import before.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
pablomlago: I don't see a clear reason to define it inside `apply_learned_round`, as it seems to me that it would clutter the function unnecessarily.

src/brevitas_examples/common/learned_round/learned_round_optimizer.py
pablomlago: Renamed it to `train`.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
pablomlago: I'll remove these functions, and convert it instead to a context manager in the specific `learned_round_utils.py` of each entrypoint.

src/brevitas_examples/common/learned_round/learned_round_method.py
Giuseppe5: Do they? You are adding new element in the register so it not strictly a constant

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
Giuseppe5: mean the last one in particular, `_T_outputs`.
The others one are perfect

src/brevitas/utils/python_utils.py
```diff
@@ -64,3 +75,88 @@ def run(*args, **kwargs):
...
```
Giuseppe5, line 144: Rebase on dev to solve this conflict

src/brevitas_examples/common/learned_round/learned_round_trainer.py
Giuseppe5: Are we doing this?

src/brevitas_examples/common/learned_round/learned_round_trainer.py
```diff
@@ -351,97 +272,249 @@ def save_inputs_output(
...
+    def _get_target_parameters(
+        self,
+        model: nn.Module,
+        get_target: TargetParamFn,
+        state_dict: Optional[OrderedDict[str,
+                                         nn.Parameter]] = None) -> OrderedDict[str, nn.Parameter]:
+
+        state_dict = OrderedDict() if state_dict is None else state_dict
+
+        def _recursive_get_target_parameters(module: nn.Module, prefix: str = "") -> None:
```
Giuseppe5, line 404: I believe we don't need the inner `_recursive` function, it should be possible to flatten it out in the previous one

src/brevitas_examples/imagenet_classification/ptq/learned_round_utils.py
```diff
@@ -132,66 +120,99 @@ def vision_block_forward(block: nn.Module, inputs: Any) -> torch.Tensor:
...
+                "total_iters": args.learned_round_iters,},
+        )
+
+    optim_args = OptimizerArgs(
+        target_params="learned_round",
+        optimizer_cls="Adam",
+        lr=args.learned_round_lr,
+        optimizer_kwargs={},
+        lr_scheduler_args=lr_scheduler_args,
+    )
```
Giuseppe5: can you remove the last comma? The formatter will take care of fixing the parenthesis.
From here and everywhere else applicable

src/brevitas/utils/python_utils.py
```diff
@@ -75,13 +78,59 @@ def run(*args, **kwargs):
...
+    "Safely checks that a passed value is a dictionary and converts any string values to their appropriate types."
+    for key, value in passed_value.items():
+        if isinstance(value, dict):
+            passed_value[key] = convert_str_dict(value)
+        elif isinstance(value, str):
+            # First check for bool and convert
+            if value.lower() in ("true", "false"):
+                passed_value[key] = value.lower() == "true"
+            # Check for digit
+            elif value.isdigit():
```
Giuseppe5: Instead of `isdigit` and `value.replace(".", "", 1).isdigit()`, which I'm not really sure what it does, maybe you can consider using `eval`?
```python
type(eval('4.5'))
# float
```
```python
type(eval('4'))
# int
```

src/brevitas_examples/common/learned_round/learned_round_method.py
```diff
@@ -148,41 +157,103 @@ def __init__(
...
+    def __call__(self, pred: torch.Tensor, tgt: torch.Tensor) -> torch.Tensor:
+        if self.reduction is None:
+            return F.mse_loss(pred, tgt)
+        return F.mse_loss(pred, tgt, reduction=self.reduction).sum(self.dim).mean()
+
+
+# Both `get_round_parameters` and `get_scale_parameters` are meant to be passed as the argument `get_target`
+# of `_get_target_parameters`, which iterates over the modules of a model in a recursive function.
+# The return value indicates whether the submodules of a given module need to be skipped.
+# For instance, for `get_round_parameters`, when a LearnedRoundSte module is found, the
```
Giuseppe5: We can be less verbose with the comments, removing everything that comes after "For instance"

src/brevitas_examples/common/learned_round/learned_round_method.py
```diff
@@ -148,41 +157,103 @@ def __init__(
...
-    def __call__(self, pred: torch.Tensor, tgt: torch.Tensor) -> Tuple[torch.Tensor, Tuple]:
+    def __call__(self, pred: torch.Tensor, tgt: torch.Tensor) -> torch.Tensor:
         self.iter += 1
 
-        rec_loss = F.mse_loss(pred, tgt, reduction='none').sum(1).mean()
-
         if self.iter < self.loss_start:
             b = self.temp_decay(self.iter)
-            round_loss = 0.
+            round_loss = torch.tensor(0., device=pred.device, dtype=torch.float32)
```
Giuseppe5: Why do we need explicit float32 cast?

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
```diff
@@ -111,63 +136,114 @@ def get_blocks(model: nn.Module, block_name_attribute: str) -> List[nn.Module]:
...
+            config=learned_round_args,
+        )]
+
+    return TrainerConfig(
+        training_args=training_args,
+        training_handlers=training_handlers,
+    )
+
+
+class llm_trainer_cm:
```
Giuseppe5, line 220: A context manager to disable/enable `use_cache` maybe be a bit overkill but not strongly opinionated about this

src/brevitas/utils/python_utils.py
pablomlago: I'd rather not use `eval`, as it can be used to run malicious code.

src/brevitas_examples/common/learned_round/learned_round_method.py
pablomlago: I changed it to retrieve the `dtype` of the weights.

src/brevitas_examples/llm/llm_quant/learned_round_utils.py
pablomlago: I think I'll probably leave as it is, as it is not much extra code, and might be useful if extra preparations are needed.

src/brevitas_examples/common/learned_round/learned_round_trainer.py
pablomlago: Iterating over `model.named_modules()` would require changing the logic to skip the children of a given module, as every children needs to be skipped. An alternative would be to do a DFS/BFS traversal, but I think the recursive version is more readable.

src/brevitas_examples/imagenet_classification/ptq/learned_round_utils.py
pablomlago: I think in this context it is useful to have the trailing comma (e.g. clearer diffs when modifying the parameters of these dataclasses), so I'd rather keep them.

tests/brevitas/utils/test_python_utils.py
```diff
@@ -115,3 +118,71 @@ def test_get_missing_raises_valueerror(self):
...
         assert msg == "'missing' not found in TestRegistry. The registered keys are: k1, k2"
+
+
+class TestConvertDictStringVals:
+
+    @pytest.mark.parametrize(
+        "value, expected_type, expected_value",
+        [
+            # bools
+            ("true", bool, True),
```
Giuseppe5, line 129: Better safe than sorry I guess

PR #1334, opened 2025-07-07
Rotation fix

Giuseppe5:
## Reason for this PR

Starting from a bug where `non valid regions ` were not skipped in rotations, I restructured the code in a way that non-valid regions are never added in the pool in the first place.

On top of that, more debug logging so that it is possible for the user to check if everything is where it is supposed to be

## Testing Summary

Current testing should be enough.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

cc @i-colbert

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1283,6 +1283,7 @@ class LayerwiseActivationEqualization(ActivationEqualization):
         scale_factors = []
         rewriters = []
         self.remove_hooks()
+        logging.debug(f"Applying LayerwiseActivationEqualization on {len(self.regions)} regions")
         for region in self.regions:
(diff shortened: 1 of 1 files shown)


tests/brevitas/graph/equalization_fixtures.py
```diff
@@ -35,7 +35,7 @@
     'resnet18': [0.487, 0.952],
     'googlenet': [0.495, 0.982],
     'inception_v3': [0.497, 0.989],
-    'alexnet': [0.875, 0.875],}
+    'alexnet': [0.75, 0.75],}
```
Giuseppe5, line 38: Earlier we were counting also the regions that we would then skip during the actual equalization process.
Now we're counting only the effective equalized region, so we have one less

PR #1338, opened 2025-07-15
Fix (eq/rotation): find value if passed as kwarg

nickfraser:
During rotation equalization, `value` is checked as the last passed `arg`, but if is not found if `value` is passed as a `kwarg`. Minor correction which fixes that.

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1835,7 +1835,7 @@ class GraphRotationEqualization(RotationEqualization):
             if sdpa_node.kwargs.get('value', False):
                 value_input = sdpa_node.kwargs['value']  # value passed as kwarg
             else:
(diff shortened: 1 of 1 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -1832,7 +1832,10 @@ def find_sink(node):
                 return output_node
 
         for sdpa_node in sdpa_nodes:
-            value_input = sdpa_node.args[-1]
+            if sdpa_node.kwargs.get('value', False):
+                value_input = sdpa_node.kwargs['value']  # value passed as kwarg
+            else:
+                value_input = sdpa_node.args[-1]  # value passed as last arg
```
Giuseppe5, line 1838: It could be worth checking if we can fully match the signature of the function (we do it for the smoothquant hook), so that we're sure we're picking the correct arg/kwargs.

I am not sure if that works well also with FX nodes though

PR #1347, opened 2025-08-05
Fix (core/scaling): handle edge cases with signed scale

Giuseppe5:
## Reason for this PR

We currently do not handle signed scale in the following scenarios:
- Log restriction (either po2 or log)
- Learned scale, when the scale can start positive and then become negative. In this case, the user might want to force the scale to stay positive for the entire QAT process

## Changes Made in this PR

At dep inj time, we verify if the scale must be signed or unsigned.
It must be unsigned if:
- The user uses Log restriction
- The user forces through a specific dep-inj flag
- The user picks an unsigned stats op (all current available stats)

It can be signed if:
- The user does not use any log restriction for the scale
- The user picks a signed stats op  (not yet implemented)

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/core/restrict_val.py
+++ b/src/brevitas/core/restrict_val.py
@@ -30,7 +30,8 @@ class _RestrictClampValue(brevitas.jit.ScriptModule):
     def __init__(
             self,
             scaling_min_val: Optional[float] = None,
-            restrict_value_impl: Optional[Module] = None):
+            restrict_value_impl: Optional[Module] = None,
+            is_unsigned=True):
         super(_RestrictClampValue, self).__init__()
         if scaling_min_val is not None and scaling_min_val != 0:
             self.clamp_min_ste = ScalarSignedClampMinSte(scaling_min_val)
@@ -41,8 +42,14 @@ class _RestrictClampValue(brevitas.jit.ScriptModule):
         else:
             self.restrict_value_impl = Identity()
 
+        if is_unsigned:
+            self.apply_abs = torch.abs
+        else:
+            self.apply_abs = Identity()
+
     @brevitas.jit.script_method
     def forward(self, x: Tensor):
+        x = self.apply_abs(x)
         x = self.restrict_value_impl(x)
         x = self.clamp_min_ste(x)
         return x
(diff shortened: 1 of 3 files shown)


src/brevitas/quant/base.py
```diff
@@ -300,6 +302,7 @@ class WeightPerTensorFloatDecoupledL2Param(SolveWeightScalingStatsInputDimsFromM
     scaling_shape = SCALAR_SHAPE
     scaling_per_output_type = ScalingPerOutputType.TENSOR
     input_view_impl = Identity
+    scaling_impl_type = ScalingImplType.PARAMETER_FROM_STATS
```
i-colbert, line 305: why specify this here when it is specified below?

Giuseppe5: The new solver for Signedness explicitly requires `type` and not just `impl` for `scaling_impl`.
The reason is that `scaling_impl_type` has fewer options than `scaling_impl`, which makes it easier to check which case we are in.


In general, the logic is the following:
- If the user is using a `scaling_impl_type` that starts from statistics (stats, parameter_from_stats), then we check if the statics op is signed or unsigned
- Otherwise, we check for `scaling_init` and a user-defined flag `force_signed_scale`, to decide if the scale is signed or unsigned


If we don't want to add this as extra constraint, there are two alternative:
- Expand the logic of the new Signedness solver to account for either `scaling_impl_type` or `scaling_impl`
- Rewrite the logic, checking first if `scaling_init` is defined (which is required for any `scaling_impl` that does not start from stats, like `Const` or `Parameter`), and then having the else branch to check `scaling_stats_impl`, assuming that it must exist in all the other cases. This way we would not have to check for `scaling_impl` or `scaling_impl_type`

The drawback of option 1 is complexity.
The drawback of option 2 is that we assume that the user will never define `scaling_init` unless it is really needed and used. Dep-inj does not protect us against defining unused arguments, that have no effect on the final quantizer.

nickfraser (approved):
LGTM! (after precommit)

PR #1357, opened 2025-08-22
Feat (benchmark): Allow easy specification of different search algorithms within benchmark scripts

nickfraser:
Builds on #1356.

## Overview

Refactors the benchmark utilities into a composition-based design and adds a
random-search strategy for generating experiments.

Previously, entry-point specifics and search behaviour were combined through
inheritance. This PR splits those responsibilities into two composable members
coordinated by `BenchmarkUtils`:

- **`EntryPointUtils`** — everything specific to a benchmark entry point:
  `parse_log`, `validate`, `entrypoint_main`, `argument_parser`, `eval_metrics`.
- **`SearchUtils`** — how the experiment search space is generated:
  `script_parser`, `gen_search_space`, `print_benchmark_summary`,
  `_default_action_handler`.
- **`BenchmarkUtils`** — the coordinator: holds an `EntryPointUtils` and a
  `SearchUtils` as members, owns config reading (`standardize_args`) and result
  parsing (`parse_results`), and drives execution (`run`). It brokers the data
  crossing between the two members (the entry point's argument parser and
  validation into the search; the generated experiment queue back into the entry
  point).

A concrete benchmark is now defined by composition rather than multiple
inheritance:

```python
class LLMEntryPointUtils(EntryPointUtils):
    argument_parser = create_llm_args_parser()
    eval_metrics = ["float_ppl", "quant_ppl"]
    # parse_log / validate / entrypoint_main ...

class LLMGridBenchmark(BenchmarkUtils):
    entry_point_utils = LLMEntryPointUtils
    search_utils = GridSearchUtils

class LLMRandomBenchmark(BenchmarkUtils):
    entry_point_utils = LLMEntryPointUtils
    search_utils = RandomSearchUtils
```

All existing benchmark entry points — LLM, Stable Diffusion, ImageNet PTQ, and
the `papers/` examples — were migrated to this interface.

## Search strategies

Two `SearchUtils` implementations are provided:

- **`GridSearchUtils`** replicates the behaviour of the benchmark utils *before*
  this PR (cartesian product over lists of argument values).
- **`RandomSearchUtils`** allows specification of various search distributions
  per parameter, for example:

```yaml
act_equalization:
  rand_type: choices
  rand_values: [null, "layerwise", "fx"]
act_equalization_alpha:
  rand_type: linear
  rand_values: [0.05, 0.95]
gptq:
  rand_type: const
  rand_values: true
learned_round_lr:
  rand_type: log2
  rand_values: [0.0001, 0.1]
learned_round_scale_lr:
  rand_type: exp2
  rand_values: [0.0001, 0.1]
```

### Random node types

Each parameter's `rand_type` selects how its value is sampled. These are
implemented as a small `RandomArgNode` class hierarchy, auto-registered and
instantiated from the YAML via `RandomArgNode.from_config`:

- `const` — a fixed value.
- `choices` — uniform choice from a list.
- `linear` — uniform sample in `[min, max]`.
- `log2` — log-uniform sample in `[min, max]` (requires strictly positive bounds).
- `exp2` — uniform sample in `2**[min, max]` space, then `log2`.

A helper script, `cfg2randbenchmarkcfg.py`, converts a plain argument config into
a random-search template by wrapping every value as a `const` node.

Adding a new node type is self-contained: subclass `RandomArgNode` (or
`_RangeNode` for `[min, max]` types), set its `rand_type` class attribute, and
implement `value()` (override `__str__` if the default description isn't
suitable). Subclasses are auto-registered by `rand_type`, so `from_config` picks
them up with no changes to the parsing logic. For example:

```python
class NormalNode(RandomArgNode):
    rand_type = "normal"

    def value(self):
        mean, std = self.rand_values
        return random.gauss(mean, std)
```

## Usage

Running as follows:

```bash
python llm_rand_benchmark.py --config benchmark_rand_template.yaml --dry-run --seed 1 --num-experiments 10
```

Will give the following output:

```text
Num. experiments: 10
Benchmark args.:
{'config': 'benchmark_rand_template.yaml',
 'results_folder': './',
 'gpus': '0',
 'num_gpus_per_process': 1,
 'max_num_retries': 1,
 'dry_run': True,
 'num_experiments': 10,
 'max_experimental_configs': 100000,
 'seed': 1}
Non-default args.:
{'--act-equalization': "type: choices, values: [None, 'layerwise', 'fx']",
 '--act-equalization-alpha': 'type: linear, min: 0.05, max: 0.95',
 '--gptq': 'type: const, value: True',
 '--learned-round-lr': 'type: log2, min: 0.0001, max: 0.01',
 '--learned-round-scale-lr': 'type: exp2, min: 0.0001, max: 0.01'}
```

## Verification

As an end-to-end check of the benchmark utilities, we reproduced the QuaRot\*
down-projection (R4) expansion sweep from
[Improving Quantization with Post-Training Model Expansion](https://arxiv.org/abs/2503.17513)
(Table 1, Llama 3.2 1B, INT4 W4A4) via `papers/expansion/quarot_star.yaml`, and
compared WikiText2 perplexity (lower is better) against the paper. The
`expansion_step` sweep (`0, 1, 3, 5, 7`) maps in order to the paper's
down-projection expansion ratios.

| Expansion (down-proj ratio) | `expansion_step` | Paper PPL | Reproduced PPL |
|---|---|---|---|
| Float (reference) | –  | 8.94  | 8.94  |
| 1.00 (baseline)   | 0  | 12.94 | 12.93 |
| 1.10              | 1  | 12.75 | 12.62 |
| 1.20              | 3  | 12.75 | 12.66 |
| 1.30              | 5  | 12.56 | 12.66 |
| 1.50              | 7  | 12.56 | 12.43 |

Reproduced perplexities track the paper within ~0.13 PPL, and expansion
consistently improves over the no-expansion baseline.

## Notes

Note, a limitation with the current approach is that the worker queue generation
(i.e., the random search space) and execution (i.e., the experiments) are two
separate steps, meaning that it's difficult to integrate experiment feedback into
the search space selection (e.g., for Bayesian optimization or simulated
annealing).

Tests have been added in `tests/brevitas_examples/test_benchmark_utils.py`
(entry-point-agnostic: search strategies, `RandomArgNode`, worker process, and
end-to-end orchestration) and `tests/brevitas_examples/test_llm_benchmark.py`
(LLM grid and random search). The non-LLM benchmark tests are disabled on Mac OSX
due to [this issue](https://github.com/python/cpython/issues/101094); if it's
important to have this function on Mac OSX, we could implement the workaround
[here](https://github.com/keras-team/autokeras/issues/368#issuecomment-461625748).
`pandas` was added to the examples test environment and the LLM benchmark tests
were wired into `noxfile.py`.

Additionally, `llm_args` validation now asserts (instead of warning) when
GPTQ/GPFQ/Qronos are enabled together, so those combinations are filtered out of
the generated search space.


src/brevitas_examples/llm/llm_args.py
```diff
@@ -490,7 +490,7 @@ def validate(args: Namespace, extra_args: Optional[List[str]] = None) -> None:
             assert args.weight_quant_rescaling_init > 0, \
                 'Error: weight_quant_rescaling_init must be positive.'
         if (int(args.gptq) + int(args.gpfq) + int(args.qronos)) > 1:
-            warn("GPTQ, GPFQ, and/or Qronos are enabled together.")
+            assert False, "GPTQ, GPFQ, and/or Qronos are enabled together."
```
nickfraser: Let me know if you prefer me to revert this. In my view, this should be fail instead of a warning. Could also be a separate PR.

src/brevitas_examples/llm/benchmark/llm_rand_benchmark.py
```diff
@@ -0,0 +1,58 @@
...
+from brevitas_examples.llm.llm_args import validate as validate_llm_args
+
+
+class LLMRandomSearchBenchmarkUtils(RandomSearchBenchmarkUtils):
+
+    argument_parser: ArgumentParser = create_llm_args_parser()
+    eval_metrics: List[str] = ["float_ppl", "quant_ppl"]
+
+    @staticmethod
+    def parse_log(job_log: str) -> Dict[str, Any]:
```
nickfraser: Move to `LLMBenchmarkUtilsMixin`?

nickfraser (commented):
Switch to Mixin.

src/brevitas_examples/llm/llm_args.py
pablomlago: I'm fine being such a small change. Nit:
```suggestion
            assert (int(args.gptq) + int(args.gpfq) + int(args.qronos)) <= 1, "GPTQ, GPFQ, and/or Qronos cannot be enabled together."
```

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +62,387 @@ def eval_metrics() -> List[str]:
...
+    @classmethod
+    @abstractmethod
+    def print_benchmark_summary(cls, args_queue: List[Dict], script_args: Namespace) -> None:
+        pass
+
+
+class GridSearchMixin(BenchmarkSearchMixin):
+
+    @classmethod
+    def standardize_args(cls, script_args: str) -> Dict[str, List]:
```
pablomlago:
```suggestion
    def standardize_args(cls, script_args: Namespace) -> Dict[str, List]:
```

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +62,387 @@ def eval_metrics() -> List[str]:
         pass
 
 
+class BenchmarkSearchMixin(ABC):
```
pablomlago, line 67: Indicate that the concrete classes need to provide the abstract class variable `argument_parser`, e.g.
```
    @property
    @abstractmethod
    def argument_parser(self) -> ArgumentParser:
        pass
```

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +62,387 @@ def eval_metrics() -> List[str]:
...
+            raise ValueError(
+                f"{self.rand_type} is not a valid random type. Choices are: 'const', 'choices', 'linear', 'log2'"
+            )
+        return id_str
+
+
+class RandomSearchMixin(BenchmarkSearchMixin):
+
+    @classmethod
+    def standardize_args(cls, script_args: str) -> Dict[str, List]:
```
pablomlago:
```suggestion
    def standardize_args(cls, script_args: Namespace) -> Dict[str, List]:
```

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +62,387 @@ def eval_metrics() -> List[str]:
...
+        elif self.rand_type == "exp2":
+            id_str = f"type: {self.rand_type}, min: {self.rand_values[0]}, max: {self.rand_values[1]}"
+        else:
+            raise ValueError(
+                f"{self.rand_type} is not a valid random type. Choices are: 'const', 'choices', 'linear', 'log2'"
+            )
+        return id_str
+
+
+class RandomSearchMixin(BenchmarkSearchMixin):
```
pablomlago, line 297: There is a subset of the functionality of `standardize_args` which is shared for `GridSearchMixin` and `RandomSearchMixing`, e.g. YAML reading. Is it possible to extract the common functionality to the abstract parent class.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +62,387 @@ def eval_metrics() -> List[str]:
...
+            # Remove unnecessary keys
+            del args_dict["help"]
+            del args_dict["config"]
+            # Save YAML in the results folder
+            with open(f"{script_args.results_folder}/benchmark_config.yaml", 'w') as f:
+                yaml.dump(args_dict, f)
+        return args_dict
+
+    @staticmethod
+    def parse_config_args(args: List[str]) -> Namespace:
```
pablomlago: Most of the arguments are shared between `GridSearchMixin` and `RandomSearchMixin`, I would consider extracting the common logic.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -74,6 +457,14 @@ def _print_indented_dict(message: str, dictionary: Dict) -> None:
         print(f"\t{key}: {value}")
 
 
+class GridSearchBenchmarkUtils(BenchmarkUtils, GridSearchMixin):
```
pablomlago: Can we remove these? I would rather go for having multiple inheritance in the entrypoints, e.g.
```
class ImagenetPTQBenchmarkUtils(BenchmarkUtils, GridSearchMixin):
```

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -382,56 +676,11 @@ def benchmark(entrypoint_utils: BenchmarkUtils, args: List[str]) -> None:
...
-            pass
-    if script_args.shuffle_seed is not None:
-        random.seed(script_args.shuffle_seed)
-        random.shuffle(q)
-    start_index = script_args.start_index
-    end_index = script_args.end_index if script_args.end_index > 0 else len(q)
-    q = q[start_index:end_index]
+    args_dict = entrypoint_utils.standardize_args(script_args)
+    # Generate a list of experiments
+    q = entrypoint_utils.gen_search_space(args_dict, script_args)
```
pablomlago: Maybe it is worth renaming `q` to something more self-explanatory.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Not 100% sure about this... It's already specified in `class BenchmarkUtils(ABC):`. You want it in `class BenchmarkSearchMixin(ABC):` as well? Or instead of or as well as `BenchmarkUtils`?

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Done in [d5f5ee4](https://github.com/Xilinx/brevitas/pull/1357/commits/d5f5ee44f0d60839562495c8b291394aa793a5c1), but I do wonder if we should simply make a config file a required argument when using benchmarks?

src/brevitas_examples/common/benchmark/utils.py
nickfraser: TBA

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Fixed in d2c771d.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Fixed in e1c7662.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +63,340 @@ def eval_metrics() -> List[str]:
...
+class GridSearchMixin(BenchmarkSearchMixin):
+
+    @classmethod
+    def _default_action_handler(cls, action: Action) -> List[Any]:
+        # Standardizes an argument for the given search class when it is not provided
+        return [action.default]
+
+    @staticmethod
+    def parse_config_args(args: List[str], parse_args: bool = True) -> Namespace:
+        parser = super(GridSearchMixin, GridSearchMixin).parse_config_args(args, parse_args=False)
```
pablomlago: Why does `GridSearchMixin` appear twice?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +63,340 @@ def eval_metrics() -> List[str]:
...
+class RandomSearchMixin(BenchmarkSearchMixin):
+
+    @classmethod
+    def _default_action_handler(cls, action: Action) -> Dict[str, Any]:
+        # Standardizes an argument for the given search class when it is not provided.
+        return {"rand_type": "const", "rand_values": action.default}
+
+    @staticmethod
+    def parse_config_args(args: List[str], parse_args: bool = True) -> Namespace:
+        parser = super(GridSearchMixin, GridSearchMixin).parse_config_args(args, parse_args=False)
```
pablomlago: Why is `GridSearchMixin` repeat here?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -334,46 +608,12 @@ def maybe_sort_values(values):
...
-    _print_indented_dict("Non-default args.:", args_combinations_dict)
-
-
 def benchmark(entrypoint_utils: BenchmarkUtils, args: List[str]) -> None:
     # A CUDA error message is issued when changing CUDA_VISIBLE_DEVICES
     # if processes are started in fork mode
     multiprocessing.set_start_method('spawn')
     # Parse benchmark arguments
-    script_args = parse_config_args(args)
+    script_args = entrypoint_utils.parse_config_args(args)
```
pablomlago: `entrypoint_utils` is declared as an instance of `BenchmarkUtils` but it does not implement `parse_config_args`.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -382,62 +622,17 @@ def benchmark(entrypoint_utils: BenchmarkUtils, args: List[str]) -> None:
...
-        except AssertionError:
-            # Invalid configuration
-            pass
-    if script_args.shuffle_seed is not None:
-        random.seed(script_args.shuffle_seed)
-        random.shuffle(q)
-    start_index = script_args.start_index
-    end_index = script_args.end_index if script_args.end_index > 0 else len(q)
-    q = q[start_index:end_index]
+    args_dict = entrypoint_utils.standardize_args(script_args)
```
pablomlago, line 614: Same here, should the type in the signature be `BenchmarkSearchMixin`?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +63,340 @@ def eval_metrics() -> List[str]:
...
+    def _default_action_handler(cls, action: Action) -> Any:
+        pass
+
+    @staticmethod
+    def parse_config_args(args: List[str], parse_args: bool = True) -> Namespace:
+        parser = ArgumentParser()
+        parser.add_argument(
+            '--config',
+            type=str,
+            default=None,
```
pablomlago: Should this be a required argument instead?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +63,340 @@ def eval_metrics() -> List[str]:
...
+            help=
+            'Number of retries for each argument combination in case a crash happens. Default: %(default)s.'
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            default=False,
+            help="Whether to skip running experiments (default: %(default)s).",
+        )
+        if parse_args:
```
pablomlago: I'd rather have this function always return the `ArgumentParser` (thus the type of the return value needs to be modified in the signature) and let the user call `parse_args` when required, or either separate this method in two: one that generates the `ArgumentParser`, and other that parses a list of `args`.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +63,340 @@ def eval_metrics() -> List[str]:
...
+            random.shuffle(exp_queue)
+        start_index = script_args.start_index
+        end_index = script_args.end_index if script_args.end_index > 0 else len(exp_queue)
+        exp_queue = exp_queue[start_index:end_index]
+        return exp_queue
+
+    @classmethod
+    def print_benchmark_summary(cls, args_queue: List[Dict], script_args: Namespace) -> None:
+        print(f"Num. experiments: {len(args_queue)}")
+        _print_indented_dict("Benchmark args.:", vars(script_args))
```
pablomlago, line 219: I think we should probably remove this function and use `pprint`.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Switched from `staticmethod` -> `classmethod` as we discussed

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Switched from `staticmethod` -> `classmethod` as we discussed

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Fixed.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Fixed.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Fixed.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,350 @@ def eval_metrics() -> List[str]:
...
+class SearchUtils(ABC):
+    """Encapsulates a search strategy over the entry point's argument space.
+
+    A search strategy does not know anything about a specific entry point. Any
+    data it needs from the entry point (its argument parser and its validation
+    function) is passed in explicitly by the :class:`BenchmarkUtils` coordinator.
+    """
+
+    @classmethod
+    def standardize_args(cls, script_args: Namespace,
```
nickfraser: Conceptually, this seems more related to the entry-point, rather than the search technique. Maybe this should move to `EntryPointUtils` or `BenchmarkUtils`.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -209,70 +563,7 @@ def run_args_bucket_process(
...
-        type=int,
-        default=None,
-        help=
-        'The seed to use to shuffle the jobs. If None, no shuffling will be applied. Default: %(default)s.'
-    )
-    return parser.parse_args(args)
-
-
-def parse_results(entrypoint_utils: BenchmarkUtils, results_folder: str) -> pd.DataFrame:
+def parse_results(entry_point_utils: Type[EntryPointUtils], results_folder: str) -> pd.DataFrame:
```
nickfraser: Should we move this to be a member of the `BenchmarkUtils` class? Like we did for `benchmark`. Or is this still useful as a standalone utility function?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -340,138 +631,81 @@ def maybe_sort_values(values):
...
+    the data the search needs from the entry point (its argument parser and its
+    validation function) and feeds the resulting experiment queue back into the
+    entry point for execution.
+    """
+
+    entry_point_utils: Type[EntryPointUtils] = None
+    search_utils: Type[SearchUtils] = None
+
+    @classmethod
+    def run(cls, args: List[str]) -> None:
```
nickfraser, line 749: Similar to `parse_results`, not clear if this should be a standalone function or a static method of `BenchmarkUtils`.

nickfraser (commented):
Need to reconsider a few design choices

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,350 @@ def eval_metrics() -> List[str]:
...
+                if value == default_value:
+                    del args_combinations[key]
+        args_combinations_dict = {
+            f"--{key.replace('_','-')}": maybe_sort_values(value) for key,
+            value in args_combinations.items()}
+        _print_indented_dict("Non-default args.:", args_combinations_dict)
+
+
+# A node in a tree of (possibly) dependent arguments
+class RandomArgNode:
```
nickfraser: There's lot of if/else statements in this class definition. The benefit of these to have simple instantiation logic when parsing parameters from a YAML file. Does it make sense to make a class hierarchy which provides the different random node types, and have separate logic to parse the YAML files and instantiate the write class?

nickfraser (commented):
Another design option described below

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Resolved in refactor.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Resolved in refactor.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Resolved in refactor.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Resolved in refactor.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Moved to benchmark utils.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Resolved in refactor.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,410 @@ def eval_metrics() -> List[str]:
...
+        return parser
+
+    @classmethod
+    @abstractmethod
+    def gen_search_space(
+            cls,
+            args_dict: Dict[str, Any],
+            script_args: Namespace,
+            entrypoint_parser: ArgumentParser,
+            validate_fn: Callable[[Namespace, List[str]], None]) -> List[Dict[str, Any]]:
```
pablomlago: The return value type hint is inconsistent with that of specific implementations, e.g. `List[Tuple[Dict[str, Any], List[str], Dict[str, Any]]]`.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,410 @@ def eval_metrics() -> List[str]:
...
+        pass
+
+    @classmethod
+    @abstractmethod
+    def print_benchmark_summary(
```
pablomlago, line 146: In the specific implementations, there is repeated code, e.g. 
```
print(f"Num. experiments: {len(args_queue)}")
        _print_indented_dict("Benchmark args.:", vars(script_args))
        # Return if there are not valid combination
        if len(args_queue) == 0:
            return
```
Consider moving the common code to the parent class.

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,410 @@ def eval_metrics() -> List[str]:
...
+# Each concrete node type samples a value for a single argument. Node types are
+# identified by a `rand_type` string (as it appears in the benchmark YAML) and
+# are auto-registered into `RandomArgNode._registry` via `__init_subclass__`, so
+# adding a new type only requires defining a new subclass. Instances are built
+# from a parsed YAML entry through `RandomArgNode.from_config`.
+class RandomArgNode(ABC):
+    # Maps a `rand_type` string to its concrete node class. Populated
+    # automatically whenever a subclass sets a (non-None) `rand_type`.
+    _registry: Dict[str, Type["RandomArgNode"]] = {}
+    rand_type: str = None  # Set by each concrete subclass
```
pablomlago: Subclassing from `RandomArgNode` without overriding should yield an error. Is there a syntax in Python to define a kind of abstract attribute? (the closest thing I can think of is `@abstractmethod` with `@property`).

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -60,6 +71,410 @@ def eval_metrics() -> List[str]:
...
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        super().__init_subclass__(**kwargs)
+        # Only register concrete leaf types; intermediate bases (e.g. _RangeNode)
+        # leave `rand_type` as None and are skipped.
+        if cls.rand_type is not None:
```
pablomlago: Is a `ValueError` required if this is not satisfied?

src/brevitas_examples/common/benchmark/utils.py
```diff
@@ -70,8 +485,7 @@ def _make_float(value: Any) -> Any:
 
 def _print_indented_dict(message: str, dictionary: Dict) -> None:
```
pablomlago: This function could be removed altogether, and have instead `pprint.pp(f"{message}: {dictionary}")`.

pablomlago (changes requested):
Some minor changes

src/brevitas_examples/common/benchmark/utils.py
nickfraser: No, this guard to exclude `ABC`s which inherit from `RandomArgNode`, e.g., `_RangeNode`.

Unless you mean something different (please specify!)

nickfraser: This is better handled by your previous comment (if there is a good design pattern to do this)

src/brevitas_examples/common/benchmark/utils.py
nickfraser: I think I prefer the helper function - we do want the configuration printing to be consistent across different benchmark classes. A helper function is a great way to achieve that.

An example of why a helper class is useful: I don't like the format change that would occur with your proposed changed.

current format:

```
Benchmark args.:                                                        
{'config': 'benchmark_rand_template.yaml', 'results_folder': './', ...} 
```
                                                                        
proposed change:                                                                     

```
{'Benchmark args.:': {'config': 'benchmark_rand_template.yaml',         
                      'results_folder': './', ...}}
```

In fact, I prefer the original format, but I prefer the current format over your proposed change:

```
Benchmark args.:
        config: benchmark_rand_template.yaml
        results_folder: ./
        ...
```

Lastly, I think I should change the helper name, since it no longer indents the dictionary entries.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Addressed in [dc68b3e](https://github.com/Xilinx/brevitas/pull/1357/commits/dc68b3e4c655ce5667438b712eaece5f912b900a). Created a type alias `Experiment` the cover this reused type.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Addressed in cf2056e748d9e4ff4ffdaa79c15c21c01fa28ca3. I switched this to be an abstract attribute and added some extra guards at class definition time.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Addressed in https://github.com/Xilinx/brevitas/commit/cf2056e748d9e4ff4ffdaa79c15c21c01fa28ca3. I switched this to be an abstract attribute and added some extra guards at class definition time.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: Addressed in [c20099e](https://github.com/Xilinx/brevitas/pull/1357/commits/c20099e641a51267d7769db9bac4448ef7e436ae). The concrete implementation is moved to `SearchUtils` while any variation is delegated to private helpers which are overridden by the concrete classes.

src/brevitas_examples/common/benchmark/utils.py
nickfraser: (Not) addressed in [064620e](https://github.com/Xilinx/brevitas/pull/1357/commits/064620e49a44ba3fb0dbe2dfd482eb15f31c03a1). Renamed helper to: `_print_config_block`.

PR #1368, opened 2025-08-27
ci (versions): Updated tests to test the entire suite of supported versions

nickfraser:
PR to trigger rerunning all tests before a release.


Giuseppe5:
Can this be closed?

nickfraser:
Note, since the workflow needs to be pushed to the default branch to run, I'm testing it in [my fork](https://github.com/nickfraser/brevitas).

.github/workflows/base_periodic.yml.template
```diff
@@ -0,0 +1,47 @@
...
+      matrix:
+&{matrix}
+&{exclude}
+
+    steps:
+
+    - name: Checkout repo
+      uses: actions/checkout@v3
+      with:
+        ref: dev
```
nickfraser: Remove switch to `dev`

.github/workflows/gen_github_actions.py
```diff
@@ -59,6 +66,30 @@
     'jit_disabled',])])
 
 # Data shared betwen Nox sessions and Github Actions, formatted as tuples
+ALL_SUPPORTED_PYTHON_VERSIONS = ('3.9', '3.10', '3.11', '3.12')
```
Giuseppe5, line 69: Python 3.9 has reached end of life

Giuseppe5 (approved):
Assuming the scheduling works as intended, it seems ready to go after the two changes (dev and python 3.9)

.github/workflows/gen_github_actions.py
nickfraser: Disagree, we stop supporting `python=3.9` once it becomes a burden.

PR #1380, opened 2025-09-25
Feat (graph/equalize): implement permute regions

Giuseppe5:
## Reason for this PR

Refactor the graph equalization code, and include supports for permutation in regions.

## Testing Summary

Missing tests for permutation yet.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/docsrc/source/user_guide/rotations.rst
+++ b/docsrc/source/user_guide/rotations.rst
@@ -1,0 +1,186 @@
+=================================
+Rotations in Brevitas
+=================================
+
+Why are rotations important?
+----------------------------------------------------------
+
+Large Language Models exhibit *computational invariance* [1]_ meaning that applying an invertible
+linear operation to the output of certain modules (sources), and its inverse to the input of others
+(sinks), leaves the model's output unchanged (assuming sufficient precision).
+This property allows for the selective application of random orthogonal transformations,
+which effectively mitigate weight and activation outliers, enhancing their quantization amenability [2]_.
+Moreover, some rotations can be fused into the module weights, thus preserving floating-point inference performance.
+
+Although random orthogonal rotations generally improve quantization amenability in low-bit regimes,
+the quantized network performance  exhibits a large variance under different random rotations,
+as observed in [4]_. Consequently, these authors propose to further optimize the rotations to
+improve quantized performance. In order to do so, they leverage the Cailey-SGD optimizer to ensure
+that the optimized rotations stay within the Stiefel manifold during optimization [5]_.
+
+
+Rotations in Brevitas
+----------------------------------------------------------
+
+In Brevitas, we different flavours of rotations:
+
+* *Layerwise* Rotations
+* *Fused* Rotations
+* *Fused Optimized* Rotations
+
+
+
+Layerwise Rotations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to merge rotations into subsequent and preceding layers, it is often required to apply certain
+transformations to the network, such as converting LayerNorm to RMSNorm, as well as identifying the
+suitable layers (or regions) in the network where such operation is doable.
+
+This comes with some drawbacks, such as the impact on quantization of merging RMSNorm affine parameters,
+and with some limitations.
+In order to detect the regions of the model where it is possible to fuse rotations, we rely on FX graph,
+in particular the one that can be extracted with `torch.dynamo._export`.
+
+This API allows to get a modular FX graph, which means that calls to modules (i.e., Linear module) are
+not flattened into their respective functional calls.
+
+The limitation of this API resides in the fact that graph breaks are not allowed, which means that a
+the entire network must contain only operators that can be traced through torch.dynamo.
+Although compatibility has greatly improved in the past months, custom networks and operators might still
+not be supported.
+
+
+For all these reasons, we offer the possibility of applying rotations without having to worry about
+merging the rotations in the preceding layers. From our experiments, this provides generally better
+accuracy compared to the case where rotations are fused, although the computational price to pay is
+generally higher.
+
+In its most simple form, this can be done in the following way:
+
+.. code-block:: python
+    
+    from brevitas.graph.equalize import LayerwiseActivationRotation
+
+    eq = LayerwiseActivationRotation()
+    model = eq.apply(model)
+
+Fused Rotations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Following more closely the approach used in [1]_ and [2]_, it is possible to fuse rotations into 
+preceding layers if the network structure allows it.
+
+As mentioned, there are few assumptions in this case:
+
+* The class accepts a FX-traced module (e.g., through `torch.dynamo._export`)
+* LayerNorm have been transformed to RMSNorm if necessary
+* RMSNorm affine parameters are merged into adjecent layers
+
+Brevitas offers utilities to facilitate these transformations, as well as a standalone class for
+FX-based rotation, called `GraphRotationEqualization`.
+
+The most important options for the class are:
+
+* orphan_sink: Whether to add standalone, unmerged rotations for the regions that do not support fusion
+* full_rotation_method: Allows to select between Hadamard ('had') or orthogonal ('ort') rotations
+* sdpa_regions: detect where torch.nn.functional.scaled_dot_product is placed in the network and appropriately 
+applies rotations in the layers around it. This assumes a typical attention structure around SDPA,
+with the presence of QKV linear layers before, and an Output linear layer afterwards.
+* use_parametrized_rotations: Register rotations are parametrization, allowing for a subsequent optimization,
+for example through Caley SGD.
+
+.. code-block:: python
+
+    from brevitas.graph.equalize import LayerNormToRMS
+    from brevitas.graph.equalize import MergeLnAffine
+    from brevitas.graph.equalize import GraphRotationEqualization
+    import torch
+    
+    with torch.no_grad():
+        fx_model, guards = torch._dynamo.export(model)(example_input)
+    
+        # Merge the learned (affine) LayerNorm parameters
+        eq = MergeLnAffine()
+        fx_model = eq.apply(fx_model)
+
+        # Convert LayerNorm to RMSNorm
+        eq = LayerNormToRMS()
+        fx_model = eq.apply(fx_model)
+
+        eq = GraphRotationEqualization(
+            orphan_sink=True
+            full_rotation_method='had',
+            sdpa_regions=True,
+            use_parametrized_rotations=False)
+        fx_model = eq.apply(fx_model)
+
+
+If `use_parametrized_rotations` is False, this resembles closely the optimization structure presented in 
+QuaRot [2]_.
+
+
+Optimized Fused Rotations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If `use_parametrized_rotations` is True, the rotations are registered as parametrizations ([6]_),
+and as a parameters.
+After adding the quantization blocks to the network, it is thus necessary to optimize the rotations and
+then fuse them into the corresponding weights.
+
+
+Brevitas offers an example of how to accomplish that in our LLM entrypoint (`brevitas_ptq_llm`).
+
+In particular, there are two ways to apply optimized fused rotations through this entrypoint, `FX` and `FUSED_NO_FX`.
+
+Although an FX representation of the model is fundamental for this process, it is possible to discard the FX representation
+once the correct regions have been identified, and the transformations can be then applied on the
+non-FX model.
+
+The advantage of this approach is that we do not lose the hierarchical structure of the original model,
+which is a side effect of `torch.dynamo._export`.
+Mantaining this hierarchical structure allows to optimize the execution of certain PTQ algorithms like
+GPTQ or Qronos, by optimizing one block at a time and caching intermediate representations.
+
+
+Moreover, similarly to [5]_, Brevitas can leverage the Cailey-SGD optimizer to further optimize the rotations..
+The rotation training procedure relies on the
+`HF Trainer <https://huggingface.co/docs/transformers/en/main_classes/trainer>`_ class, and, therefore,
+can be configured by passing arguments accepted by the dataclass
+`TrainingArguments <https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments>`_.
+Moreover, the number of samples used for rotation calibration can be configured through the parameter ``--nsamples-rot-calibration``.
+
+Following, we provide a minimal example configuration for optimizing, in a single GPU, the rotations
+of a ``HuggingfaceTB/SmolLM2-135M`` model, with its weights quantized to 4 bits:
+
+.. code-block:: yaml
+   dataset: wikitext2
+   eval: true
+   model: HuggingfaceTB/SmolLM2-135M
+   rotation: fused_no_fx
+   optimize_rotations: true
+   nsamples_rot_calibration: 800
+   replace_rmsnorm: true
+   weight_bit_width: 4
+   dtype: float32
+   learning_rate: 1.5
+   weight_decay: 0.0
+   lr_scheduler_type: cosine
+   max_steps: 100
+   per_device_train_batch_size: 2
+   gradient_accumulation_steps: 4
+   save_safetensors: false
+   logging_steps: 10
+   log_on_each_node: false
+
+
+Note that the training parameters used in the SpinQuant paper [5]_ can be found in their `repository <https://github.com/facebookresearch/SpinQuant>`_.
+
+.. rubric:: References
+
+.. [1] Ashkboos, S., Croci, M. L., Nascimento, M. G. D., Hoefler, T., & Hensman, J. (2024). Slicegpt: Compress large language models by deleting rows and columns. arXiv preprint arXiv:2401.15024.
+.. [2] Ashkboos, S., Mohtashami, A., Croci, M., Li, B., Cameron, P., Jaggi, M., ... & Hensman, J. (2025). Quarot: Outlier-free 4-bit inference in rotated llms. Advances in Neural Information Processing Systems, 37, 100213-100240.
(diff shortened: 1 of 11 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -583,6 +682,11 @@ def get_weight_range(
         weight = weight.cpu().to(torch.float32)
         return scale_fn(weight.reshape(weight.size(0), -1))
 
+    def permute(self, permute_index):
+        self.module.weight.data = self.module.weight.data[permute_index]
+        bias = getattr(self.module, self.bias_tensor_name)
+        bias.data = self.module.weight.data[permute_index]
```
i-colbert: should this be `bias.data = bias.data[permute_index]`?

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +712,43 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
         return scale_fn(weight.reshape(
             weight.size(0), -1))[self.equalization_indexes.start:self.equalization_indexes.end]
 
+    def permute(self, permute_index):
+        self.module.weight.data = self.module.weight.data[permute_index]
+        bias = getattr(self.module, self.bias_tensor_name)
+        bias.data = self.module.weight.data[permute_index]
```
i-colbert: should this be `bias.data = bias.data[permute_index]`?

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +237,95 @@ def is_valid(self):
...
+
+            new_srcs_dict[name] = SourceWrapper(
+                module=module,
+                weight_tensor_name="weight",
+                weight_axis=weight_axis,
+                equalization_indexes=indexes,
+                bias_tensor_name="bias",
+                bias_axis=0,
+                act_axis=act_axis)
+        act_sink_axes = {name: module.act_axis for (name, module) in new_sink_dict.items()}
```
pablomlago, line 298: I feel that the logic for determining `act_axis` is not self-explanatory, and some comments to explain these operations could be useful.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +237,95 @@ def is_valid(self):
...
+        act_sources_axes = {name: module.act_axis for (name, module) in new_srcs_dict.items()}
+        list_of_sink_axes = [x for x in list(act_sink_axes.values()) if x is not None]
```
pablomlago, line 300: I think there is no need to cast to `list`.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +237,95 @@ def is_valid(self):
...
+        return self.act_axis is not None
+
+    def apply_permute(self, indexes):
+        for src in self.srcs.values():
+            src.permute(indexes)
+        for sink in self.sinks.values():
+            sink.permute(indexes)
+
+    @staticmethod
+    def initialize_class(srcs, sinks, name_to_module):
```
pablomlago, line 251: I would expect that this factory methods returns an initialized `Region`, not the attributes to later instantiate manually a `Region`.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +237,95 @@ def is_valid(self):
...
+            srcs = dict()
+        if sinks is None:
+            sinks = dict()
+        if acts is None:
+            acts = tuple()
+        if name_to_module is None:
+            name_to_module = dict()
+
+        srcs, sinks, act_axis = cls.initialize_class(srcs, sinks, name_to_module)
+        return cls(srcs, sinks, acts, name_to_module, expand_region, act_axis)
```
pablomlago, line 327: This would not be needed if `initialize_class` returned an initialized instance of `Region`, and maybe this method could be removed altogether in favor of adding the safe-guards for `None` dictionaries to `initialize_class`.

src/brevitas/graph/equalize.py
```diff
@@ -279,6 +394,26 @@ def _select_scale_computation_fn(
...
+class rotate_permute_mode:
+
+    def __init__(self, model, **kwargs):
+        self.rotation = GraphRotationEqualization(
+            **kwargs, use_parametrized_rotations=True, delay_rewriters=True)
+        self.model = model
+        self.rewriters = None
+
+    def __enter__(self):
+        model, rewriters = self.rotation.apply(self.model)
```
pablomlago, line 406:
```suggestion
        self.model, self.rewriters = self.rotation.apply(self.model)
```
and remove the two following lines.

src/brevitas/graph/equalize.py
```diff
@@ -476,13 +611,14 @@ def transpose(tensor: torch.Tensor, axis: int):
     return tensor.permute(shape)
 
 
-class EqualizationModuleWrapper:
+class ModuleWrapper:
```
pablomlago, line 614: I understand that `EqualizationModuleWrapper` might be too verbose, but I feel `ModuleWrapper` is a too general name, and does not convey the intended use of this class.

src/brevitas/graph/equalize.py
pablomlago, line 1: This file is over 2K lines of code at the moment and it contains a lot of functionality, so maybe it is worth splitting it in a future refactor. One possibility would be to have a file with the base equalization functionality, and then other file with the code for scalar/rotation/permutation equalization, but other options could be explored.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
         return scale_fn(weight.reshape(
             weight.size(0), -1))[self.equalization_indexes.start:self.equalization_indexes.end]
 
+    def permute(self, permute_index):
+        permutation_list = []
```
pablomlago, line 764: Maybe I would opt for doing a list comprehension:
```
permutation_tuple = tuple(permute_index if i == dim else slice(size)for dim, size in enumerate(self.module.weight.shape))
```

Also, for indexing the tensor I would pass a tuple, instead of a list, to prevent the following warning:
```
<stdin>:1: UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at /pytorch/torch/csrc/autograd/python_variable_indexing.cpp:306.)
```

src/brevitas/graph/equalize.py
```diff
@@ -1102,20 +1279,122 @@ def _extract_regions(
...
                 if region not in regions and region.is_valid:
                     regions.append(region)
+
     return regions
 
 
+class PermuteGraph():
+
+    def __init__(self):
+        super().__init__()
```
pablomlago, line 1300: Is this needed? `PermuteGraph` does not seem to inherit from other class.

docsrc/source/user_guide/rotations.rst
```diff
@@ -0,0 +1,186 @@
...
+
+In particular, there are two ways to apply optimized fused rotations through this entrypoint, `FX` and `FUSED_NO_FX`.
+
+Although an FX representation of the model is fundamental for this process, it is possible to discard the FX representation
+once the correct regions have been identified, and the transformations can be then applied on the
+non-FX model.
+
+The advantage of this approach is that we do not lose the hierarchical structure of the original model,
+which is a side effect of `torch.dynamo._export`.
+Mantaining this hierarchical structure allows to optimize the execution of certain PTQ algorithms like
```
nickfraser, line 142: ~Mantaining~ -> Maintaining

Article missing:
> this hierarchical structure allows _us_ to optimize

docsrc/source/user_guide/rotations.rst
```diff
@@ -0,0 +1,186 @@
...
+GPTQ or Qronos, by optimizing one block at a time and caching intermediate representations.
+
+
+Moreover, similarly to [5]_, Brevitas can leverage the Cailey-SGD optimizer to further optimize the rotations..
```
nickfraser, line 146: .. -> .

src/brevitas/graph/equalize.py
nickfraser: > One possibility would be to have a file with the base equalization functionality, and then other file with the code for scalar/rotation/permutation equalization, but other options could be explored.

Agreed, this would be better... Do we do that in this PR or in a future refactor of this section?

src/brevitas/graph/equalize.py
nickfraser: I don't mind the name `EqualizationModuleWrapper` TBH.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+
+@torch.no_grad()
+def _permute(region, list_of_act_val):
+
+    # scale_fn = permute_op_type
+    single_module = region.get_module_from_name(next(iter(region.sinks_names)))
+    device = next(single_module.parameters()).device
+    dtype = next(single_module.parameters()).dtype
+
+    # If equalization criteria are not met, we return a scalar one to indicate that no equalization
```
nickfraser, line 808: Is this supposed to return a scalar 1?

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +2247,52 @@ def apply(self,
...
+        from brevitas_examples.common.accelerate_utils.accelerate import offload_model
+        from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
+    except:
+        offload_model = None
+        remove_hooks = None
+
+    if delay_rewriters:
+        return model
+
+    if not hasattr(model, '_hf_map'):
```
nickfraser, line 2271: Let's move this to a method in `accelerate_utils`.

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +2247,52 @@ def apply(self,
...
+        for r in rewriters:
+            model = r.apply(model)
+        return model
+
+    if offload_model is None or remove_hooks is None:
+        raise RuntimeError("Accelerate is not installed")
+    # if we use _hf_map to check and all the model is on a single GPU, then all rewriters are safe
+    if len(model._hf_map.values()) > 1:
```
nickfraser, line 2279: Let's move this to a method in `accelerate_utils`.

tests/brevitas/graph/test_equalization.py
```diff
@@ -72,11 +72,14 @@ def test_resnet18_equalization():
         equalized_layers.update(r[1])
 
     # Check that we found all the expected regions
+    print(len(regions))
```
nickfraser, line 75: Remove or use logger?

tests/brevitas/graph/test_equalization.py
```diff
@@ -72,11 +72,14 @@ def test_resnet18_equalization():
...
     for region, expected_region in zip(regions, resnet_18_regions):
         srcs = region.srcs_names
         sources_check = set(srcs) == set(expected_region[0])
         sinks = region.sinks_names
         sinks_check = set(sinks) == set(expected_region[1])
+        print(len(srcs), len(expected_region[0]))
+        print(srcs)
```
nickfraser, line 82: Remove or use logger?

nickfraser (changes requested):

src/brevitas/graph/equalize.py
```diff
@@ -164,6 +164,25 @@ class Region:
     acts: Tuple = field(default_factory=tuple)
     name_to_module: Dict = field(default_factory=dict)
     expand_region: bool = False
+    act_axis: Optional[int] = None
+
+    def __eq__(self, other):
+        if not isinstance(other, Region):
+            return False
+
+        if len(self.srcs) != len(other.srcs):
```
pablomlago, line 173: The check over `srcs` and `sinks` is the same, so the logic is repeated twice. I would consider if it is possible to prevent this repetition.

src/brevitas/graph/equalize.py
```diff
@@ -583,16 +722,30 @@ def get_weight_range(
         weight = weight.cpu().to(torch.float32)
         return scale_fn(weight.reshape(weight.size(0), -1))
 
+    def permute(self, permute_index):
```
pablomlago, line 725: This logic is repeated in `SinkWrapper` (excluding the bias permutation). I would factor out the common logic to `ModuleWrapper` (or even a standalone method) and then do the appropiate calls in Sink/SourceWrapper.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+        for i, v in enumerate(self.module.weight.shape):
+            if self.weight_axis == i:
+                permutation_list.append(permute_index)
+            else:
+                s = slice(0, v, 1)
+                permutation_list.append(s)
+        self.module.weight.data = self.module.weight.data[permutation_list]
+
+
+def new_axis(x, block_size=32):
```
pablomlago, line 774: I would add a docstring to this method, and potentially change its name.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+
+    # If equalization criteria are not met, we return a scalar one to indicate that no equalization
+    # has been performed
+    def _no_permute():
+        return
+
+    # If act_val is enabled, use source or sink weights to determine the activation channel
+    # For example, if the source is BatchNorm, we need to use the information coming from the sinks
+    if not region.is_valid_activation_equalization:
+        return _no_permute()
```
pablomlago, line 816: Note that with the current implementation of `_no_permute`, this line is returning `None`.

src/brevitas/graph/equalize.py
```diff
@@ -1102,20 +1279,122 @@ def _extract_regions(
                 sorted_sinks = dict(sorted(state.sinks.items()))
                 sorted_acts = tuple(sorted(state.acts))
                 if return_acts:
-                    region = Region(
+                    region = Region.from_dicts(
```
pablomlago, line 1282: If the factory method is refactored, I would use it here, instead of `from_dicts`.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+    idx = idx.t()
+
+    # flatten
+    idx = idx.flatten()
+
+    return idx
+
+
+@torch.no_grad()
+def _permute(region, list_of_act_val):
```
pablomlago, line 801: I feel that at the moment there are too many functions related to permutations, which makes it difficult to follow: `permute` in both `EqualizationSinkWrapper` and `EqualizationSourceWrapper`, `_permute` as a standalone method and `apply_permute` for `Region`. From my point of view, we could either move the logic of `_permute` into `apply_permute` or viceversa. Also, I would consider refactoring the common logic of `permute` (`EqualizationSink/SourceWrapper`) into the base class.

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+
+    # scale_fn = permute_op_type
+    single_module = region.get_module_from_name(next(iter(region.sinks_names)))
+    device = next(single_module.parameters()).device
```
pablomlago, line 805: Are `device` and `dtype` used in this method?

src/brevitas/graph/equalize.py
```diff
@@ -607,6 +760,78 @@ def get_weight_range(self, scale_fn: Callable) -> torch.Tensor:
...
+    # has been performed
+    def _no_permute():
+        return
+
+    # If act_val is enabled, use source or sink weights to determine the activation channel
+    # For example, if the source is BatchNorm, we need to use the information coming from the sinks
+    if not region.is_valid_activation_equalization:
+        return _no_permute()
+
+    list_of_act_val_shapes = [act_val.shape for act_val in list_of_act_val]
```
pablomlago, line 818: This piece of code does not seem self-explanatory. I would consider adding some comments to explain its functionality.

src/brevitas/graph/equalize.py
pablomlago: Also, I would add type hints to the new methods.

src/brevitas/graph/equalize.py
```diff
@@ -647,82 +872,30 @@ def _no_equalize():
...
     # In that case, do not perform graph equalization
-    axes_to_check = [m.weight_axis for m in list(src_axes.values()) + list(sink_axes.values())]
+    axes_to_check = [
+        m.weight_axis for m in list(region.srcs.values()) + list(region.sinks.values())]
     if None in axes_to_check:
         return _no_equalize()
 
     # If act_val is enabled, use source or sink weights to determine the activation channel
     # For example, if the source is BatchNorm, we need to use the information coming from the sinks
     if list_of_act_val is not None:
```
pablomlago, line 888: There is a lot of logic around `list_of_act_val` scattered around this file. I feel that, we should stop using a dictionary for `float_act_map` and define a class for it to encapsulate its functionality. Probably, this should be done in a future PR.

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +2247,52 @@ def apply(self,
...
+        graph_model = apply_rewriters(graph_model, rewriters, self.delay_rewriters)
+
         if self.return_rewriters:
             return graph_model, rewriters
         else:
             return graph_model
 
 
+@torch.no_grad()
+def apply_rewriters(
```
pablomlago, line 2259: If moving the `accelerate`-related functionality into `equalize.py` is not strictly needed to support permutations, I would leave this changes for a future PR, as we might need to discuss whether to introduce this dependency at `src/brevitas` level.

src/brevitas/graph/equalize.py
Giuseppe5: It will be done in the next release, as part of the re-org of our PTQ algorithms.
I understand it can be done without necessary breaking anything on the outside, but it will still be a half-job if we did it that way

src/brevitas/graph/equalize.py
```diff
@@ -279,6 +394,26 @@ def _select_scale_computation_fn(
...
+            **kwargs, use_parametrized_rotations=True, delay_rewriters=True)
+        self.model = model
+        self.rewriters = None
+
+    def __enter__(self):
+        model, rewriters = self.rotation.apply(self.model)
+        self.model = model
+        self.rewriters = rewriters
+        self.rotation.permute_class.setup_permute()
+
```
SpiritSeeker, line 410:
```suggestion
        return self
```
to use rewriters outside the context.

nickfraser:
Superseed by #1448.

PR #1382, opened 2025-10-01
Fix (graph/rotation): small refactor

Giuseppe5:
## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.


src/brevitas_examples/llm/main.py
```diff
@@ -110,24 +111,38 @@ def fused_rotation_no_fx(model, calibration_loader, args):
 
     for r in rewriters:
         r.apply(model)
+
     new_model = offload_model(new_model)
```
pablomlago: This is not needed if rewriters are delayed. Probably a conditional should be added to offload depending on the value of `delay_rewriters` . Also, consider having a context manager for offload/remove_hooks blocks.

src/brevitas_examples/llm/main.py
```diff
@@ -110,24 +111,38 @@ def fused_rotation_no_fx(model, calibration_loader, args):
...
         use_parametrized_rotations=args.optimize_rotations,
+        delay_rewriters=delay_rewriters,
         expansion_step=args.expansion_step,
         layers_to_expand=layers_to_expand)
     new_model, rewriters = eq.apply(new_model)
+    remove_hooks(new_model)
+
+    # if we use _hf_map to check and all the model is on a single GPU, then all rewriters are safe
     rewriters = fix_rewriter(rewriters, model, 'weight')
+    safe_rewriters = [
```
pablomlago: I feel like `safe` and `unsafe` is not a self-descriptive nomenclature, even though there is a comment explaining what it means for a rewriter to be `unsafe`. Maybe `parametrization_rewriters` and `inplace_rewriters` would be better.

src/brevitas/graph/equalize.py
```diff
@@ -1764,6 +1763,9 @@ def __init__(
         self.return_rewriters = return_rewriters
         self.sdpa_regions = sdpa_regions
         self.expansion_step = expansion_step
+        self.delay_rewriters = delay_rewriters
+        if not self.delay_rewriters:
+            assert return_rewriters, "If rewriters are not applied immediately, they must be returned and applied at a second moment"
```
pablomlago: I would make it clear for the user which flags are involved:
```suggestion
            assert return_rewriters, "If  `delay_rewriters=True`, rewriters are not applied immediately. Therefore, these must be returned, by setting `return_rewriters=True`, to be applied at a later stage."
```

src/brevitas_examples/common/accelerate_utils/accelerate.py
```diff
@@ -29,6 +29,10 @@
 logger = logging.getLogger(__name__)
 
 
+def is_main_process():
```
pablomlago: This is not strictly needed for this PR. Also, we have some distributed utilities in `torch_utils.py` (namely `init_process_group`), so I feel like we should either (1) add this function there, or (2) have a separate file with the utilities for distributed runs.

src/brevitas_examples/llm/main.py
Giuseppe5: Makes sense, I was more concerned with functionality than naming but you're right

src/brevitas/graph/equalize.py
```diff
@@ -1473,7 +1473,7 @@ def _apply_rotate(
         regions: List[Region],
         full_rotation_method='had',
         fuse_rotations: bool = True,
-        apply_inplace_rotations: bool = True,
+        delay_rewriters: bool = False,
```
pablomlago: I think this is not needed anymore in this function. I would also consider changing the name of the function, since the prefix `_apply` seems to imply that rotations are immediately applied.

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +1954,44 @@ def apply(self,
...
 
+@torch.no_grad()
+def apply_rewriters(model: torch.nn.Module, rewriters: List, delay_rewriters: bool = False):
+    from brevitas_examples.common.accelerate_utils.accelerate import offload_model
+    from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
+
+    if delay_rewriters:
+        return model
+    if not hasattr(model, '_hf_map'):
+        for r in inplace_rewriters:
```
pablomlago: `inplace_rewriters` has not been defined at this point?

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +1954,44 @@ def apply(self,
...
+        graph_model = apply_rewriters(graph_model, rewriters, self.delay_writers)
+
         if self.return_rewriters:
             return graph_model, rewriters
         else:
             return graph_model
 
 
+@torch.no_grad()
+def apply_rewriters(model: torch.nn.Module, rewriters: List, delay_rewriters: bool = False):
```
pablomlago:
```suggestion
def apply_rewriters(model: torch.nn.Module, rewriters: List[Transform], delay_rewriters: bool = False):
```

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +1954,44 @@ def apply(self,
...
+    else:
+        inplace_rewriters = rewriters
+
+    for r in inplace_rewriters:
+        model = r.apply(model)
+    model = remove_hooks(model)
+    with torch.no_grad():
+        for r in parametrization_rewriters:
+            model = r.apply(model)
+    model = offload_model(model)
```
pablomlago: I would probably avoid having the calls to `offload_model` and `remove_hooks` inside this function. Instead, I would make this function just apply the rewriters as they are given, but including a check `len(model._hf_map.values()) > 1 and any([isinstance(r, ModuleInstanceRegisterParametrization) for r in rewriters])` and raising an error/warning when this evaluates to `True`. A problem with the current implementation of this function is that it implicitly assumes that the passed model has been offloaded (as it returns an offloaded model), which does not need to be necessarily the case.

Giuseppe5: It does not assume it, it checks for it.
It checks if the model has `_hf_map`, which means the model was offloaded.

If it does have one, then it does its modifications and put back the model in the same state it was as the beginning.

If the model was not offloaded, then it simply applies the transformations and returns it.

pablomlago: I mean that it assumes it since it always returns an offloaded model, irrespective of whether the passed model had any HF hooks attached:
```
1991 model = offload_model(model)
1992 return model
```

Giuseppe5: There are other returns before that

pablomlago: I had missed the one at L1975, sorry.

docsrc/source/user_guide/rotations.rst
```diff
@@ -0,0 +1,186 @@
...
+
+In particular, there are two ways to apply optimized fused rotations through this entrypoint, `FX` and `FUSED_NO_FX`.
+
+Although an FX representation of the model is fundamental for this process, it is possible to discard the FX representation
+once the correct regions have been identified, and the transformations can be then applied on the
+non-FX model.
+
+The advantage of this approach is that we do not lose the hierarchical structure of the original model,
+which is a side effect of `torch.dynamo._export`.
+Mantaining this hierarchical structure allows to optimize the execution of certain PTQ algorithms like
```
nickfraser: Mantaining -> Maintaining

Article missing:

> this hierarchical structure allows _us_ to optimize

docsrc/source/user_guide/rotations.rst
```diff
@@ -0,0 +1,186 @@
...
+GPTQ or Qronos, by optimizing one block at a time and caching intermediate representations.
+
+
+Moreover, similarly to [5]_, Brevitas can leverage the Cailey-SGD optimizer to further optimize the rotations..
```
nickfraser: .. -> .

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +1952,52 @@ def apply(self,
...
+        from brevitas_examples.common.accelerate_utils.accelerate import offload_model
+        from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
+    except:
+        offload_model = None
+        remove_hooks = None
+
+    if delay_rewriters:
+        return model
+
+    if not hasattr(model, '_hf_map'):
```
nickfraser: Let's move this to a method in `accelerate_utils`.

src/brevitas/graph/equalize.py
```diff
@@ -1955,12 +1952,52 @@ def apply(self,
...
+        for r in rewriters:
+            model = r.apply(model)
+        return model
+
+    if offload_model is None or remove_hooks is None:
+        raise RuntimeError("Accelerate is not installed")
+    # if we use _hf_map to check and all the model is on a single GPU, then all rewriters are safe
+    if len(model._hf_map.values()) > 1:
```
nickfraser: Let's move this to a method in `accelerate_utils`.

nickfraser (changes requested):
Some small changes that need addressing, otherwise LGTM!

src/brevitas/graph/equalize.py
```diff
@@ -1955,11 +1992,33 @@ def apply(self,
...
             return graph_model, rewriters
         else:
             return graph_model
 
+    def transform_model(self, model, rewriters, delay_rewriters):
+        # In some circumstances, it might be useful to apply model transformations at a later moment
+        # The user should not be resposible for this in any case
+        if delay_rewriters:
+            return model
+        # If the model has `_hf_map`, it means accelerate was used, and we need to be careful when
```
nickfraser: Move this comment to the function definition of: `is_model_offloaded_accelerate`

src/brevitas/graph/equalize.py
```diff
@@ -279,6 +279,48 @@ def _select_scale_computation_fn(
...
+
+
+@torch.no_grad()
+def apply_rewriters_accelerate(
+        model: torch.nn.Module, rewriters: List[Transform], delay_rewriters: bool = False):
+    from brevitas_examples.common.accelerate_utils.accelerate import offload_model
+    from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
+
+    # if we use _hf_map to check and all the model is on a single GPU, then all rewriters are safe
+    if len(model._hf_map.values()) > 1:
```
nickfraser: Can you factor out this to a function too? Something like:

```python
def num_accelerators(model):
    return len(model._hf_map.values())
```

Then call this function?

nickfraser (approved):
LGTM - 2 small changes suggested then merge. No need for re-review.

PR #1390, opened 2025-10-13
Fix (ex/llm): Recursively unwrap equalized layer

nickfraser:
Fixes `KeyError` under the following conditions:
 - `rotation="layerwise"`;
 - `act_equalization="layerwise"`; and
 - `require_fx=True`

The fix recursively unwraps the last module until a non-equalized layer is found.


src/brevitas_examples/llm/main.py
```diff
@@ -194,6 +194,14 @@ def fx_required(args):
     return True if args.weight_equalization or args.act_equalization == 'fx' or args.rotation == 'fx' or args.ln_affine_merge or args.convert_layernorm_to_rmsnorm or args.quant_sdpa == 'fx' else False
 
 
+# Recursive function unwrap equalized layers
```
pablomlago:
```suggestion
# Recursive function to unwrap equalized layers
```

src/brevitas_examples/llm/main.py
```diff
@@ -194,6 +194,14 @@ def fx_required(args):
...
+def find_equalized_layer(layer):
+    if hasattr(layer, 'layer'):
+        return find_equalized_layer(layer.layer)
+    else:
```
pablomlago: No need to have the `else` statement, as the `if` returns.

pablomlago (approved):
LGTM.

PR #1391, opened 2025-10-14
Feat (core/float): better max mantissa computation

Giuseppe5:
## Reason for this PR

Small refactor to improve computation of max available mantissa given a certain mantissa bit width.
This avoids data-dependent for loop, which has two main benefits:
- Allows for gradient flow of `mantissa_bit_width` through the function
- Allows for fullgraph compile through the quantizers

The last point seems particularly relevant. From a custom training script, the following times have been observed:
- Baseline - 700 imgs/s
- Compile before this change (with graph breaks) - 1400 imgs/s
- Compile with this change (no graph breaks) - 2000 imgs/s

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.


tests/brevitas/core/test_float_quant.py
```diff
@@ -31,7 +31,7 @@ class BitwidthWrapper(torch.nn.Module):
 
     def __init__(self, value):
         super().__init__()
-        self.value = torch.tensor(value)
+        self.value = torch.tensor(float(value))
```
nickfraser, line 36: Not obvious why this change is necessary. Worth a comment?

tests/brevitas/core/test_float_quant.py
```diff
@@ -252,7 +252,7 @@ def test_inner_scale(inp, minifloat_format, scale):
 
         # scale inp manually
         scaled_inp = inp / scale
-        max_mantissa = compute_max_mantissa(torch.tensor(mantissa_bit_width))
+        max_mantissa = compute_max_mantissa(torch.tensor(float(mantissa_bit_width)))
```
nickfraser, line 257: Same here - not obvious why this change is necessary. Worth a comment?

nickfraser:
Not sure why your tests are failing - I even sanity checked that your fix makes sense on my end:

```python
import torch

def cmm1(mm):
    return torch.sum((2. ** torch.arange(0, -1. * mm - 1., -1.)))

def cmm2(mm):
    return 2 * (1 - 2 ** (-mm - 1))

for i in range(0, 20):
    mm = float(i)
    assert cmm1(mm) == cmm2(mm)
```

^Works with no error.

nickfraser (approved):
A few comments. Not sure why this change is causing failing tests - it seems pretty benign to me!

Please fix the tests though 🙏

tests/brevitas/core/test_float_quant.py
Giuseppe5: And I had to add another one.

We need to make sure bitwidth is a float, so that the max mantissa computation is a float, otherwise it gets rounded to an int.

nickfraser: Got it, but I meant a code comment ;)

Issue #1309, opened 2025-06-04
Lifting quantization parameter constraints


pablomlago:
Currently, there are a number of constraints on the quantization parameters, namely the sign of the scales (e.g. https://github.com/Xilinx/brevitas/blob/dev/src/brevitas/core/scaling/standalone.py#L274) and the sign of the zero-point (e.g. https://github.com/Xilinx/brevitas/blob/dev/src/brevitas/core/zero_point.py#L300).

Moreover, these constraints, which were imposed, to a certain extent, for compatibility with certain export flows, can lead to sub-optimal performance, as discussed for the scales in #1308. An analogous discussion should be done for the zero-point.

However, lifting these constraints might break retro-compatibility, as well as require a good amount of refactoring, given that these are also taken as implicit assumptions in certain parts of the codebase.

In this regard, a non-comprehensive list of issues, that are likely to come up when lifting these constraints, is provided below:
- Avoid duplication for signed/unsigned scales/zero-point computation.
- Handling of Po2 scales.
- Interaction of signed scales with MSE.
- Impact of signed scales in the quantization of the zero-point.

PR #1393, opened 2025-10-14
Feat (core/stats): signed scaling stats

pablomlago:
## Reason for this PR

Provides support for signed scales by tackling some of the outstanding items in [1309](https://github.com/Xilinx/brevitas/issues/1309), specifically the handling of the power of two scales and MSE scale optimization for signed scales. Closely related to [1308](https://github.com/Xilinx/brevitas/pull/1308) and [1347](https://github.com/Xilinx/brevitas/pull/1347).

## Changes Made in this PR

Minor refactoring of the MSE class, and fixes in the passing of `is_scale_unsigned` in the dependency injector.

## Testing Summary

Adds tests for `SignedAbsMax`.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/core/function_wrapper/misc.py
+++ b/src/brevitas/core/function_wrapper/misc.py
@@ -118,7 +118,7 @@ class InplaceAbs(torch.nn.Module):
     Examples:
         >>> inplace_abs = InplaceAbs()
         >>> x = torch.tensor(-1.0)
-        >>> inplace_log_two(x)
+        >>> inplace_abs(x)
         >>> x
         tensor(1.)
 
--- a/src/brevitas/core/function_wrapper/misc.py
+++ b/src/brevitas/core/function_wrapper/misc.py
@@ -109,3 +109,27 @@ class InplaceLogTwo(torch.nn.Module):
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         x.log2_()
         return x
+
+
+class InplaceAbs(torch.nn.Module):
+    """
+    Module wrapper for :func:`~torch.abs_`.
+
+    Examples:
+        >>> inplace_abs = InplaceAbs()
+        >>> x = torch.tensor(-1.0)
+        >>> inplace_log_two(x)
+        >>> x
+        tensor(1.)
+
+    Notes:
+        Inplace operations in TorchScript can be problematic, compilation is disabled.
+    """
+
+    def __init__(self) -> None:
+        super(InplaceAbs, self).__init__()
+
+    @torch.jit.ignore
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x.abs_()
+        return x
--- a/src/brevitas/core/function_wrapper/misc.py
+++ b/src/brevitas/core/function_wrapper/misc.py
@@ -29,6 +29,30 @@ class Abs(brevitas.jit.ScriptModule):
         return torch.abs(x)
 
 
+class InplaceAbs(torch.nn.Module):
+    """
+    Module wrapper for :func:`~torch.abs_`.
+
+    Examples:
+        >>> inplace_abs = InplaceAbs()
+        >>> x = torch.tensor(-1.0)
+        >>> inplace_abs(x)
+        >>> x
+        tensor(1.)
+
+    Notes:
+        Inplace operations in TorchScript can be problematic, compilation is disabled.
+    """
+
+    def __init__(self) -> None:
+        super(InplaceAbs, self).__init__()
+
+    @torch.jit.ignore
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x.abs_()
+        return x
+
+
 class Identity(brevitas.jit.ScriptModule):
     """
     Identity ScriptModule.
@@ -108,28 +132,4 @@ class InplaceLogTwo(torch.nn.Module):
     @torch.jit.ignore
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         x.log2_()
-        return x
-
-
-class InplaceAbs(torch.nn.Module):
-    """
-    Module wrapper for :func:`~torch.abs_`.
-
-    Examples:
-        >>> inplace_abs = InplaceAbs()
-        >>> x = torch.tensor(-1.0)
-        >>> inplace_abs(x)
-        >>> x
-        tensor(1.)
-
-    Notes:
-        Inplace operations in TorchScript can be problematic, compilation is disabled.
-    """
-
-    def __init__(self) -> None:
-        super(InplaceAbs, self).__init__()
-
-    @torch.jit.ignore
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
-        x.abs_()
         return x
--- a/src/brevitas/core/stats/stats_op.py
+++ b/src/brevitas/core/stats/stats_op.py
@@ -24,11 +24,7 @@ from .stats_wrapper import SCALAR_SHAPE
 
 DEFAULT_STD_DEV_EPSILON = 1e-8
 
-# When restricting scales to powers of two, signed statistics cannot be used because
-# taking the logarithm of a negative number would result a in NaN. To avoid this issue,
-# the signedness of the statistic must be tracked. This allows raising an error at
-# dependency-injection time if a signed statistic is combined with a power-of-two
-# scale restriction.
+# The signedness of the statistics needs to be tracked to be used in subsequent checks
 SIGNEDNESS_STATS = {}
 
 
--- a/src/brevitas/core/stats/stats_op.py
+++ b/src/brevitas/core/stats/stats_op.py
@@ -1,10 +1,10 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from functools import partial
 import math
 from typing import Optional
 from typing import Tuple
-from typing import Type
 
 import torch
 from torch import Tensor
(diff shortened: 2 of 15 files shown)


src/brevitas/quant/solver/common.py
```diff
@@ -71,8 +74,14 @@ def solve_bit_width_impl_from_enum(impl_type):
         raise Exception(f"{impl_type} not recognized.")
 
 
+# TODO (pml): For retrocompatibility, the scale is assumed to be unsigned when
+# impl_type == RestrictValueType.FP. In the future, FP should return
+# FloatRestrictValue, with SIGNED_FP being removed in favour of UNSIGNED_FP for
+# consistent naming.
 def solve_restrict_value_impl_from_enum(impl_type):
     if impl_type == RestrictValueType.FP:
+        return PositiveFloatRestrictValue
```
Giuseppe5: This stays `FloatRestrictValue`, and line 84 is `SignedFloatRestrictValue`

src/brevitas/quant/solver/common.py
```diff
@@ -142,28 +151,41 @@ def bit_width_impl(bit_width_impl_type):
...
-            return PercentileInterval
+            scaling_stats_impl = PercentileInterval
+        elif scaling_stats_op == StatsOp.SIGNED_MAX:
+            scaling_stats_impl = SignedAbsMax
         else:
             raise RuntimeError(f"{scaling_stats_op} not recognized.")
 
+        # For power of two scales, the stat needs to be unsigned
+        if restrict_scaling_type == RestrictValueType.POWER_OF_TWO and scaling_stats_op in [
+                StatsOp.SIGNED_MAX]:
```
Giuseppe5: the issue is that, if the user implements a new stats that is signed, this won't be triggered.

The idea of having Stats with attributes to say if they are signed or not was to "force" the users to implement new stats based on those base classes, and then we could check more easily things like this

src/brevitas/core/stats/stats_op.py
```diff
@@ -530,52 +587,29 @@ def evaluate_loss(self, x, candidate):
...
+        loss_fn = lambda candidate: self.evaluate_loss(x, candidate)
+        best_candidate, best_loss = self._mse_search(xl=base, xr=init, loss_fn=loss_fn, num_iter=self.num if self.is_scale_unsigned else self.num // 2)
+        if not self.is_scale_unsigned:
+            best_neg_candidate, best_neg_loss = self._mse_search(
+                xl=-init,
+                xr=-base,
+                loss_fn=loss_fn,
+                num_iter=self.num // 2,
+            )
+            best_candidate = torch.where(
```
Giuseppe5, line 632: In the case of `fib_search`, having it split this way is not equivalent to a single search (as if the values are all positive), or am I missing something?

src/brevitas/quant/solver/common.py
```diff
@@ -142,28 +151,41 @@ def bit_width_impl(bit_width_impl_type):
...
+            warnings.warn(
+                f"Statistic {scaling_stats_impl.__name__} is signed, which is incompatible with the restriction to "
+                f"power of twos, so its absolute value will be taken. Consider switching to an unsigned statistic."
+            )
```
Giuseppe5: Didn't we say that this should just raise an error?
The user is responsible to pick the correct stats

src/brevitas/quant/solver/common.py
```diff
@@ -142,28 +151,41 @@ def bit_width_impl(bit_width_impl_type):
...
+            scaling_stats_impl = wrap_signed_stat(scaling_stats_impl)
```
Giuseppe5: If the comment above holds, we don't need a wrapper.

In general, I'm against "silent" fails or silent change of behaviours (although the warning is there...).

Better fail fast and tell the user to try again with correct stats

Giuseppe5 (changes requested):
A few more changes required

src/brevitas/quant/solver/common.py
pablomlago: Following your suggestion, I implemented a decorator to register the signedness of the statistics.

src/brevitas/quant/solver/common.py
pablomlago: I changed it to raise a `ValueError` in this scenario.

src/brevitas/quant/solver/common.py
pablomlago: Agreed, I removed the wrapper.

src/brevitas/quant/solver/common.py
```diff
@@ -142,28 +151,44 @@ def bit_width_impl(bit_width_impl_type):
...
         else:
             raise RuntimeError(f"{scaling_stats_op} not recognized.")
 
+        # For power of two scales, the stat needs to be unsigned
+        if restrict_scaling_type == RestrictValueType.POWER_OF_TWO:
+            if scaling_stats_impl not in SIGNEDNESS_STATS:
+                raise ValueError(
+                    f"Signedness of statistic {scaling_stats_impl.__name__} is not known."
+                    f"Register the statistic using the decorator @register_stat_implementation and "
+                    f"note that only unsigned statistics can be used with power-of-two scales.")
```
Giuseppe5: We don't need this note

src/brevitas/quant/solver/common.py
```diff
@@ -289,19 +316,5 @@ def device(tracked_parameter_list):
...
-            else:
-                is_scale_negative = any(scaling_init.flatten() < 0)
-            return not is_scale_negative
-        elif hasattr(scaling_stats_impl, 'is_scale_unsigned'):
-            return scaling_stats_impl.is_scale_unsigned
-        else:
-            # If it is not possible to infer the scale of the sign, we assume it is unsigned
-            return True
+    def is_scale_unsigned(restrict_scaling_type=None):
+        return restrict_scaling_type in [RestrictValueType.FP, RestrictValueType.POWER_OF_TWO]
```
Giuseppe5, line 319: This is resolved by looking at the register dict

src/brevitas/quant/solver/common.py
```diff
@@ -142,28 +151,44 @@ def bit_width_impl(bit_width_impl_type):
...
         elif scaling_stats_op == StatsOp.PERCENTILE_INTERVAL:
-            return PercentileInterval
+            scaling_stats_impl = PercentileInterval
+        elif scaling_stats_op == StatsOp.SIGNED_MAX:
+            scaling_stats_impl = SignedAbsMax
         else:
             raise RuntimeError(f"{scaling_stats_op} not recognized.")
 
+        # For power of two scales, the stat needs to be unsigned
+        if restrict_scaling_type == RestrictValueType.POWER_OF_TWO:
```
Giuseppe5, line 179: We could put these cheks at the beginning, and leave the return structure as it was before?

src/brevitas/quant/solver/common.py
pablomlago: Could you clarify?

src/brevitas/quant/solver/common.py
pablomlago: I think we need it to enforce the user to register its statistic.

src/brevitas/quant/solver/common.py
Giuseppe5: We need to register statistics with a decorator and put them in a dict that tells us if they are signed or unsigned.
We should use that dict for this check.
If the stats is not registered, this fails

src/brevitas/quant/solver/common.py
Giuseppe5: I meant the second part of the Error, about the power-of-two scales, it has nothing to do with the first part

src/brevitas/quant/solver/common.py
pablomlago: The problem is that `SIGNEDNESS_STATS` tracks the classes, so `scaling_stats_impl` needs to be resolved beforehand. Otherwise, I would need to populate `SIGNEDNESS_STATS` with the values of the enum `StatsOp`.

Giuseppe5: Then let's leave it like this

src/brevitas/core/stats/stats_op.py
```diff
@@ -24,21 +24,28 @@
 
 DEFAULT_STD_DEV_EPSILON = 1e-8
 
+# When restricting scales to powers of two, signed statistics cannot be used because
```
Giuseppe5: I understand this comment but it is too specific.
The point of this is to simply keep track of which scales have a sign or not, all the logarithmic check part is someone else responsibility so I would avoid making this comment too long/complex

src/brevitas/core/function_wrapper/misc.py
```diff
@@ -109,3 +109,27 @@ def __init__(self) -> None:
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         x.log2_()
         return x
+
+
+class InplaceAbs(torch.nn.Module):
```
Giuseppe5: Move this near Abs

Giuseppe5 (approved):
Two quick comments then we can merge

src/brevitas/core/stats/stats_op.py
pablomlago: From my perspective, an user should be only aware of the existence `register_stats_implementation`, while `SIGNEDNESS_STATS` is an internal structure, that a developer should be aware of. Therefore, the comment is targeted for a developer to understand the need of this structure, and its usage, so I think it might be better to be a bit overly verbose.

PR #1395, opened 2025-10-16
Feat (export/onnx): fallback export to fake quantized weights

Giuseppe5:
## Reason for this PR

ONNX's QuantizeLinear node only supports ROUND as rounding function, but there are cases where we might be using FLOOR.

## Changes Made in this PR

This PR addresses this issue by falling back to saving and exporting fake quantized weights.
With this trick, we use whatever interal rounding operation to pre-quantize the weights, so that round is basically a no-op during export time.

The drawback of this approach is that we effectively duplicate weights during export.

This could be remedied for example by destructively replacing the original weights with the fake quantized ones, with the obvious drawback that we lose access to the original model's weights.

## Testing Summary

Added tests with floor rounding format for weights.
This effectively double the testing configurations and the testing time.

### DIFF
--- a/tests/brevitas_ort/test_quant_module.py
+++ b/tests/brevitas_ort/test_quant_module.py
@@ -25,6 +25,7 @@ from .quant_module_cases import QuantWBIOLCases
 def test_ort_wbiol(model, export_type, current_cases):
     cases_generator_func = current_cases['model'][1]
     case_id = get_case_id(cases_generator_func)
+    rounding = case_id.split('-')[0]
     impl = case_id.split('-')[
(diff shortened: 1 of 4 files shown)


tests/brevitas_ort/test_quant_module.py
```diff
@@ -19,19 +20,32 @@
 from .quant_module_cases import QuantWBIOLCases
 
 
+def rm_onnx(path):
```
nickfraser: Maybe load up the fixture from `tests/export_fixture.py`?

nickfraser (approved):
One comment about a duplicated function, otherwise LGTM!

nickfraser:
FYI, the FINN integration issues are likely fixed in #1400

Issue #1397, opened 2025-10-21
ONNX QCDQ Export Fails for Pytorch 2.9.0+cpu and dynamo=False


JP-Amboage:
Trying to export a Quant model with `export_onnx_qcdq`  and  `dynamo=False` results in the following error when using Pytorch 2.9.0+cpu :
```
ModuleNotFoundError: No module named 'torch.onnx._globals'
```
I provide the file needed to create the environment to reproduce the issue [issue-env.yaml](https://github.com/user-attachments/files/23020894/issue-env.yaml) and a jupyter notebook [issue.ipynb](https://github.com/user-attachments/files/23020904/issue.ipynb) with the code that causes the issue and the error message visible in the output cells.

Finally here is a simple code snippet where the problem happens:
```python
import torch
import brevitas.nn as qnn
from brevitas.export import export_onnx_qcdq

IN_CH = 3
OUT_CH = 128
BATCH_SIZE = 1

linear = qnn.QuantLinear(IN_CH, OUT_CH, bias=True)
inp = torch.randn(BATCH_SIZE, IN_CH)
path = 'quant_linear_qcdq.onnx'

exported_model = export_onnx_qcdq(linear, args=inp, export_path=path, dynamo=False)

```

Note that this is a different Error from the one in [this other issue](https://github.com/Xilinx/brevitas/issues/1396).

nickfraser:
Looks like the location of this file [has moved](https://github.com/pytorch/pytorch/commit/524b78d4f67045b83bb69edc56ab16efe282971c), from `torch.onnx._globals` -> `torch.onnx._internal.torchscript_importer._globals` in `pytorch==2.9`.

nickfraser:
Fixed in #1398

PR #1398, opened 2025-10-21
Fix (export/onnx): import GLOBALS from correct location depending on torch version

JP-Amboage:
## Reason for this PR

Exporting with `export_onnx_qcdq` was not working for PyTorch 2.9.0 even with `dynamo=False` as described in [this issue](https://github.com/Xilinx/brevitas/issues/1397). The reason for this is a change in location from `torch.onnx._globals` to `torch.onnx._internal.torchscript_importer._globals` in  PyTorch 2.9.0 (see [this pytorch commit](https://github.com/pytorch/pytorch/commit/524b78d4f67045b83bb69edc56ab16efe282971c)).

## Changes Made in this PR
Edited the file `brevitas/export/onnx/__init__.py` to import `GLOBALS` from the right location depending on the torch version. An `if`/`else` statement that checks the torch version was added in the import of `GLOBALS` to make the correct import according to the torch version. This approach was chosen instead of using `try`/`except` to make it easier to search for code related to specific torch versions in future changes.

## Testing Summary
- Tests run locally

### DIFF
--- a/src/brevitas/export/onnx/__init__.py
+++ b/src/brevitas/export/onnx/__init__.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
-
+from brevitas import torch_version
+from packaging import version
 
 def onnx_export_opset():
     try:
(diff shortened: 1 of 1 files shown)


nickfraser:
Thanks for the contribution - although PyTorch=2.9 is not currently being tested in our CI. I can reproduce #1397 and confirm that this fixes it. Can you please run our pre-commit script to fix the style errors? You can run it as follows in the root directory of the cloned repo:

```bash
pre-commit run -a
```

nickfraser (approved):
Please fix the style, otherwise LGTM!

PR #1404, opened 2025-10-28
Fix (core/scaling): fix dtype for threshold

Giuseppe5:
## Reason for this PR

In some circumstances, we see that the result of:
```python
2 ** bit_width
```
where bit_width is a `torch.float16` number might return a `torch.float32` value, causing typing mismatches.

We already have a similar mechanism for `float_scaling.py`, so this also helps with consistency.

## Changes Made in this PR

Explicit dtype cast

## Testing Summary

Unsure still how to reproduce the behaviour, but it happens in the wild.


nickfraser (approved):
LGTM!

PR #1407, opened 2025-11-03
Feat (loss/bit_width): removable hooks

Giuseppe5:
## Reason for this PR

When using any derivative of `BitWidthWeighted`, hooks are attached to the model but there is no way of removing them.

## Changes Made in this PR

Keep track of the hooks and expose a new method `remove_hooks` that iterate and removes all of them once the training is complete.

### DIFF
--- a/src/brevitas/loss/weighted_bit_width.py
+++ b/src/brevitas/loss/weighted_bit_width.py
@@ -25,6 +25,7 @@ class BitWidthWeighted(object):
         self.model: nn.Module = model
         self.weighted_bit_width_list: List[torch.Tensor] = []
         self.tot_num_elements: int = 0
+        self.list_of_hooks = []
         self.register_hooks()
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


src/brevitas/loss/weighted_bit_width.py
```diff
@@ -99,12 +106,13 @@ def register_hooks(self):
         def hook_fn(module, input, output: QuantTensor):
             num_elements = reduce(mul, output.value.size(), 1)
             self.weighted_bit_width_list.append(
-                num_elements * (output.mantissa_bit_width + output.exponent_bit_width))
+                num_elements * (output.mantissa_bit_width + output.exponent_bit_width + 1))
```
Giuseppe5: This is to account for the (always present) bit sign

nickfraser: Let's grab this from `output.signed`.

src/brevitas/loss/weighted_bit_width.py
```diff
@@ -118,12 +126,13 @@ def register_hooks(self):
         def hook_fn(module, input, output: QuantTensor):
             num_elements = reduce(mul, output.value.size()[1:], 1)  # exclude batch size
             self.weighted_bit_width_list.append(
-                num_elements * (output.mantissa_bit_width + output.exponent_bit_width))
+                num_elements * (output.mantissa_bit_width + output.exponent_bit_width + 1))
```
nickfraser: Same, add `output.signed`.

nickfraser (approved):
Two small comments, then this is ready to go.

PR #1410, opened 2025-11-06
Fix (brevitas.core.zero_point.pre_scaling): Fixed a circular import preventing me from running an import

JP-Amboage:
## Reason for this PR

I was unable to import `ZeroZeroPoint` from `brevitas.core.zero_point` due to the following error:
```
>>> from brevitas.core.zero_point import ZeroZeroPoint
  File "<stdin>", line 1, in <module>
  File "/home/jgarciaa/.local/miniforge3/envs/brv/lib/python3.12/site-packages/brevitas/core/zero_point.py", line 17, in <module>
    from brevitas.core.scaling.runtime import _AffineRescaling
  File "/home/jgarciaa/.local/miniforge3/envs/brv/lib/python3.12/site-packages/brevitas/core/scaling/__init__.py", line 15, in <module>
    from .pre_scaling import AccumulatorAwareParameterPreScaling
  File "/home/jgarciaa/.local/miniforge3/envs/brv/lib/python3.12/site-packages/brevitas/core/scaling/pre_scaling.py", line 18, in <module>
    from brevitas.core.zero_point import PreZeroCenterZeroPoint
ImportError: cannot import name 'PreZeroCenterZeroPoint' from partially initialized module 'brevitas.core.zero_point' (most likely due to a circular import) (/home/jgarciaa/.local/miniforge3/envs/brv/lib/python3.12/site-packages/brevitas/core/zero_point.py)
```
The problem comes from a circular import as `brevitas.core.zero_point.pre_scaling` imports `PreZeroCenterZeroPoint` from `brevitas.core.zero_point` 

## Changes Made in this PR

I removed the import of `PreZeroCenterZeroPoint` in `brevitas.core.zero_point.pre_scaling` and added an auxiliar function that runs the import and returns the class. Then, in the places (it was actually only one), where  `PreZeroCenterZeroPoint` was used in  `brevitas.core.zero_point.pre_scaling` I replaced `PreZeroCenterZeroPoint` by a call to the auxiliar funciton. In this way the import gets resolved at runtime instead of during the module loading fixing the issue of the circular import.

## Testing Summary
Tested locally

### DIFF
--- a/src/brevitas/core/scaling/pre_scaling.py
+++ b/src/brevitas/core/scaling/pre_scaling.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from typing import Any
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -15,13 +16,18 @@ import brevitas.config as config
 from brevitas.core.restrict_val import _RestrictClampValue
 from brevitas.core.stats import SCALAR_SHAPE
 from brevitas.core.stats.stats_wrapper import _Stats
-from brevitas.core.zero_point import PreZeroCenterZeroPoint
 from brevitas.function import get_upper_bound_on_l1_norm
 
 __all__ = [
     "ParameterPreScalingWeightNorm",
     "AccumulatorAwareParameterPreScaling",
     "AccumulatorAwareZeroCenterParameterPreScaling"]
+
+
+# Avoid ciruclar imports
+def get_PreZeroCenterZeroPoint_class() -> Any:
+    from brevitas.core.zero_point import PreZeroCenterZeroPoint
+    return PreZeroCenterZeroPoint
 
 
 class ParameterPreScalingWeightNorm(brevitas.jit.ScriptModule):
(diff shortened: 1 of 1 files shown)


nickfraser (approved):
I'm not sure that this is the best approach, but I'm wary of over-engineering this. @Giuseppe5, @pablomlago, any concerns? Otherwise I'm inclined to merge this...

src/brevitas/core/scaling/pre_scaling.py
```diff
@@ -24,6 +24,12 @@
     "AccumulatorAwareZeroCenterParameterPreScaling"]
 
 
+# Avoid ciruclar imports
```
pablomlago, line 27:
```suggestion
# Avoid circular imports
```

src/brevitas/core/scaling/pre_scaling.py
```diff
@@ -248,7 +254,7 @@ def __init__(
             restrict_pre_scaling_impl,
             pre_scaling_min_val)
         assert isinstance(
-            pre_zero_point_impl, PreZeroCenterZeroPoint
+            pre_zero_point_impl, get_PreZeroCenterZeroPoint_class()
```
Giuseppe5, line 257: Another solution is to check
```suggestion
            pre_zero_point_impl.__class__.__name__, 'PreZeroCenterZeroPoint'
```

I kind of hate the function to import something. I guess this is slightly hard-coded but the check seems fairly hard-coded itself anyway

pablomlago:
@nickfraser @Giuseppe5 Should we consider refactoring the code to prevent this circular import? For instance, if we remove the `pre_scaling` imports in `brevitas/src/brevitas/core/scaling/__init__.py`, there would no longer be this circular import. If there is not a strong reason for having these classes in the `scaling` namespace, I would consider this option.

Giuseppe5:
pre_scaling are part of scaling, so I don't see why we should remove the imports from `__init__.py`

nickfraser:
I agree, I think we should keep these in the scaling namespace.

nickfraser:
Superseeded by #1422.

PR #1418, opened 2025-11-19
Feat (core): module for runtime computation of exp bias

Giuseppe5:
## Reason for this PR

Exponent bias is current its own class but we don't have any utility to compute it at runtime based on the current value of exponent bit width

### DIFF
--- a/src/brevitas/core/bit_width/float.py
+++ b/src/brevitas/core/bit_width/float.py
@@ -91,6 +91,17 @@ class StaticExponentBias(torch.nn.Module):
 
 
 class ComputeExponentBias(torch.nn.Module):
+    """
+    Module that returns a runtime-computed exponent bias value.
+
+    Args:
+        exponent_bit_width_impl: Module that returns the exponent bit width
+
+    Examples:
+        >>> exp_bias = ComputeExponentBias(4.)
+        >>> exp_bias()
+        tensor(7.)
+    """
 
     def __init__(self, exponent_bit_width_impl):
         super().__init__()
--- a/src/brevitas/core/bit_width/float.py
+++ b/src/brevitas/core/bit_width/float.py
@@ -88,3 +88,13 @@ class StaticExponentBias(torch.nn.Module):
 
     def forward(self):
         return self.exponent_bias()
+
+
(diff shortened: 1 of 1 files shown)


src/brevitas/core/bit_width/float.py
```diff
@@ -88,3 +88,24 @@ def __init__(self, exponent_bias, device=None, dtype=None):
...
+        >>> exp_bias()
+        tensor(7.)
+    """
+
+    def __init__(self, exponent_bit_width_impl):
+        super().__init__()
+        self.exponent_bit_width_impl = exponent_bit_width_impl
+
+    def forward(self):
+        return 2 ** (self.exponent_bit_width_impl() - 1) - 1
```
nickfraser, line 112: Does that mean this can be deleted: https://github.com/Xilinx/brevitas/blob/dev/src/brevitas/quant/experimental/float_base.py#L31-L33

src/brevitas/core/bit_width/float.py
```diff
@@ -88,3 +88,24 @@ def __init__(self, exponent_bias, device=None, dtype=None):
...
+    Args:
+        exponent_bit_width_impl: Module that returns the exponent bit width
+
+    Examples:
+        >>> exp_bias = ComputeExponentBias(4.)
+        >>> exp_bias()
+        tensor(7.)
+    """
+
+    def __init__(self, exponent_bit_width_impl):
```
nickfraser: Please add type annotations to here and `forward`

nickfraser (approved):
LGTM, pending the minor changes requested and fixing the tests.

src/brevitas/core/bit_width/float.py
nickfraser: I.e., should runtime exponent bias calculation using this module be the default?

Giuseppe5: No the default now is still https://github.com/Xilinx/brevitas/blob/c10ef8764967e9cacc60347ce185be14e4ad97c4/src/brevitas/quant/experimental/float_base.py#L29

Which requires the value function above.

I'd leave it like this for now

PR #1419, opened 2025-11-20
Fix (proxy): preserve training state after tensor_quant re-init

Giuseppe5:
## Reason for this PR

When merging rotations, the state and device of the quantizers might be lost.

## Changes Made in this PR

Preserve and restore device and training state.

## Testing Summary

TBD

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1679,15 +1679,11 @@ def fuse_parametrizations(model: nn.Module) -> nn.Module:
                 if parametrize.is_parametrized(module) and tensor_name in module.parametrizations:
                     # Check if the module has any quantization-related children
                     state_dict = None
-                    training_state = None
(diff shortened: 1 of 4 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -1679,11 +1679,15 @@ def fuse_parametrizations(model: nn.Module) -> nn.Module:
                 if parametrize.is_parametrized(module) and tensor_name in module.parametrizations:
                     # Check if the module has any quantization-related children
                     state_dict = None
+                    training_state = None
```
pablomlago: Should this be handled within the proxies?

pablomlago (commented):
Is it possible to have some test to verify that the fix works as intended?

pablomlago (approved):

pablomlago:
LGTM (assuming tests pass)

PR #1427, opened 2025-12-02
Feat (brevitas_examples/llm): Support for batched inputs in GPXQ/Qronos forward passes.

JP-Amboage:
## Reason for this PR

Currently in `brevitas_examples/llm` when GPTQ, GPFQ or Qronos are applied there is a need to run separate LLM forward passes for each of the samples in the calibration data. By performing these forward passes per batch instead of per sample it is possible to save considerable time when running these algorithms. 

For example, for Llama-3.2-1B applying GPTQ with the current setup and a calibration size of 512 takes around 1h whereas using the same configuration but processing the calibration data in batches of 128 samples reduces the runtime of the algorithm to ~0.5h (similar or greater speedups were observed for Qronos and GPFQ).

In private discussions during the PR review it was decided to extend the use of batched inputs to the other algorithms in `brevitas_examples/llm.main/py` which could directly support it.

## Changes Made in this PR
-  Added `--calibration-batchsize` as a new argument for LLM experiments allowing to choose the batch size for the forward pass of the LLM in GPXQ/Qronos algorithms as well as in the other algorithms supporting batched inputs. The default value is 1 to mimic the existing setup.
- <s>Added a function in `llm_quant/data_utils.py` to build a DataLoader from a DatasetToDevice which is the class currently used to handle the calibration data. </s>. Deleted after PR review this is now done directly in `main.py`.
- Added a custom collate function in `llm_quant/data_utils.py` that is needed to create the DataLoader.
- In `main.py` added code to instantiate and use the DataLoader, the data loader is stored in the variable `calibration_loader` and the dataset (previously named `calibration_loader`) is now named `calibration_dataset`. Similarly, the variable `validation_dataloader` (which was actually storing a dataset) has been renamed to `validation_dataset`.
- Type hints were rectified in some of the methods that required a dataset but were typed as DataLoader.
- In `src/brevitas_examples/llm/llm_quant/rotation_optimization.py` removed the `collate_fn` to avoid having duplicated code and added instead a new function named `data_collator` accommodating to Hugging Face's interface that internally relies in the collate function that this PR adds in `llm_quant/data_utils.py`.
- Added a unit test to `tests/brevitas_examples/test_llm_data.py` to ensure the DataLoader is built correctly.
> [!WARNING]
>  **Had to edit the quantized perplexity values in some test cases** as for Pytorch inner reasons the random state changes after calling `iter(calibration_loader)` (even when shuffle is False, and everything is set to be deterministic in the data loader). The change in the random state affected some algorithms, as for example random matrices may be used.

## Testing Summary
- Tests in `tests/brevitas_examples/test_llm_data.py` including the new test for the DataLoader were run locally.
- Python debugger was used to inspect that the creation of the Hessian matrix was correct in the batched version (i.e. that it matched the values in the original code). Small discrepancies were found, when using a deterministic environment and setting the precision in the computation of the Hessian to be float64 no differences were observed. 
- The additional sanity checks performed following the PR review provided the following results for `HuggingFaceTB/SmolLM-135M`  with weight only quantization to int4 for `wikitext2` dataset:

| Method | Batch Size | Run Time (s) | Quantized Perplexity |
| -------- | ------- | ------- | ------- |
| GPTQ | 1   | 7249 | 19.125 |
| GPTQ | 64 | 1059 | 19.125 |
| GPFQ | 1 | 29650 | 19.500 |
| GPFQ | 64 |1880 | 19.500 |
| Qronos | 1 | 30073 |18.875 |
| Qronos | 64 | 1957 | 18.875 |


tests/brevitas_examples/test_llm_data.py
```diff
@@ -175,3 +177,35 @@ def test_wikitext2_tokenization(add_bos_token: bool, split: str):
...
+            {
+                'input_ids': torch.tensor([[5, 6, 7, 8]], dtype=torch.int64),
+                'attention_mask': torch.tensor([[1, 0, 1, 0]], dtype=torch.int64),}]
+    dataset2device = DatasetToDevice(data, device='cpu')
+
+    # create dataloader with batch size 1 (default)
+    data_loader = get_dataloader_from_dataset(dataset2device)
+    assert len(data_loader) == 2, 'data loader has length != num_samples/batch_size'
+    idx = 0
+    for batch in data_loader:
```
pablomlago:
```suggestion
    for idx, batch in enumerate(data_loader):
```

src/brevitas_examples/llm/llm_args.py
```diff
@@ -470,6 +470,13 @@ def create_args_parser() -> ArgumentParser:
         "--awq-clip",
         action="store_true",
         help="Whether to apply AWQ clipping (default: %(default)s).")
+
+    parser.add_argument(
+        '--gpxq-batchsize',
```
pablomlago:
```suggestion
        '--gpxq-batch-size',
```

src/brevitas_examples/llm/llm_args.py
```diff
@@ -470,6 +470,13 @@ def create_args_parser() -> ArgumentParser:
...
+        type=int,
+        default=1,
+        help='Batchsize of dataloader for gpxq/qronos forward passes (default: %(default)s).')
```
pablomlago:
```suggestion
        help='Batch size of dataloader for gpxq/qronos forward passes (default: %(default)s).')
```

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -165,3 +168,29 @@ def get_dataset_for_model(
...
+            else:
+                if key not in kwargs:
+                    kwargs[key] = value
+    for key, value in kwargs.items():
+        if isinstance(value, list) and len(value) > 0:
+            kwargs[key] = torch.cat(kwargs[key], dim=0)
+    return kwargs
+
+
+def get_dataloader_from_dataset(
```
pablomlago: I'm not sure if it is worth having a function for this one-liner. Maybe it is worth being a bit verbose and make the collate function explicit when instantiating the data loader.

src/brevitas_examples/llm/main.py
```diff
@@ -240,6 +241,11 @@ def quantize_llm(args, extra_args=None):
         require_fx=require_fx and args.export_target is not None,
         device=None)
 
+    # Batched data loader to accelerate GPXQ algorithms
+    if args.gptq or args.gpfq or args.qronos:
+        calibration_batched_loader = get_dataloader_from_dataset(
```
pablomlago: Maybe it is worth making it explicit which algorithms leverage this calibration loader at the moment?

```suggestion
        gpxq_calibration_loader = get_dataloader_from_dataset(
```

src/brevitas_examples/llm/llm_quant/data_utils.py
JP-Amboage: What do you mean exactly by making the collate function explicit when instantiating the data loader? Removing the collate function in lines 173-187 and defining it inside the function that instantiates the data loader?

pablomlago: At the moment, this function is invoked in the entrypoint as `calibration_batched_loader = get_dataloader_from_dataset(calibration_loader, batch_size=args.gpxq_batchsize)`, so the only thing that you get from having this `get_dataloader_from_dataset` wrapper is that you do not need to set the `collate_fn` explicitly, as you have a default value for it. Therefore, I mean having `calibration_batched_loader = DataLoader(calibration_loader, collate_fn=collate_fn, batch_size=args.gpxq_batchsize)` so it is explicit in the entrypoint which collate function is being used.

JP-Amboage: Committed this change. I'm marking the conversation as resolved.

src/brevitas_examples/llm/main.py
JP-Amboage: Committed this change. I'm marking the conversation as resolved.

tests/brevitas_examples/test_llm_data.py
```diff
@@ -175,3 +177,36 @@ def test_wikitext2_tokenization(add_bos_token: bool, split: str):
...
+        'attention_mask': torch.tensor([[1, 1, 1, 1]], dtype=torch.int64),},
+            {
+                'input_ids': torch.tensor([[5, 6, 7, 8]], dtype=torch.int64),
+                'attention_mask': torch.tensor([[1, 0, 1, 0]], dtype=torch.int64),}]
+    dataset2device = DatasetToDevice(data, device='cpu')
+
+    # create dataloader with batch size 1 (default)
+    data_loader = torch.utils.data.DataLoader(dataset=dataset2device, collate_fn=collate_fn)
+    assert len(data_loader) == 2, 'data loader has length != num_samples/batch_size'
+    idx = 0
```
pablomlago: This should be removed.

tests/brevitas_examples/test_llm_data.py
```diff
@@ -175,3 +177,36 @@ def test_wikitext2_tokenization(add_bos_token: bool, split: str):
...
+    for idx, batch in enumerate(data_loader):
+        assert torch.allclose(batch['input_ids'], data[idx]['input_ids']), 'input_ids mismatch'
+        assert torch.allclose(batch['attention_mask'], data[idx]['attention_mask']), 'attention_mask mismatch'
+        assert set(batch.keys()) == set(['input_ids', 'attention_mask']), 'unexpected keys in dataloader'
+        idx += 1
```
pablomlago: This should be removed.

pablomlago (approved):
LGTM, I'll wait for @Giuseppe5 to review it before merging.

Giuseppe5:
Update the description above to match the new version.
Also, since we had to change the results of the tests, would you mind running some sanity checks, comparing a few (max 2/3 configurations) with batch size = 1 and batch size > 1?
I would run all the quantization configurations that benefit from higher batch size.

You could paste the results in the PR description above

PR #1429, opened 2025-12-09
Fix (brevitas_examples/llm): better compile with merged rotations

Giuseppe5:
## Reason for this PR

After merging the rotations, the compiled module is destroyed

## Changes Made in this PR

If we detect that there is a compiled module before rotations, we re-compile after merging the rotations.

## Testing Summary

TBD


nickfraser (approved):
Add property with default value in base proxy, then merge!

PR #1379, opened 2025-09-24
Fix (ex/llm): Fix integration with Lighteval Python API

pablomlago:
## Reason for this PR

Refactored integration with `lighteval` Python API to not avoid relying on the custom [fork](https://github.com/Giuseppe5/lighteval) (see https://github.com/huggingface/lighteval/issues/489 for extra context).

## Changes Made in this PR

Refactored the `lighteval` evaluation code out of the LLM entrypoint, using a similar structure to https://github.com/huggingface/lighteval/blob/main/src/lighteval/main_accelerate.py.

## Testing Summary

Added a new end-to-end test in `test_llm` (`test_few_shot_eval`) to evaluate a quantized model in `lighteval`.

## Risk Highlight

- [x] This PR includes code from another work (please detail).
- [ ] This PR contains API-breaking changes.
- [ ] This PR depends on work in another PR (please provide links/details).
- [x] This PR introduces new dependencies (please detail).
- [ ] There are coverage gaps not covered by tests.
- [ ] Documentation updates required in subsequent PR.

## Checklist

- [x] Code comments added to any hard-to-understand areas, if applicable.
- [x] Changes generate no new warnings.
- [x] Updated any relevant tests, if applicable.
- [x] No conflicts with destination `dev` branch.
- [x] I reviewed my own code changes.
- [x] Initial CI/CD passing.
- [x] 1+ reviews given, and any review issues addressed and approved.
- [x] Post-review full CI/CD passing.

PR #1430, opened 2025-12-15
Fix (brevitas_examples/llm): correct batch size for lm_eval

Giuseppe5:
## Reason for this PR

lm_eval test was incredibly slow, mostly due to the fact that batch size was not handled properly

## Changes Made in this PR

Now we pass the batch size to the model, which speeds up the computation considerably.

Passing the batch size to the evaluator does not produce any effect, mostly because we pass a torch.nn.Module to the evaluator, which means that a lot of the evaluator args/kwargs are ignored.

## Testing Summary

N/A as lm_eval is an option al dependency.
Do we want any?
@nickfraser @pablomlago

### DIFF
--- a/src/brevitas_examples/llm/main.py
+++ b/src/brevitas_examples/llm/main.py
@@ -654,11 +654,11 @@ def quantize_llm(args, extra_args=None):
             from lm_eval.models.huggingface import HFLM
             with torch.no_grad(), quant_inference_mode(model, compile=args.compile_eval):
                 model(**calibration_loader[0])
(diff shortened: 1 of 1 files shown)


pablomlago (approved):
LGTM. Consider adding a test similar to the one that we have for `lighteval` or open an issue otherwise to do it in the future. Also, it might be worth considering a future refactoring of this evaluation pipeline, similar to the one done in [1379](https://github.com/Xilinx/brevitas/pull/1379)

Giuseppe5:
I am merging this and opening another PR for tests

PR #1431, opened 2025-12-15
Test (brevitas_examples/llm): new lm_eval tests

Giuseppe5:
## Changes Made in this PR
Test for LM Eval


pablomlago:
The structure of the `ligtheval` and `lm_eval` tests is the same. Is it possible to unify them within a single test with multiple cases? This should be allowed by the safeguard:
```
for lib in request.param["imports"]:
    pytest.importorskip(lib, reason=f"`{lib}` needs to be installed.")
```
Also, do we need to have different nox environments? Can't we just have an enviroment with `lm_eval` and `lighteval`, and test both few shot evaluations with that single environment?

pablomlago (approved):
LGTM

PR #1432, opened 2025-12-15
Feat (graph/equalize): minor refactor

Giuseppe5:
## Reason for this PR

Clean up and small refactor of equalize

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -164,6 +164,23 @@ class Region:
     acts: Tuple = field(default_factory=tuple)
     name_to_module: Dict = field(default_factory=dict)
     expand_region: bool = False
+    act_axis: Optional[int] = None
+
+    def __eq__(self, other):
+        if not isinstance(other, Region):
+            return False
+
+        def check_set(module_set, other_set):
+            if len(module_set) != len(other_set):
+                return False
+            for k, v in other_set.items():
+                if k not in module_set:
+                    return False
+
+        is_srcs_equal = check_set(self.srcs, other.srcs)
+        is_sinks_equal = check_set(self.sinks, other.sinks)
+
+        return is_srcs_equal and is_sinks_equal
 
     @property
     def srcs_names(self):
@@ -184,7 +201,8 @@ class Region:
         # number of output channel.
         # Furthermore, all output channels of all the sources are always fully equalized.
         max_shape_srcs = 0
-        for name, indexes in self.srcs.items():
+        for name, module in self.srcs.items():
+            indexes = module.equalization_indexes
             max_shape_srcs = max(max_shape_srcs, indexes.end + indexes.offset)
         return max_shape_srcs
 
@@ -193,7 +211,12 @@ class Region:
         # Compute the number of input channel from the sinks. If we are equalizing through cat,
         # we need to slice and potentially select only a subset of input channel from sinks.
         max_shape_sinks = 0
-        for name, indexes in self.sinks.items():
+        for name, module in self.sinks.items():
+            indexes = module.equalization_indexes
+
+            if indexes == _UNSUPPORTED_OP:
+                # Invalidate region
+                return 0
             max_shape_sinks = max(max_shape_sinks, indexes.offset + (indexes.end - indexes.start))
         return max_shape_sinks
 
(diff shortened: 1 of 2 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -164,6 +164,23 @@ class Region:
     acts: Tuple = field(default_factory=tuple)
     name_to_module: Dict = field(default_factory=dict)
     expand_region: bool = False
+    act_axis: Optional[int] = None
+
+    def __eq__(self, other):
+        if not isinstance(other, Region):
+            return False
+
+        def check_set(module_set, other_set):
```
pablomlago: No need to define this extra function, the check in the `return` can be written as:
```
return self.srcs.keys() == other.srcs.keys() and  self.sinks.keys() == other.sinks.keys() 
```

src/brevitas/graph/equalize.py
```diff
@@ -164,6 +164,23 @@ class Region:
     acts: Tuple = field(default_factory=tuple)
```
pablomlago: Can we add more detailed type hinting to `srcs`, `sinks` and `acts`? For instance, I would expect the type of `srcs` to be `Dict[str, EqualizationSourceWrapper]`.

src/brevitas/graph/equalize.py
```diff
@@ -184,7 +201,8 @@ def max_shape_srcs(self):
         # number of output channel.
         # Furthermore, all output channels of all the sources are always fully equalized.
         max_shape_srcs = 0
-        for name, indexes in self.srcs.items():
+        for name, module in self.srcs.items():
```
pablomlago, line 194: Maybe I would go for a more Pythonic implementation like:
```
return max(
    (module.equalization_indexes.end + module.equalization_indexes.offset
     for module in self.srcs.values()),
    default=0
)
```
Also, is this formula correct? By reading the comment, I would be expecting something like `indexes.end - indexes.start` instead of `indexes.end + indexes.offset`.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +235,102 @@ def is_valid(self):
         # match the equalizable channels of the sinks
         return self.max_shape_srcs == self.max_shape_sinks
 
+    @property
+    def is_valid_activation_equalization(self):
+        return self.act_axis is not None
+
+    def apply_permute(self, indexes):
```
pablomlago: Remove in this PR.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +235,102 @@ def is_valid(self):
...
+
+    @classmethod
+    def from_dicts(
+            cls,
+            srcs: Optional[Dict[str, EqualizationIndexes]] = None,
+            sinks: Optional[Dict[str, EqualizationIndexes]] = None,
+            name_to_module: Optional[Dict[str, nn.Module]] = None,
+            acts: Optional[Tuple[nn.Module]] = None,
+            expand_region: bool = False):
+        if srcs is None:
```
pablomlago: Maybe a one-linear would be neater, i.e. `srcs = {} if srcs is None else srcs`

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +235,102 @@ def is_valid(self):
...
+            srcs = dict()
+        if sinks is None:
+            sinks = dict()
+        if acts is None:
+            acts = tuple()
+        if name_to_module is None:
+            name_to_module = dict()
+
+        def internal_name_to_module(name_to_module, name):
```
pablomlago: Maybe make a private method within `Region` to declutter this function, i.e. `Region._internal_name_to_module`.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +235,102 @@ def is_valid(self):
...
+            if isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is not None:
+                # The weight attribute to equalize in nn.MultiheadAttention sinks is named "in_proj_weight"
+                weight_tensor_name = "in_proj_weight"
+            elif isinstance(module, nn.MultiheadAttention) and module.in_proj_weight is None:
+                new_sink_dict[name] = _UNSUPPORTED_OP
+                continue
+            else:
+                weight_tensor_name = "weight"
+
+            new_sink_dict[name] = EqualizationSinkWrapper(
```
pablomlago: Maybe I would include the logic above (lines 271-283) in a factory method in `EqualizationSinkWrapper` (and the same for the instantiation of `EqualizationSourceWrapper`) as, in the end, this logic only handles how to properly initialize these classes. With this change, this function would be greatly simplified, i.e. `new_sink_dict = {name: EqualizationSinkWrapper.from_module(module=self._internal_name_to_module(name_to_module, name), indexes=indexes) for name, indexes in srcs.item()}`.

src/brevitas/graph/equalize.py
```diff
@@ -212,6 +235,102 @@ def is_valid(self):
...
+                bias_axis=0,
+                act_axis=act_axis)
+
+        # Act axis tries to determine the channel/feature dimension of the activation tensor
+        # To do this we check sources and sinks, and the type of layers will give us the axis.
+        # The logic is the follow:
+        # - If sources or sinks have undefined act axis (like Batch Norm), we use the other group
+        # - If the act axis of sources and sinks is different, then we set it to None (no axis)
+        # - If both are unsupported/undefined, then we set it to None (no axis)
+        act_sink_axes = {name: module.act_axis for (name, module) in new_sink_dict.items()}
```
pablomlago: This feels like a lot of code and probably could be simplified. For instance:
```
act_axis_map = map(lambda module : module.act_axis, itertools.chain(new_srcs_dict.values(), new_sink_dict.values))
if len(act_axis_map) == 0:
  act_axis = None
else:
   act_axis = next(iter(act_axis_map))
   if any(act_axis != act_axis_el for el in act_axis_map if el is not None)
     ...
```

src/brevitas/graph/equalize.py
```diff
@@ -531,6 +652,7 @@ def __init__(
             module: nn.Module,
             weight_tensor_name: str,
             weight_axis: int,
+            act_axis: int,
```
pablomlago, line 604: Maybe it should be? I'm not 100% sure though.
```suggestion
            act_axis: Optional[int],
```

src/brevitas/graph/equalize.py
```diff
@@ -694,82 +820,30 @@ def _no_equalize():
...
-            module=module,
-            weight_tensor_name=weight_tensor_name,
-            weight_axis=weight_axis,
-            equalization_indexes=indexes,
-        )
-
     # Check if any of the axis is None, which means that the module is not supported.
     # In that case, do not perform graph equalization
-    axes_to_check = [m.weight_axis for m in list(src_axes.values()) + list(sink_axes.values())]
+    axes_to_check = [
```
pablomlago: No need to cast to list if using `itertools.chain`:
```suggestion
    axes_to_check =[m.weight_axis for m in chain(region.srcs.values(), region.sinks.values())]
```

src/brevitas/graph/equalize.py
Giuseppe5: I would rather have it an explicit for-loop. It is more readable (considering we have double property/attribute access)

Giuseppe5: > Also, is this formula correct? 

I would have to double check all the edge-cases of the cat problem but let's say that the current versions cover everything that we planned it to cover.
If anything, we might have to adjust the comment but I'd leave it for now

src/brevitas/graph/equalize.py
Giuseppe5: Not a big deal, and it will be a private class method, not sure if it makes things cleaner anyway

pablomlago (approved):
LGTM, maybe before merging I would ensure that the updated methods have appropriate type hints.

PR #1433, opened 2025-12-17
Feat (brevitas_examples/common): support for mse scale for weights quantized to float formats

JP-Amboage:
## Reason for this PR

It was not possible to run the LLM Examples with a configuration using `weight_quant_format: float_eXmY` and `weight_param_method: mse`. That is, if a "regular" (no OCP) float was being used to quantize the weights, then it was not possible to use mean squared error to choose the scale. I need to use this configuration for some of my experiments.
## Changes Made in this PR
- In `src/brevitas_examples/common/generative/quantizers.py`, I added the class `Fp8e4m3WeightPerChannelFloatMSE` that inherits from `MSESymmetricScale` and `Fp8e4m3WeightPerChannelFloat`. This was done following the same pattern used to support mse scale for Float OCP per-channel weights which was already available.
- In `src/brevitas_examples/common/generative/quantize.py`, I added the entry 
```
'mse': {
                'per_channel': {
                    'sym': Fp8e4m3WeightPerChannelFloatMSE}}}},
```
for `float`, `float_scale` in the dictionary `WEIGHT_QUANT_MAP`. This is how it is currently done for other data types (i.e. Float OCP).

## Testing Summary
- Added a new test in `tests/brevitas_examples/test_llm_cases.py` to ensure that the combination of `weight_quant_format: float_e2m1` with `weight_param_method: mse` runs.
  - In the second commit a small change was done to the tests to skip the new added test when JIT is enabled as this does not support `weight_param_method: mse`.
- Tested locally

### DIFF
--- a/tests/brevitas_examples/test_llm_cases.py
+++ b/tests/brevitas_examples/test_llm_cases.py
@@ -1,7 +1,9 @@
 # Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import pytest
 import pytest_cases
+from brevitas import config
 
 from tests.brevitas_examples.common import process_args_and_metrics
 
@@ -87,6 +89,8 @@ class LLMRunCases:
             "float_e2m1_and_mse"
         ],)
     def case_small_models_toggle_args(self, run_dict, default_run_args, request):
+        if config.JIT_ENABLED and run_dict.get("weight_param_method") == "mse":
+            pytest.skip(reason=f'MSE as weight_param_method requires JIT to be disabled')(f)
         yield process_args_and_metrics(default_run_args, run_dict)
 
 class LLMPerplexityCases:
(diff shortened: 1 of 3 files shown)


tests/brevitas_examples/test_llm_cases.py
```diff
@@ -81,8 +86,11 @@ def case_small_models_with_ppl(self, run_dict, default_run_args, request):
             "quant_sdpa_functional_per_row",
             "functional_sdpa_quant=True,rotation=fused_no_fx",
             "per_group_w_padding,learned_round=linear_round",
+            "float_e2m1_and_mse"
         ],)
     def case_small_models_toggle_args(self, run_dict, default_run_args, request):
+        if config.JIT_ENABLED and run_dict.get("weight_param_method") == "mse":
+            pytest.skip(reason=f'MSE as weight_param_method requires JIT to be disabled')(f)
```
pablomlago: (f) is a leftover from the method this line was extracted, right? I assume it needs to be removed.

JP-Amboage: Yes, I have just removed it and pushed the changes. Thanks for noticing. I'm resolving the conversation as the change has just been addressed.

pablomlago (approved):
LGTM

PR #1434, opened 2025-12-17
Feat (brevitas_examples/llm): support for blockwise rotations

Giuseppe5:
## Reason for this PR
Support for blockwise rotation

## Changes Made in this PR

Allow to select the block dimension for rotation equalization.
The blockwise is applied to all rotations that support it.

To support this, the block dimension should be a divisor of the dimension along which the rotation is performed, otherwise the original shape is used.


src/brevitas/graph/equalize.py
```diff
@@ -1613,6 +1615,10 @@ def _compute_rotations(
                     logging.info("Skipping region")
                     continue
 
+        hidden_dim = hidden_dim if not region.expand_region else expanded_hidden_dim
+        if block_rotation_dim is not None and hidden_dim // block_rotation_dim > 1 and hidden_dim % block_rotation_dim == 0:
```
pablomlago: I would raise an warning in case `hidden_dim` is not divisible by `block_rotation_dim`  (or even an error?), so it does not silently ignore the passed value.

src/brevitas/graph/equalize.py
```diff
@@ -1649,8 +1651,8 @@ def _compute_rotations(
             if region.expand_region:
                 rot_mat, K = expanded_rot_mat, expanded_K
                 assert isinstance(module, nn.Linear), "Currently only Linear layers support expanded hadamard"
-                hidden_dim = module.weight.shape[1]
-                new_hidden = find_closest_hadamard_number(hidden_dim, steps=expansion_step)
+                dim_to_expand = module.weight.shape[1]
+                new_hidden = find_closest_hadamard_number(dim_to_expand, steps=expansion_step)
```
pablomlago: Does this work as intended with block Hadamards? Does it make sense to combine expansion with block rotations?

tests/brevitas_examples/test_llm_cases.py
```diff
@@ -650,6 +650,36 @@ class LLMRotationOptimizationCases:
                 "<class 'torch.nn.utils.parametrize.ParametrizedLinear'>": 1,
                 "<class 'torch.nn.utils.parametrize.ParametrizedEmbedding'>": 1,
                 "<class 'torch.nn.utils.parametrize.ParametrizedQuantLinear'>": 14,}},
+        {
```
pablomlago, line 653: Maybe create a test with expansion to verify that expansion and block Hadamards interact well.

src/brevitas/nn/equalized_layer.py
```diff
@@ -94,6 +96,12 @@ def forward(self, inp, **kwargs):
                 num_features, steps=self.expansion_step)
             inp = pad_to_dim(inp, featured_dim, expanded_num_features)
 
+        init_shape = inp.shape
+        if self.hidden_dim is not None:
+            # This allows us to perform hadamard on a subset of the channel dimension
+            # If init_shape[-1] == had_shape, the next reshape+squeeze is a no-op
+            inp = inp.reshape(*init_shape[:-1], init_shape[-1] // self.hidden_dim,
```
pablomlago, line 103: Ideally, I would have a unit test for this, not just an end-to-end test.

src/brevitas/graph/equalize.py
Giuseppe5: Warning, not an error, just because there is no way of deciding where to apply block hadamard so I would rather have a fallback

src/brevitas/graph/equalize.py
Giuseppe5: This means expansion? I'll add a test

src/brevitas/graph/equalize.py
```diff
@@ -1585,8 +1587,9 @@ def _compute_rotations(
         # Initialize variables
         hidden_dim = region.max_shape_sinks
         expanded_hidden_dim, expanded_rot_mat, expanded_K = None, None, None
-
         if not insert_rotation_module and full_rotation_method == 'ort':
+            assert not region.expand_region, "Orthogonal rotation not compatible with expansion"
```
pablomlago: Can we move these asserts to the start of the method, instead of having them interleaved with the method's logic?

src/brevitas/graph/equalize.py
```diff
@@ -1605,14 +1608,29 @@ def _compute_rotations(
                 rot_func = _apply_had_device
             except AssertionError as e:
                 logging.info(f"Incompatible dim {hidden_dim} for hadamard rotation")
-                if not insert_rotation_module:
+                if not insert_rotation_module and not region.expand_region and block_rotation_dim is None:
```
pablomlago, line 1614: Do we actually needs these extra checks when we are doing the asserts before?

tests/brevitas/graph/test_equalization.py
```diff
@@ -424,12 +429,25 @@ def compare_model_weights(model_fused, model_unfused, classes_to_compare=(nn.Lin
     'mask',
     itertools.product([False, True], repeat=3),
     ids=lambda mask: "-".join([rot for mask_el, rot in zip(mask, ["R1", "R2", "R3"]) if mask_el]))
-@pytest_cases.parametrize('full_rotation_method', ['ort', 'had'])
+@pytest_cases.parametrize('full_rotation_method', ['ort'])
```
pablomlago: Why are we removing `had` here? I see that later in the test there are some checks on `full_rotation_method == 'ort'`.

src/brevitas/graph/equalize.py
```diff
@@ -1605,14 +1608,29 @@ def _compute_rotations(
...
+        hidden_dim = hidden_dim if not region.expand_region else expanded_hidden_dim
+        # Check if we are doing block_rotation and if it is compatible with the current shape
+        if block_rotation_dim is not None:
+            if hidden_dim // block_rotation_dim > 1 and hidden_dim % block_rotation_dim == 0:
+                hidden_dim = block_rotation_dim
+                rot_mat, K = get_hadK(block_rotation_dim)
+                if region.expand_region:
+                    expanded_rot_mat, expanded_K = rot_mat, K
+            else:
+                logging.info(
```
pablomlago, line 1631: If we are skipping, maybe a `return` should be added, no?

pablomlago (approved):
LGTM (assuming test pass)

PR #1436, opened 2025-12-24
Feat (brevitas_examples/llm): better RMSNorm replacement

Giuseppe5:
## Reason for this PR

Some models have weird forward passes for RMSNorm, like gemma that uses `(1 + weight) * input` instead of `(weight) * input`.

Similarly, we were not handling correctly the fact that llama/gemma and maybe other models cast up to float32 and then back to (b)float16 during the forward pass.

## Changes Made in this PR

~~We dynamically create a new class type that inherits from torch.nn.RMSNorm and from whatever RMSNorm class the LLM is using.~~

~~Then we initialize the internal dict manually (a bit scary but it works).~~

~~Finally we replace the newly created Frankestein instance with the old one.~~

~~Importantly, the new instance passes the check `isinstance(module, torch.nn.RMSNorm)`, which we use to apply rotations.~~

All of the above does not work because dynamo cannot trace through custom classes.
Instead what we do now is that we replace just for dynamo to pick up the OG torch.nn.RMSNorm, and then we put back whatever RMSNorm class was originally intended.

Side effect, we need to carry around these classes types for some of our algorithms.

## Testing Summary

All existing, hopefully I didn't break too many.

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1866,7 +1866,8 @@ class GraphRotationEqualization(RotationEqualization):
             layers_to_expand: Optional[List[str]] = None,
             expansion_step: int = None,
             delay_rewriters: bool = False,
-            return_rewriters: bool = False) -> None:
+            return_rewriters: bool = False,
+            extra_rmsnorm_classes: Optional[Tuple] = None) -> None:
         super(GraphRotationEqualization, self).__init__(blacklist_layers, layers_to_expand)
 
         self.supported_srcs = (nn.Linear, nn.Embedding)
@@ -1875,6 +1876,8 @@ class GraphRotationEqualization(RotationEqualization):
         common_scale_invariant.remove(torch.nn.ReLU)
         common_scale_invariant.remove(torch.nn.LeakyReLU)
         self.scale_invariant_layers = tuple(common_scale_invariant) + (RMSNorm,)
+        if extra_rmsnorm_classes is not None:
+            self.scale_invariant_layers = self.scale_invariant_layers + extra_rmsnorm_classes
         self.scale_invariant_function = ()
         self.orphan_sink = orphan_sink
         self.rotate_matmul = rotate_matmul
@@ -2143,9 +2146,11 @@ class LayerNormToRMS(GraphTransform):
 
 class MergeLnAffine(GraphTransform):
 
-    def __init__(self) -> None:
+    def __init__(self, extra_rmsnorm_classes: Optional[Tuple] = None) -> None:
         super(MergeLnAffine, self).__init__()
         self.supported_srcs = (RMSNorm, nn.LayerNorm)
+        if extra_rmsnorm_classes is not None:
+            self.supported_srcs = self.supported_srcs + extra_rmsnorm_classes
         self.supported_sinks = (nn.Linear)
 
     def apply(self, graph_model: GraphModule) -> GraphModule:
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1885,7 +1885,7 @@ class RotationEqualization(GraphTransform):
             return apply_rewriters(model, rewriters)
 
 
-class GraphRotationEqualization(RotationEqualization, StateMixin):
+class GraphRotationEqualization(RotationEqualization, RegionWalkMixin):
 
     def __init__(
             self,
@@ -1911,7 +1911,7 @@ class GraphRotationEqualization(RotationEqualization, StateMixin):
             'supported_sinks': (nn.Linear,),
             'scale_invariant_layers': tuple(common_scale_invariant) + (RMSNorm,),
             'scale_invariant_functions': ()}
-        StateMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
+        RegionWalkMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
 
         self.orphan_sink = orphan_sink
         self.rotate_matmul = rotate_matmul
@@ -2130,7 +2130,7 @@ def apply_rewriters(
     return model
 
 
-class LayerNormToRMS(GraphTransform, StateMixin):
+class LayerNormToRMS(GraphTransform, RegionWalkMixin):
 
     def __init__(
             self,
@@ -2140,7 +2140,7 @@ class LayerNormToRMS(GraphTransform, StateMixin):
 
         base_state_kwargs = {
             'supported_srcs': (nn.Linear, nn.Embedding), 'supported_sinks': (nn.LayerNorm,)}
-        StateMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
+        RegionWalkMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
 
         self.return_rewriters = return_rewriters
         assert RMSNorm is not object, 'Update your Pytorch version to 2.4+'
@@ -2179,14 +2179,14 @@ class LayerNormToRMS(GraphTransform, StateMixin):
             return graph_model
 
 
-class MergeLnAffine(GraphTransform, StateMixin):
+class MergeLnAffine(GraphTransform, RegionWalkMixin):
 
     def __init__(self, extra_state_kwargs: Optional[Dict[str, Tuple]] = None) -> None:
         GraphTransform.__init__(self)
         self.supported_srcs = (RMSNorm, nn.LayerNorm)
         base_state_kwargs = {
             'supported_srcs': (RMSNorm, nn.LayerNorm), 'supported_sinks': (nn.Linear,)}
-        StateMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
+        RegionWalkMixin.__init__(self, **base_state_kwargs, extra_state_kwargs=extra_state_kwargs)
 
     def apply(self, graph_model: GraphModule) -> GraphModule:
         regions = _extract_regions(graph_model, state_impl_kwargs=self.full_state_kwargs)
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -2204,8 +2204,6 @@ class LayerwiseActivationRotation(RotationEqualization):
         self.expansion_step = expansion_step
         self.block_rotation_dim = block_rotation_dim
         self.supported_sinks = (nn.Linear,)
-        # base_state_kwargs = {'supported_sinks': (nn.Linear,)}
-        # StateMixin.__init__(self, base_state_kwargs, extra_state_kwargs)
 
     def apply(self, model: nn.Module) -> nn.Module:
         regions: List[Region] = []
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1785,10 +1785,23 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
         _replace_bias(next_module, new_bias)
 
 
+class StateMixin:
+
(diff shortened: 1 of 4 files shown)


src/brevitas/graph/equalize.py
```diff
@@ -1866,7 +1866,8 @@ def __init__(
             layers_to_expand: Optional[List[str]] = None,
             expansion_step: int = None,
             delay_rewriters: bool = False,
-            return_rewriters: bool = False) -> None:
+            return_rewriters: bool = False,
+            extra_rmsnorm_classes: Optional[Tuple] = None) -> None:
```
pablomlago: I am not a big fan of having this extra argument and requiring to propagate these classes from the context manager. Can we instead extend `_is_scale_invariant_module` (https://github.com/Xilinx/brevitas/blob/master/src/brevitas/graph/equalize.py#L855) to have this extra logic? Ideally, part of the logic of `set(type(x) for x in model.modules() if 'RMS' in type(x).__name__))` could be extracted in a standalone method to be used both by  `_is_scale_invariant_module` and `rmsnorm_patch`.

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
-    for r in rewriters:
-        model = r.apply(model)
-    model = model.to(dtype)
-    return model
+class rmsnorm_patch:
+
+    def __init__(self, model, config, enabled=True):
+        self.model = model
+        self.config = config
+        self.enabled = enabled
```
pablomlago: This attribute seem to only be used in the next line, so I would remove it and just do `if enabled`.

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
+                torch.nn.RMSNorm,
+                normalized_shape=lambda module: module.weight.shape[0],
+                eps=self.config.rms_norm_eps,
+                dtype=dtype,
+                device=device) for rms_cls in self.rmsnorm_classes]
+        dtype = next(iter(self.model.parameters())).dtype
+        for r in rewriters:
+            self.model = r.apply(self.model)
+
+        self.model = self.model.to(dtype)
```
pablomlago, line 52: Why is this cast necessary?

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
+        device = next(self.model.parameters()).device
+        rewriters = [
+            ModuleToModuleByClass(
+                rms_cls,
+                torch.nn.RMSNorm,
+                normalized_shape=lambda module: module.weight.shape[0],
+                eps=self.config.rms_norm_eps,
+                dtype=dtype,
+                device=device) for rms_cls in self.rmsnorm_classes]
+        dtype = next(iter(self.model.parameters())).dtype
```
pablomlago: Why does `dtype` need to be retrieved again? It seems to have been registered at L35 already.

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
+            self.model = r.apply(self.model)
+
+        self.model = self.model.to(dtype)
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        rewriters = []
+        dtype = next(self.model.parameters()).dtype
+        device = next(self.model.parameters()).device
+        for rms_class in self.rmsnorm_classes:
```
pablomlago: This logic seems fragile to me, e.g. if the signature had more than two parameters this would crash. Moreover, there would be no real need to instantiate again the original RMSNorm modules if their references are kept and then assigned back to the model. However, this would require some extra logic, as this behaviour is not supported by `ModuleToModuleByClass`, but I think it could be done easily enough.

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
+            hidden_dim, eps = list(signature(rms_class).parameters.keys())
+            kwargs = {
+                hidden_dim: lambda module: module.normalized_shape, eps: lambda module: module.eps}
+            rewriter = ModuleToModuleByClass(torch.nn.RMSNorm, rms_class, **kwargs)
+            rewriters.append(rewriter)
+
+        for r in rewriters:
+            self.model = r.apply(self.model)
+
+        self.model = self.model.to(dtype)
```
pablomlago, line 62: Why is this cast necessary?

src/brevitas_examples/llm/main.py
```diff
@@ -283,11 +285,13 @@ def quantize_llm(args, extra_args=None):
         remove_hooks(model)
         print(f"Float perplexity ({args.dataset}): {float_ppl:.3f}")
 
-    if args.replace_rmsnorm:
-        model = replace_rmsnorm_with_torch(model, model.config)
+    # if args.replace_rmsnorm:
```
pablomlago: Remove?

pablomlago (changes requested):
Some extra changes might be needed. Also, it might be worth adding some tests for random tiny Gemma models, maybe `hf-internal-testing/dummy-gemma`?

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
Giuseppe5: Because when we re-init modules, we don't propagate correctly dtype and device to quant modules. It could be redudant but better safe

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
Giuseppe5: This is not meant to be robust and to support all type of LLM (present or future).

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
Giuseppe5: As above

src/brevitas/graph/equalize.py
Giuseppe5: The issue is, RMS are not always "scale invariant", for example in the case of other algorithms like weight equalization, they aren't.

Also, this decouples better how we identify the RMSNorm modules versus how we check whether that is a scale invariant function/module.

I agree this is not ideal, but I don't fully agree with your suggestion either.
If anything, there should be a fully general way to customize all the attributes that are used during the region walk algorithm.

The easiest way would be to have a `dict` where a user can override all the keys, but then we would need to handle that within each class (or GraphRotation to start with), which could be verbose but doable.

pablomlago: Why is RMS not "scale invariant" for weight equalization?

Giuseppe5: Because you compute the per-channel variance of the input tensor, which changes if you have a scale factor per channel before/after this op

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
pablomlago: Agreed, but this seems too patchy, specially considering that the only reason why those modules need to be instantiated again is because their references were lost. If you kept those references, there would be no need to make any assumptions on the method's signature, as you would only need to assign the original modules back to the model.

src/brevitas_examples/llm/llm_quant/ln_affine_merge.py
```diff
@@ -3,36 +3,67 @@
...
+            self.mapping.update(r.old_new_module_dict)
+
+        self.model = self.model.to(dtype)
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        rewriters = []
+        dtype = next(self.model.parameters()).dtype
+
+        for old_module, new_module in self.mapping.items():
```
pablomlago, line 58: No need to iterate twice. It can be done as:
```
for old_module, new_module in self.mapping.items():
  r = ModuleInstanceToModuleInstance(old_module, new_module)
  self.model = r.apply(self.model)
```

src/brevitas/graph/equalize.py
```diff
@@ -1785,10 +1785,44 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
         _replace_bias(next_module, new_bias)
 
 
+class StateMixin:
```
pablomlago: Maybe call it `RegionWalkMixin` or `RegionWalkSpec`?

src/brevitas/graph/equalize.py
```diff
@@ -1785,10 +1785,44 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
...
+
+    def __init__(
+            self,
+            supported_srcs: Tuple[nn.Module] = _supported_layers,
+            supported_sinks: Tuple[nn.Module] = _supported_layers,
+            scale_invariant_layers: Tuple[nn.Module] = _scale_invariant_layers,
+            scale_invariant_functions: Tuple[nn.Module] = _scale_invariant_op,
+            residual_fns: Tuple[nn.Module] = _residual_fns,
+            residual_methods: Tuple[nn.Module] = _residual_methods,
+            extra_state_kwargs=None):
```
pablomlago:
```suggestion
            extra_state_kwargs: Optional[Dict[str, Tuple[Type[nn.Module]]]]=None):
```

src/brevitas/graph/equalize.py
```diff
@@ -1785,10 +1785,44 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
...
+        self.supported_srcs = supported_srcs
+        self.supported_sinks = supported_sinks
+        self.scale_invariant_layers = scale_invariant_layers
+        self.scale_invariant_functions = scale_invariant_functions
+        self.residual_fns = residual_fns
+        self.residual_methods = residual_methods
+
+        if extra_state_kwargs is not None:
+            for attr_name in extra_state_kwargs:
```
pablomlago: Next line can be ommited by doing:
```suggestion
            for attr_name, value in extra_state_kwargs.items():
```

src/brevitas/graph/equalize.py
```diff
@@ -1785,10 +1785,44 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
...
+                value = extra_state_kwargs[attr_name]
+                combined_value = value + getattr(self, attr_name)
+                setattr(self, attr_name, combined_value)
+
+    @property
+    def full_state_kwargs(self):
```
pablomlago:
```suggestion
    def full_state_kwargs(self) -> Dict[str, Tuple[Type[nn.Module]]]:
```

src/brevitas/graph/equalize.py
```diff
@@ -1797,19 +1831,19 @@ def __init__(self, blacklist_layers, layers_to_expand) -> None:
             self.layers_to_expand = layers_to_expand
         else:
             self.layers_to_expand = []
-        self.supported_sinks = ()
 
     def find_module(
             self,
             model: nn.Module,
             regions: List[Region],
+            supported_sinks: tuple,
```
pablomlago: For consistency, use legacy type hint `Tuple`:
```suggestion
            supported_sinks: Tuple[nn.Module],
```

src/brevitas/graph/equalize.py
```diff
@@ -1785,10 +1785,44 @@ def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
         _replace_bias(next_module, new_bias)
 
 
+class StateMixin:
+
+    def __init__(
+            self,
+            supported_srcs: Tuple[nn.Module] = _supported_layers,
```
pablomlago: These are not instances of `nn.Module`, but types instead, right? If so, the type hint might be:
```suggestion
            supported_srcs: Tuple[Type[nn.Module]] = _supported_layers,
```

pablomlago (approved):
LGTM (double-check that the failing tests are flaky)

PR #1437, opened 2025-12-24
Feat (graph/gptq): Expose lambda as an argument

JP-Amboage:
## Reason for this PR
This PR allows users to set a custom lambda parameter for GPTQ. Lambda is a hyperparameter of the GPTQ algorithm but currently it always takes a default value (variable `percdamp` in `GPTQ.single_layer_update()`). Therefore, this change gives users more freedom to run GPTQ with different "configurations".

## Changes Made in this PR
- Added a new argument named `gptq_lambda` to `src/brevitas_examples/llm/llm_args.py`. The default value is 0.01 which is the previous default value. The name `gptq_lambda` was chosen instead of `percdamp` to make the argument name more informative for the users.
- The method `apply_gptq` from `src/brevitas_examples/llm/llm_quant/gpxq.py` now takes the argument `percdamp` which is propagated to  the `gptq_class` (`GPTQ` or `A2GPTQ`) using `partial` to make it available in `single_layer_update()`. It was chosen to keep the name `percdamp` internally as this was the one already in use in the GPTQ implementation. Following the review comments and private discussions, it was chosen to use `partial` to keep the parameter `percdamp` decoupled from the context manager `gptq_mode`.

## Testing Summary
- A new test has been added to `LLMRunCases` in `tests/brevitas_examples/test_llm_cases.py` to ensure that the code runs when `gptq_lambda`  is set in the configuration.
- Python debugger has been used to check that the value of  `percdamp` does indeed take the value set with the argument `gptq_lambda`.
- Tests were ran locally.
The following values were achieved for GPTQ used to quantize `meta-llama/Llama-3.2-1B` to int4 with weight only quantization, per channel quantization, and 512 samples from `wikitext2` for calibration:

| `gptq_lambda` | Quantized Perplexity |
| -------- | ------- | 
| 0.005|  10.438| 
| 0.01 (default)| 10.438|
| 0.5| 15.87|

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/gpxq.py
+++ b/src/brevitas_examples/llm/llm_quant/gpxq.py
@@ -120,7 +120,8 @@ def apply_gptq(
         max_accumulator_bit_width=None,
         max_accumulator_tile_size=None,
         buffer_device='cpu',
-        buffer_dtype=torch.float32):
+        buffer_dtype=torch.float32,
+        percdamp: float = .01):
     if max_accumulator_bit_width is not None:
         # Use accumulator-aware extension (AXE) framework
         print(f"Using AXE to target {max_accumulator_bit_width}-bit accumulation...")
@@ -138,7 +139,8 @@ def apply_gptq(
             'use_quant_activations': use_quant_activations,
             'gptq_class': gptq_class,
             'device': buffer_device,
-            'dtype': buffer_dtype}
+            'dtype': buffer_dtype,
+            'percdamp': percdamp}
         block_optimization(model, dataloader, block_name, gptq_mode, context_manager_kwargs)
     else:
         with gptq_mode(model,
@@ -148,7 +150,8 @@ def apply_gptq(
                        create_weight_orig=create_weight_orig,
                        gptq_class=gptq_class,
                        device=buffer_device,
-                       dtype=buffer_dtype) as gptq:
+                       dtype=buffer_dtype,
+                       percdamp=percdamp) as gptq:
             gptq_model = gptq.model
             for _ in tqdm(range(gptq.num_layers)):
                 for inps in dataloader:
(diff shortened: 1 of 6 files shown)


Giuseppe5:
Is there a case where the default is so bad, we need to pick a better option?

src/brevitas_examples/llm/llm_quant/gpxq.py
```diff
@@ -138,7 +139,8 @@ def apply_gptq(
             'use_quant_activations': use_quant_activations,
             'gptq_class': gptq_class,
             'device': buffer_device,
-            'dtype': buffer_dtype}
+            'dtype': buffer_dtype,
+            'percdamp': percdamp}
```
pablomlago: Note that `gptq_class` can be `A2GPTQ`, if `max_accumulator_bit_width is not None`. However, the `__init__` function of this class does not accept `percdamp`, but this parameter is passed through `context_manager_kwargs`, so probably this would crash.  To fix this, one possibility would be to remove `percdamp` from `gptq_mode`, and, in line 133 of this file, do `gptq_class = partial(GPTQ, percdamp=percdamp)`. In the end, there is no need to register `percdamp` in the context manager `gptq_mode`, other than pass it to the `GPTQ` class in `initialize_module_optimizer`.

JP-Amboage: Since `A2GPTQ` also uses the `percdamp` parameter I have added it as an argument in its `__init__` method so that it is also exposed. Following our private discussion, I have removed `percdamp` from the context manager `gptq_mode` and created a `partial` in `apply_gptq` to pass `percdamp` to either `A2GPTQ` or `GPTQ`. The PR description has been updated accordingly.

pablomlago (approved):
LGTM

nickfraser:
> Is there a case where the default is so bad, we need to pick a better option?

There are many cases where GPTQ catastrophically fails. We don't know if this parameter would improve things in these cases, because we don't expose it in many (any?) experiments.

If you're concerned, I'm happy to hold off merging this until some more internal testing has occurred.

PR #1440, opened 2026-01-08
Fix (graph/qronos): Normalize contribution to H and G when buffer is disabled

JP-Amboage:
## Reason for this PR

In  the method `Qronos().update_batch()` (in the file `src/brevitas/graph/qronos.py`), in the case where `self._use_intermediate_buffer`  is `False`, the contribution from the current samples that is being added to the matrices `H` or `G`  is not normalized. However, the current ("pre-update") value of `G` or `H`  is normalized (multiplied by the number of samples and divided by the number of samples plus the number of "incoming" samples).

This is a bug as it causes the contribution from samples from the later updates to weight more on the final `H` and `G` matrices than the contribution of the earlier samples. However, the order of the samples is not relevant and they should all be equally weighted, as it is done in the case where `self._use_intermediate_buffer` is `True` or also in `GPTQ` (only for `H` in this case as `G` is not needed there).

## Changes Made in this PR

In the file `src/brevitas/graph/qronos.py` the line `self.G += inp_processed.bmm(self.quant_input.transpose(2, 1))` inside `Qronos().update_batch()` has been updated so that the incoming contribution is normalized. The new line is 
```python
self.G += (inp_processed * (1 / math.sqrt(self.nsamples))).bmm(
                    self.quant_input.transpose(2, 1) * (1 / math.sqrt(self.nsamples)))
```
The analogous change was done to normalize the contribution to `self.H`.

## Testing Summary

- Tests run locally.
- Ran the code before and after the fix to quantize a Llama 1B. When quantizing to int4 the perplexity result wasn't affected by the bug/the fix. When quantizing to int2 the quantized perplexity before the fix was 1032 and after fixing this bug it went down to 804.

### DIFF
--- a/src/brevitas/graph/qronos.py
+++ b/src/brevitas/graph/qronos.py
@@ -62,25 +62,24 @@ class Qronos(GPFQ):
         if not is_quant_enabled:
             # Computing the normalized G matrix
             self.G *= (self.nsamples - batch_size) / self.nsamples
+            inp_processed /= math.sqrt(
+                self.nsamples)  # NOTE: quant_input is normalized before, in the H update
             if self.use_intermediate_buffer:
-                self.B.copy_((inp_processed * (1 / math.sqrt(self.nsamples))).bmm(
-                    self.quant_input.transpose(2, 1) * (1 / math.sqrt(self.nsamples))))
+                self.B.copy_(inp_processed.bmm(self.quant_input.transpose(2, 1)))
                 self.G += self.B
             else:
-                self.G += (inp_processed * (1 / math.sqrt(self.nsamples))).bmm(
-                    self.quant_input.transpose(2, 1) * (1 / math.sqrt(self.nsamples)))
+                self.G += inp_processed.bmm(self.quant_input.transpose(2, 1))
             self.quant_input = None  # NOTE: set back to None now that we've used it
         else:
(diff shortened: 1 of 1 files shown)


src/brevitas/graph/qronos.py
```diff
@@ -63,20 +63,24 @@ def update_batch(self, module, input, current_layer):
             # Computing the normalized G matrix
             self.G *= (self.nsamples - batch_size) / self.nsamples
             if self.use_intermediate_buffer:
```
pablomlago, line 67: Can we factor the scaling out of the if/else branches to avoid repetition? I think it would be possible to do:
```
self.G *= (self.nsamples - batch_size) / self.nsamples
inp_proccessed /= math.sqrt(self.num_samples)
if self.use_intermediate_buffer:
  quant_input = self.quant_input / math.sqrt(num_samples)
  ...
```

pablomlago (changes requested):
Is it possible to avoid having the normalization scattered in multiple places?

src/brevitas/graph/qronos.py
JP-Amboage: Sounds good!

pablomlago (approved):
LGTM

PR #1441, opened 2026-01-13
Fix (brevitas_examples/llm): more checks for FX-related args

Giuseppe5:
## Reason for this PR
We're missing some checks about how various FX options combine (or don't combine)

## Changes Made in this PR
Added the aforementioned checks

### DIFF
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -499,7 +499,7 @@ def validate(args: Namespace, extra_args: Optional[List[str]] = None) -> None:
         assert args.replace_rmsnorm, 'Graph rotation requires to replace HF RMSNorm with PyTorch ones (torch 2.4+ require)'
         assert args.convert_layernorm_to_rmsnorm, 'Graph rotation requires to replace LayerNorm with RMSNorm'
         # FX is not compatible with few-shot evaluation
-        assert args.few_shot_eval != 'lm_eval' or args.few_shot_eval != 'lighteval', "FX is not compatible with few shot evaluation, use fused_no_fx"
+        assert args.few_shot_eval is None, "FX is not compatible with few shot evaluation, use fused_no_fx"
     # Otherwise we might end up tracing through dynamo twice and other weird errors.
     # Fused_no_fx takes care of all the rotations-related transformations
     if fx_required(args):
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -485,7 +485,7 @@ def create_args_parser() -> ArgumentParser:
     return parser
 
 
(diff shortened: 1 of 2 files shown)


src/brevitas_examples/llm/llm_args.py
```diff
@@ -478,6 +478,10 @@ def create_args_parser() -> ArgumentParser:
     return parser
 
 
+def fx_required(args):
```
pablomlago:
```suggestion
def fx_required(args: Namespace):
```

src/brevitas_examples/llm/llm_args.py
```diff
@@ -478,6 +478,10 @@ def create_args_parser() -> ArgumentParser:
...
+    return True if args.weight_equalization or args.act_equalization == 'fx' or args.rotation == 'fx' or args.ln_affine_merge or args.convert_layernorm_to_rmsnorm or args.quant_sdpa == 'fx' else False
```
pablomlago: This function is also defined in `main.py`. Can we factor it out in a single place? Also:
```suggestion
    return args.weight_equalization or args.act_equalization == 'fx' or args.rotation == 'fx' or args.ln_affine_merge or args.convert_layernorm_to_rmsnorm or args.quant_sdpa == 'fx'
```

src/brevitas_examples/llm/llm_args.py
```diff
@@ -487,6 +491,12 @@ def validate(args: Namespace, extra_args: Optional[List[str]] = None) -> None:
         assert args.ln_affine_merge, 'Graph rotation requires to merge LN/RMS norm affine parameters'
         assert args.replace_rmsnorm, 'Graph rotation requires to replace HF RMSNorm with PyTorch ones (torch 2.4+ require)'
         assert args.convert_layernorm_to_rmsnorm, 'Graph rotation requires to replace LayerNorm with RMSNorm'
+        # FX is not compatible with few-shot evaluation
+        assert args.few_shot_eval != 'lm_eval' or args.few_shot_eval != 'lighteval', "FX is not compatible with few shot evaluation, use fused_no_fx"
```
pablomlago: Is this right? I feel this always evaluates to `True`. Consider:
```suggestion
        assert args.few_shot_eval is None, "FX is not compatible with few shot evaluation, use fused_no_fx"
```

src/brevitas_examples/llm/llm_args.py
Giuseppe5: Already removed from main

src/brevitas_examples/llm/llm_args.py
Giuseppe5: I believe it was `and` not `or` but your suggestion works too

pablomlago (approved):
LGTM

PR #1444, opened 2026-01-24
Feat (vLLM): experimental support for vLLM export

Giuseppe5:
## Reason for this PR

Initial support for vLLM export.

To do:
- [x]  Check that input/output quantization work as intended
- [x] Test multiple quantizers
- [x] Improve quantizers interface
- [x] Support for rotation (and smoothquant?)

## Changes Made in this PR

We are re-using the inference quantizers also for vLLM.
This is still fake-quantization style, but _should_ be faster than plain torch execution, even in eager mode.

The same template could be easily extended to support real quantization, torch.compile, etc. etc.

## Testing Summary

Added tests for vLLM inference handlers


requirements/requirements-llm.txt
```diff
@@ -10,3 +10,4 @@ pydantic
 torch>=2.4
 tqdm
 transformers[sentencepiece]<5.0
+vllm
```
pablomlago: I feel like vLLM should be an optional dependency.

Giuseppe5: Maybe we can do it in a similar way to what we did for lighteval/lm_eval

Giuseppe5: I'm leaving it for now so that test run and I can see what other things I'm breaking in the process, but I'll remove before this PR is merged

pablomlago: I'm fine with doing it similarly as for _lighteval/lm_eval_

src/brevitas/core/function_wrapper/shape.py
```diff
@@ -186,19 +186,21 @@ def __init__(self, group_size, group_dim) -> None:
...
-        tensor_shape_list.insert(block_dim, self.group_size)
-        x = x.view(tensor_shape_list)
-        return x
+
+def dynamic_over_sub_channel_block_view(x, group_size, group_dim):
+    tensor_shape = x.shape
+    tensor_shape_list = list(tensor_shape)
+    x = padding_to_multiple(x, group_dim, group_size)
+
+    tensor_shape = x.shape
```
pablomlago: Is there a way to do simplify this logic? A function like `torch.unflatten` might be useful.

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,184 @@
...
+    @classmethod
+    def get_supported_act_dtypes(cls) -> list[torch.dtype]:
+        return [torch.float16, torch.bfloat16, torch.float32]
+
+    @staticmethod
+    def get_config_filenames() -> list[str]:
+        return ["brevitas_config.json"]
+
+    def get_quant_method(self, layer: torch.nn.Module,
+                         prefix: str) -> Optional["QuantizeMethodBase"]:
```
pablomlago, line 84: The `__init__` uses typing from Python 3.10+, e.g. `ignored_layers: list[str] | None = None`, while, in this method, the legacy typing is used `Optional["QuantizeMethodBase"]`. Can we use Python 3.10+ typing for consistency with vLLM.

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,184 @@
...
+from ..manager import _override_bias_caching_mode
+from ..manager import _override_create_quant_tensor
+from ..manager import _override_weight_caching_mode
+
+
+@register_quantization_config("quant_brevitas")
+@dataclass
+class QuantConfigBrevitas(QuantizationConfig):
+
+    def __init__(self, ignored_layers: list[str] | None = None, config: str | None = None):
```
pablomlago: Is this `config: str | None = None` the correct typing?

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,184 @@
...
+        super().__init__()
+        self.ignored_layers = ignored_layers
+        self.config = config
+
+    @classmethod
+    def from_config(cls, config: dict[str, Any]) -> "QuantConfigTcast":
```
pablomlago:
```suggestion
    def from_config(cls, config: dict[str, Any]) -> "QuantConfigBrevitas":
```

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @staticmethod
+    def get_config_filenames() -> list[str]:
+        return ["brevitas_config.json"]
+
+    def get_quant_method(self, layer: torch.nn.Module,
+                         prefix: str) -> Optional["QuantizeMethodBase"]:
+        if isinstance(layer, RowParallelLinear) or isinstance(
+                layer, MergedColumnParallelLinear) or isinstance(layer, QKVParallelLinear):
+            if self.ignored_layers and is_layer_skipped(
```
pablomlago: Is the check `self.ignored_layers` needed? Maybe
```suggestion
            if is_layer_skipped(
```
would suffice, from what I see in other classes, e.g. `Fp8Config`.

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,184 @@
...
+class_mapping = {
+    'GroupwiseFloatInferenceHandler': GroupwiseFloatInferenceHandler,
+    'GroupwiseFloatWeightInferenceHandler': GroupwiseFloatWeightInferenceHandler,
+    'FloatInferencetHandler': FloatInferencetHandler,
+    'FloatWeightInferencetHandler': FloatWeightInferencetHandler,
+    'IntWeightInferencetHandler': IntWeightInferencetHandler,
+    'IntInferencetHandler': IntInferencetHandler,}
+
+
+class QuantLinear(LinearMethodBase):
```
pablomlago: Consider adapting this class to better match the structure of examples such as `vllm.model_executor.layers.quantization.fp_quant`

src/brevitas/quant/solver/common.py
```diff
@@ -63,6 +72,25 @@ def solve_float_to_int_impl_from_enum(impl_type):
         raise Exception(f"{impl_type} not recognized.")
 
 
+def solve_float_to_int_enum_from_impl(impl_type):
```
pablomlago: This function introduces an extra maintainability burden, i.e. every time one adds a new `FloatToIntImplType`, two updates need to be done. However, this pattern matching could be done programatically by iterating over the possible values `FloatToIntImplType`, resolving them to classes and then selecting the value of the enumeration corresponding to the given class (which would amount to generating the dictionary `FLOAT_TO_INT_IMPL_TO_ENUM` programatically).

src/brevitas/quant/solver/common.py
```diff
@@ -91,6 +119,19 @@ def solve_restrict_value_impl_from_enum(impl_type):
         raise RuntimeError(f"{impl_type} not recognized.")
 
 
+def solve_restrict_value_enum_from_impl(impl):
```
pablomlago: Same as above

src/brevitas/quant/solver/common.py
```diff
@@ -43,6 +43,15 @@
     'SolveDtypeDeviceFromTrackedParameterList',
     'SolveRestrictScaleSign']
 
+# FLOAT_TO_INT_ENUM_TO_IMPL = {FloatToIntImplType.ROUND: RoundSte,
```
pablomlago: Remove

src/brevitas/nn/equalized_layer.py
```diff
@@ -112,10 +123,24 @@ def forward(self, inp, **kwargs):
...
-        o = self.layer(inp)
+        return inp
+
+    def forward(self, inp, **kwargs):
+        inp = self.rotation_forward(inp)
+        o = self.layer(inp, **kwargs)
 
         return o
 
+    def state_dict(self, destination=None, prefix='', keep_vars=False):
```
pablomlago: This change in how `state_dict` works is specifically required for vLLM, but I don't see a reason why it should be done always. Is it possible to do this override only in the vLLM export flow?

src/brevitas/nn/equalized_layer.py
```diff
@@ -64,6 +64,17 @@ def forward(self, *args, **kwargs):
         out = self.layer(*kwargs.values())
         return out
 
+    def state_dict(self, destination=None, prefix='', keep_vars=False):
```
pablomlago: Same as below

src/brevitas/export/inference/handler.py
```diff
@@ -25,10 +32,166 @@
...
+        init_shape = list(x.shape)
+        shape = init_shape
+        assert shape[group_dim] % group_size == 0
+        shape[group_dim] = shape[group_dim] // group_size
+        extra_dim = group_dim + 1 if group_dim != -1 else -1
+        shape.insert(extra_dim, group_size)
+        x = x.reshape(shape)
+        return x
+
+    # def compute_scale(self, x, group_dim):
```
pablomlago: Remove

src/brevitas/nn/equalized_layer.py
Giuseppe5: I think we could do it only for export, but I wouldn't be against having this as default behaviour, since these layers are wrappers and modifying the state dict is annoying more often than not.
Also, there is no obvious drawback in flattening the state dict to ignore the added hierarchy.

src/brevitas/export/inference/vLLM/manager.py
Giuseppe5: The actual typing is much more complicated than Dict, not sure if we should be super explicit

src/brevitas/core/function_wrapper/shape.py
```diff
@@ -186,19 +186,19 @@ def __init__(self, group_size, group_dim) -> None:
...
-        tensor_shape = x.shape
-        tensor_shape_list = list(tensor_shape)
-        tensor_shape_list[self.group_dim] = (
-            tensor_shape_list[self.group_dim] + self.group_size - 1) // self.group_size
-        block_dim = self.group_dim + 1 if self.group_dim != -1 else len(tensor_shape_list)
-        tensor_shape_list.insert(block_dim, self.group_size)
-        x = x.view(tensor_shape_list)
-        return x
+
+def dynamic_over_sub_channel_block_view(x, group_size, group_dim):
```
pablomlago: Add type hinting in new methods

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,47 @@
...
+from ..handler import GroupwiseIntInferenceHandler
+
+
+class StandaloneGroupwiseQuantMixin(DynamicScaleZeroPointMixin):
+
+    def compute_scale(self, x, group_dim):
+        scale = torch.clamp(torch.max(torch.abs(x), dim=group_dim, keepdim=True)[0], 1e-4)
+        threshold = self.threshold
+        if self.scaling_restriction == RestrictValueType.POWER_OF_TWO:
+            scale = torch.clamp(torch.pow(2, torch.floor(torch.log2(scale))), 1e-7)
```
pablomlago: Is this clamp needed, considering the scale is already clamped in L15?

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,150 @@
...
+from ..handler import IntInferencetHandler
+from ..handler import IntWeightInferencetHandler
+
+class_mapping = {
+    'GroupwiseFloatInferenceHandler': GroupwiseFloatInferenceHandler,
+    'GroupwiseFloatWeightInferenceHandler': GroupwiseFloatWeightInferenceHandler,
+    'FloatInferencetHandler': FloatInferencetHandler,
+    'FloatWeightInferencetHandler': FloatWeightInferencetHandler,
+    'IntWeightInferencetHandler': IntWeightInferencetHandler,
+    'IntInferencetHandler': IntInferencetHandler,}
```
pablomlago: Typo
```suggestion
    'IntInferenceHandler': IntInferenceHandler,}
```

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(
+            self,
+            input_config=None,
+            weight_config=None,
+            bias_config=None,
+            output_config=None,
+            rotation_config=None):
+        self.input_quant = self.configure_proxy(input_config)
+        if isinstance(weight_config, list):
+            self.weight_quant = dict()
```
pablomlago:
```suggestion
            self.weight_quant = {i: self.configure_proxy(config) for i, config in enumerate(weight_config)}
```

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,150 @@
...
+    'FloatWeightInferencetHandler': FloatWeightInferencetHandler,
+    'IntWeightInferencetHandler': IntWeightInferencetHandler,
+    'IntInferencetHandler': IntInferencetHandler,}
+
+
+class QuantLinear(LinearMethodBase):
+
+    def __init__(
+            self,
+            input_config=None,
```
pablomlago:
```suggestion
            input_config: Optional[Dict[str, Any]]=None,
```

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,150 @@
...
+            self.weight_quant = dict()
+            for i, config in enumerate(weight_config):
+                self.weight_quant[i] = self.configure_proxy(config)
+        else:
+            self.weight_quant = self.configure_proxy(weight_config)
+        self.bias_quant = self.configure_proxy(bias_config)
+        self.output_quant = self.configure_proxy(output_config)
+        self.rotation = self.configure_rotation(rotation_config)
+
+    def configure_rotation(self, rotation_config):
```
pablomlago: A bit weird, to either return a `Module` (`torch.nn.Identity()`) or return a method from an instance of `RotatedModule` (which has a parameter `rot_mat` that cannot be accessed as its reference is not kept). Is it possible to refactor `rotation_forward` in a stateless method, and have `rot_mat` registered to `QuantLinear`?

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,150 @@
...
+        self,
+        layer: torch.nn.Module,
+        input_size_per_partition: int,
+        output_partition_sizes: List[int],
+        input_size: int,
+        output_size: int,
+        params_dtype: torch.dtype,
+        **extra_weight_attrs,
+    ):
+        from vllm.model_executor.parameter import ModelWeightParameter
```
pablomlago: Does it need to live here?

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,245 @@
...
+                    prefix=prefix,
+                    ignored_layers=self.ignored_layers,
+                    fused_mapping=self.packed_modules_mapping,
+            ):
+                return UnquantizedLinearMethod()
+            else:
+
+                if prefix in self.config:
+                    base_config = self.config[prefix]
+                    input_config = base_config.get('input_quant', None)
```
pablomlago: Can we streamline this? E.g.:
```
quant_configs = {f"{param}_config": base_config.get(f"{param}_quant", None) for param in ["weight",...]},
```
and unpack in the initialization of `QuantLinear`.

src/brevitas/quant/solver/common.py
```diff
@@ -91,6 +101,16 @@ def solve_restrict_value_impl_from_enum(impl_type):
         raise RuntimeError(f"{impl_type} not recognized.")
 
 
+def solve_restrict_value_enum_from_impl(impl):
+    programmatic_dict = dict()
```
pablomlago: I'd use a more descriptive name, and do dict-comprehension.

src/brevitas/quant/solver/common.py
```diff
@@ -91,6 +101,16 @@ def solve_restrict_value_impl_from_enum(impl_type):
...
+    for enum_value in RestrictValueType:
+        programmatic_dict[solve_restrict_value_impl_from_enum(enum_value)] = enum_value
+    if impl in programmatic_dict:
```
pablomlago: For clarity, I'd do:
```
if impl not in programmatic_dict:
  raise RuntimeError(f"{impl} not recognized.")
return programmatic_dict[impl]
```

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,109 @@
+from typing import Tuple
```
pablomlago, line 6: Add Copyright header (there are several files missing it).

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,109 @@
...
+
+from ..handler import DynamicFloatInferenceHandler
+from ..handler import DynamicScaleZeroPointMixin
+from ..handler import GroupwiseFloatInferenceHandler
+from ..handler import GroupwiseIntInferenceHandler
+
+EPS = 1e-16
+
+
+def maybe_permute(x, permute_dims):
```
pablomlago: Add type hints.

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,109 @@
...
+        out = groupwise_dequant_expand(out, scale, zero_point, self.group_dim, inp_shape)[0]
+        return out, scale, zero_point, self.exponent_bit_width, self.mantissa_bit_width, self.exponent_bias, self.saturating, self.inf_values, self.nan_values
+
+
+class vLLMDynamicPerRowFloatInferenceHandler(DynamicFloatInferenceHandler,
+                                             StandaloneGroupwiseQuantMixin):
+    handled_layer = DynamicActFloatQuantProxyFromInjector
+
+    def __init__(self):
+        DynamicFloatInferenceHandler.__init__(self)
```
pablomlago, line 84: Why not call `super().__init__()` instead of L79-80?

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,109 @@
...
+        StandaloneGroupwiseQuantMixin.__init__(self)
+        self.register_buffer("permute_dims", None)
+        self.stats_reduce_dim = 1
+
+    def prepare_for_export(self, module):
+        DynamicFloatInferenceHandler.prepare_for_export(self, module)
```
pablomlago: I think it is also possible to simplify this by calling super().prepare_for_export(module) (my understanding is that this would call `prepare_for_export` for both classes in MRO order).

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,155 @@
...
+    'vLLMGroupwiseFloatInferenceHandler': vLLMGroupwiseFloatInferenceHandler,
+    'GroupwiseFloatWeightInferenceHandler': GroupwiseFloatWeightInferenceHandler,
+    'FloatInferencetHandler': FloatInferencetHandler,
+    'FloatWeightInferencetHandler': FloatWeightInferencetHandler,
+    'IntWeightInferencetHandler': IntWeightInferencetHandler,
+    'IntInferenceHandler': IntInferenceHandler,
+    'vLLMDynamicPerRowFloatInferenceHandler': vLLMDynamicPerRowFloatInferenceHandler}
+
+
+class QuantLinear(LinearMethodBase):
```
pablomlago, line 39: is it possible to have tests in the style of [test_fp8.py](https://github.com/vllm-project/vllm/blob/4ff8c3c8f9ece010a1d0e376f5cc1b468b95f366/tests/quantization/test_fp8.py)  for this layer? Even if it is in a future PR, it would be desirable.

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+        elif isinstance(layer, LinearBase):
+            return UnquantizedLinearMethod()
+
+        return None
+
+
+def combine_configs(config, *names):
+    base_config = config[names[0]]
+    scale = None  #base_config['scale']
```
pablomlago, line 125: Should this comment be removed?

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,243 @@
...
+    def set_export_mode(cls, model: Module, enabled: bool):
+        _set_proxy_export_mode(model, enabled)
+        _set_recurrent_layer_export_mode(model, enabled)
+
+    @classmethod
+    def set_export_handler(cls, module: Module):
+        _set_proxy_export_handler(cls, module)
+        _set_recurrent_layer_export_handler(cls, module)
+
+    wrap_layers = (EqualizedModule, RotatedModule)
```
pablomlago: Define class variables together and before the methods for readability.

src/brevitas/export/inference/handler.py
```diff
@@ -25,10 +32,152 @@
 from brevitas.proxy.runtime_quant import ActQuantProxyFromInjector
 from brevitas.proxy.runtime_quant import DynamicActQuantProxyFromInjector
 from brevitas.quant.experimental.mx_quant_ocp import GroupwiseActQuantProxyFromInjector
+from brevitas.quant.solver.act import solve_float_to_int_impl_from_enum
+from brevitas.quant.solver.common import solve_float_to_int_enum_from_impl
+from brevitas.quant.solver.common import \
+    solve_restrict_value_enum_from_impl  # FLOAT_TO_INT_IMPL_TO_ENUM
```
pablomlago:
```suggestion
    solve_restrict_value_enum_from_impl
```

src/brevitas/export/inference/handler.py
```diff
@@ -25,10 +32,152 @@
...
+        return self._scaling_restriction
+
+    @scaling_restriction.setter
+    def scaling_restriction(self, value):
+        if isclass(value):
+            self._scaling_restriction = solve_restrict_value_enum_from_impl(value)
+        elif isinstance(value, str):
+            self._scaling_restriction = value
+        else:
+            raise "Unrecognized scaling restriction"
```
pablomlago: I think this would fail, with the following error:
```
  File "<stdin>", line 1, in <module>
TypeError: exceptions must derive from BaseException
```
```suggestion
            raise ValueError("Unrecognized scaling restriction")
```

src/brevitas/export/inference/handler.py
```diff
@@ -25,10 +32,152 @@
...
+        return self._threshold_restriction
+
+    @threshold_restriction.setter
+    def threshold_restriction(self, value):
+        if isclass(value):
+            self._threshold_restriction = solve_restrict_value_enum_from_impl(value)
+        elif isinstance(value, str):
+            self._threshold_restriction = value
+        else:
+            raise "Unrecognized scaling restriction"
```
pablomlago: Same as above

src/brevitas/export/inference/handler.py
```diff
@@ -46,32 +195,30 @@ def quantize(self, x: Tensor):
...
-class IntInferencetHandler(InferenceHandler):
-    handled_layer = (ActQuantProxyFromInjector, BiasQuantProxyFromInjector)
+
+class IntInferenceHandlerBase(InferenceHandler, FloatToIntMixin):
 
     def __init__(self):
-        super().__init__()
-        self.register_buffer('scale', torch.ones(1))
-        self.register_buffer('zero_point', torch.ones(0))
+        InferenceHandler.__init__(self)
```
pablomlago: Same comment as above, call `super().__init__()`. It applies to the following classes too.

src/brevitas/export/inference/handler.py
```diff
@@ -166,84 +336,78 @@ def inner_forward(self, x: Tensor, scale: Tensor, zero_point: Tensor) -> Tensor:
...
+        FloatToIntMixin.__init__(self)
+        self.register_buffer('min_clamp', torch.ones(()))
+        self.register_buffer('max_clamp', torch.ones(()))
+        self.register_buffer('mantissa_bit_width', torch.ones(()))
+        self.register_buffer('exponent_bit_width', torch.ones(()))
+        self.register_buffer('exponent_bias', torch.ones(()))
+        self.register_buffer('fp_internal_scale_min', torch.ones(()))
+        self.inf_values = None
+        self.nan_values = None
+        self.eps = 1e-8  #torch.finfo(self.scale.dtype).tiny
```
pablomlago: Remove comment or do change.

src/brevitas/export/inference/handler.py
```diff
@@ -166,84 +336,78 @@ def inner_forward(self, x: Tensor, scale: Tensor, zero_point: Tensor) -> Tensor:
...
-    handled_layer = (ActFloatQuantProxyFromInjector, BiasQuantProxyFromInjector)
+class FloatInferenceHandlerBase(InferenceHandler, FloatToIntMixin):
 
     def __init__(self):
-        super().__init__()
-        self.register_buffer('scale', torch.ones(1))
-        self.register_buffer('zero_point', torch.ones(0))
+        InferenceHandler.__init__(self)
+        FloatToIntMixin.__init__(self)
+        self.register_buffer('min_clamp', torch.ones(()))
```
pablomlago: I'm not a big fan of these tensors that do not have a `dtype` or `device`. Isn't it possible for the constructor to accept these arguments, thus avoiding the casts/device in the `forward`. In fact, the `.to` in `forward` only handle device moves, so I would expect errors to appear when `x` has a `dtype` different to the default one.

src/brevitas/export/inference/manager.py
```diff
@@ -55,12 +53,18 @@ def _override_create_quant_tensor(m: nn.Module, state: bool):
...
+            cache_quant_weight=False,
+            compile=False,
+            enabled=True,
+            export_manager=None):
         self.model = model
         self.enabled = enabled
         self.compile = compile
         self.cache_quant_weight = cache_quant_weight
-        self.export_manager = InferenceManager
+        self.export_manager = export_manager if export_manager is not None else InferenceManager
```
pablomlago: Why not set the default value of `export_manager` to `InferenceManager` instead of `None` to avoid this check?

src/brevitas/quant/solver/common.py
```diff
@@ -63,6 +63,15 @@ def solve_float_to_int_impl_from_enum(impl_type):
         raise Exception(f"{impl_type} not recognized.")
 
 
+def solve_float_to_int_enum_from_impl(impl_type):
+    impl_to_enum_dict = {
+        solve_float_to_int_impl_from_enum(enum_value): enum_value
+        for enum_value in FloatToIntImplType}
+    if impl_type not in impl_to_enum_dict:
+        raise Exception(f"{impl_type} not recognized.")
```
pablomlago: I think `NotImplementedError` or `ValueError` might be more appropiate.
```suggestion
        raise ValueError(f"{impl_type} not recognized.")
```

src/brevitas/config.py
```diff
@@ -16,6 +16,7 @@ def env_to_bool(name, default):
 
 REINIT_ON_STATE_DICT_LOAD = env_to_bool('BREVITAS_REINIT_ON_STATE_DICT_LOAD', True)
 IGNORE_MISSING_KEYS = env_to_bool('BREVITAS_IGNORE_MISSING_KEYS', False)
+IGNORE_EXPORT_KEYS = env_to_bool('BREVITAS_IGNORE_EXPORT_KEYS', False)
```
pablomlago: I do not think we should define this variable as if it was an environment variable. We are not expecting the user to set it to any value, since, to the best of my knowledge, it is used as a global variable that can be modified in any part of the codebase. Moreover, the pattern is always
```
config.IGNORE_EXPORT_KEYS = True
# some mode
config.IGNORE_EXPORT_KEYS = False
```
which suggests a context manager might be a good option, to avoid situations in which the last statement is forgotten, or similar. Therefore, I think a sensible option would be to rely on a `ContextVar` for this variable, and maybe the pattern used in this [PR](https://github.com/Xilinx/brevitas/pull/1452) is a good reference.

pablomlago (changes requested):
Added some extra comments.

src/brevitas/export/inference/vLLM/handler.py
Giuseppe5: MRO is a nightmare, and from what I saw, super will stop at the first parent class that implements the desired method. If multiple parents have the same method, all subsequent one after the first are ignored, which we do not want.

src/brevitas/export/inference/vLLM/handler.py
Giuseppe5: No only the first, unfortunately.

src/brevitas/export/inference/vLLM/layer.py
Giuseppe5: Sure, I'd leave it for another PR since for now we're testing all the proxies.
I'll add it to the test issue

src/brevitas/export/inference/handler.py
Giuseppe5: I can set device and dtype as optional arguments (which is the typical structure of torch layers)

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class vLLMGroupwiseIntInferenceHandler(GroupwiseIntInferenceHandler, StandaloneGroupwiseQuantMixin):
+
+    def forward(self, x: Tensor) -> Tuple[Tensor, ...]:
+        inp_shape = x.shape
+        x = dynamic_over_sub_channel_block_view(x, self.group_size, self.group_dim)
+        group_dim = self.group_dim + 1 if self.group_dim > 0 else self.group_dim
+        scale = self.compute_scale(x, group_dim)
+        zero_point = torch.zeros(()).type_as(x)
```
pablomlago, line 59: Consider doing:
```
zero_point = torch.zeros((), dtype=x.dtype, device=x.device)
```
or
```
zero_point = x.new_zeros(())
```

to avoid unnecessary CPU to GPU moves, and casting.

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,114 @@
...
+
+class vLLMGroupwiseFloatInferenceHandler(GroupwiseFloatInferenceHandler,
+                                         StandaloneGroupwiseQuantMixin):
+
+    def forward(self, x: Tensor) -> Tuple[Tensor, ...]:
+        inp_shape = x.shape
+        x = dynamic_over_sub_channel_block_view(x, self.group_size, self.group_dim)
+        group_dim = self.group_dim + 1 if self.group_dim > 0 else self.group_dim
+        scale = self.compute_scale(x, group_dim)
+        zero_point = torch.zeros(()).type_as(x)
```
pablomlago: Same as above

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,159 @@
...
+            if not isinstance(v, torch.Tensor):
+                if k == 'scale' or k == 'zero_point':
+                    quant_config[k] = torch.tensor(v)
+                else:
+                    quant_config[k] = torch.tensor(v, dtype=torch.int)
+
+        # Shapes must be set otherwise the state dict loading will fail
+        scale = quant_config.get('scale', None)
+        zero_point = quant_config.get('zero_point', None)
+        quant_class_type = class_mapping[quant_class_name]
```
pablomlago: The names of the variables are a bit confusing, I would opt for:

- `quant_class_type` -> `quant_class`
- `quant_class` -> `quantizer`

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,159 @@
...
+            return None
+        rot_mat_shape = rotation_config['rot_mat_shape']
+        k = rotation_config['k']
+        if rot_mat_shape is None:
+            had_mat = None
+        else:
+            had_mat, _ = get_hadK(rot_mat_shape)
+        return RotatedModule(self, had_mat, k)
+
+    # TODO: return type is Union[torch.nn.Identity, InferenceHandler] but handler type hierarchy is complex
```
pablomlago: Does this comment need to be removed?

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def create_weights(
+            self,
+            layer: torch.nn.Module,
+            input_size_per_partition: int,
+            output_partition_sizes: List[int],
+            input_size: int,
+            output_size: int,
+            params_dtype: torch.dtype,
+            **extra_weight_attrs):
```
pablomlago:
```suggestion
            **extra_weight_attrs) -> None:
```

src/brevitas/export/inference/vLLM/layer.py
```diff
@@ -0,0 +1,159 @@
...
+    def apply(
+            self,
+            layer: torch.nn.Module,
+            x: torch.Tensor,
+            bias: Optional[torch.Tensor] = None) -> torch.Tensor:
+        if self.rotation is not None:
+            x = self.rotation.rotation_forward(x)
+        x = self.input_quant(x)[0]
+        bias = self.bias_quant(bias) if bias is not None else None
+        y = x.matmul(layer.weight.t())
```
pablomlago: Can we replace the block:
```
        y = x.matmul(layer.weight.t())
        if bias is not None:
            y = y + bias
```
by a call to `torch.nn.functional.linear`?

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -0,0 +1,114 @@
...
+    handled_layer = DynamicActFloatQuantProxyFromInjector
+
+    def __init__(self) -> None:
+        DynamicFloatInferenceHandler.__init__(self)
+        StandaloneGroupwiseQuantMixin.__init__(self)
+        self.register_buffer("permute_dims", None)
+        self.stats_reduce_dim = 1
+
+    def prepare_for_export(self, module):
+        DynamicFloatInferenceHandler.prepare_for_export(self, module)
```
pablomlago: Is this missing the change to use `super`?

src/brevitas/export/inference/vLLM/manager.py
```diff
@@ -0,0 +1,235 @@
...
+                            export_handler, 'scaling_restriction', None)
+                        proxy_dict['float_to_int_impl_type'] = export_handler.float_to_int_impl_type
+                        proxy_dict['class_type'] = export_handler.__class__.__name__
+                        json_to_save[name][proxy_name] = proxy_dict
+
+        json_filename = os.path.join(filepath, 'brevitas_config.json')
+        with open(json_filename, 'w+') as f:
+            json.dump(json_to_save, f, cls=EncodeTensor)
+
+        token = config.IGNORE_PROXY_KEYS.set(True)
```
pablomlago, line 232: Can we do this as a context manager? I am also wondering whether that `IGNORE_PROXY_KEYS` should live under `config`.

src/brevitas/export/inference/handler.py
```diff
@@ -166,84 +371,80 @@ def inner_forward(self, x: Tensor, scale: Tensor, zero_point: Tensor) -> Tensor:
...
                 self.float_clamp_impl = module.fused_activation_quant_proxy.tensor_quant.float_clamp_impl
                 self.max_available_float = module.fused_activation_quant_proxy.tensor_quant.float_clamp_impl.max_available_float
 
             self.pre_compute_max_mantissa = compute_max_mantissa(self.mantissa_bit_width)
             self.max_clamp = max_float(
                 self.exponent_bit_width, self.pre_compute_max_mantissa, self.exponent_bias)
-            self.min_clamp = -self.max_clamp
+            self.max_clamp = self.max_clamp if self.max_available_float is None else torch.min(
+                self.max_clamp, self.max_available_float())
+            self.min_clamp = torch.tensor(0.) if not module.is_signed else -self.max_clamp
```
pablomlago: `torch.tensor(0.)` will potentially not match the device/dtype of the other tensors without explicit casting.

src/brevitas/config.py
```diff
@@ -22,6 +23,8 @@ def env_to_bool(name, default):
 VERBOSE = env_to_bool('BREVITAS_VERBOSE', False)
 LOGGING_LEVEL = os.environ.get('BREVITAS_LOGGING', 'CRITICAL')
 
+IGNORE_PROXY_KEYS = ContextVar("IGNORE_PROXY_KEYS", default=False)
```
pablomlago, line 26: Should this live somewhere else? Also, consider doing the pattern:
```
@contextmanager
def set_proxy_keys_cm(enabled: bool):
    token = IGNORE_PROXY_KEYS.set(enabled)
    try:
        yield
    finally:
        IGNORE_PROXY_KEYS.reset(token)
```

tests/brevitas/export/test_vllm_inference_mode.py
```diff
@@ -0,0 +1,124 @@
...
+    'fp8': Fp8e4m3ActPerTensorFloat,
+    'per_row_dynamic_fp8': FP8e4m3OCPDynamicActPerRowFloat,
+    'mxint8': MXInt8Act,
+    'mxfloat8': MXFloat8e4m3Act,}
+
+
+@pytest_cases.parametrize('weight_quantizer', WEIGHT_QUANTIZERS.items())
+@given(weight=float_tensor_st(shape=(8, 16), max_val=1e10, min_val=-1e10))
+@requires_pt_ge('2.1')
+def test_vllm_compile_weight(weight, weight_quantizer):
```
pablomlago: Why the test name has `compile` if this flag is set to `False`? Wouldn't it be possible to parametrize the test with `compile` being either `True` or `False`?

pablomlago (approved):
Some minor changes

src/brevitas/export/inference/vLLM/manager.py
Giuseppe5: We could do it as a context manager but considering this is the only case (for now) where it's used, I don't think it's fundamental.
For config, we have everything there now so I'd leave it there for consistency while we decide if we want to move them somewhere else

tests/brevitas/export/test_vllm_inference_mode.py
Giuseppe5: I think it started as a way to test both, but compile tests are flaky (especially on cpu) so I gave up.
I'll remove `compile` for now.

PR #1445, opened 2026-01-26
Fix (brevitas_examples/quantizers): correct stats for dynamic quants

Giuseppe5:
## Reason for this PR
AbsMinMax is commonly used for asymettric quantization, with AbsMax is more common for symmetric quantization.

## Changes Made in this PR

Set AbsMax for symmetric quant

### DIFF
--- a/src/brevitas_examples/common/generative/quantizers.py
+++ b/src/brevitas_examples/common/generative/quantizers.py
@@ -11,6 +11,7 @@ from brevitas.core.function_wrapper.shape import OverTensorView
 from brevitas.core.scaling.runtime import RuntimeDynamicGroupStatsScaling
 from brevitas.core.stats import AbsMinMax
 from brevitas.core.stats import NegativeMinOrZero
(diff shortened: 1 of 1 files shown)


src/brevitas_examples/common/generative/quantizers.py
```diff
@@ -72,7 +72,7 @@ class Int8DynamicActPerTensorFloat(DynamicActProxyMixin, Int8ActPerTensorFloat):
     """
     scaling_impl = RuntimeDynamicStatsScaling
     scaling_stats_input_view_shape_impl = OverTensorView
-    scaling_stats_op = 'min_max'
+    scaling_stats_op = 'max'
```
pablomlago: Can we use the enumeration? I understand that they are functionally equivalent atm, but it seems more robust.
`scaling_stats_op = StatsOp.MAX`

pablomlago (approved):
LGTM, consider openning an issue on potential PPL changes

PR #1446, opened 2026-01-26
Fix (brevitas_examples/llm): fix transformers tests

Giuseppe5:
## Reason for this PR

- Import file has changed
- Old FX interface has been deprecated
- Fix PPL change after quantizer update
- ~~Skip lm-eval tests until [this ](https://github.com/huggingface/datasets/issues/7965)is fixed~~
- Other compatibility errors

The first two are fixed, the third is a bit unpredictable.
For now I am pinning the transformer version until a broader compatibility across libraries is reached.

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/rotation_optimization.py
+++ b/src/brevitas_examples/llm/llm_quant/rotation_optimization.py
@@ -16,7 +16,12 @@ import torch.nn.functional as F
 import transformers
 from transformers import Trainer
 from transformers.data.data_collator import InputDataClass
-from transformers.tokenization_utils import PreTrainedTokenizerBase
+
+try:
+    from transformers.tokenization_utils import PreTrainedTokenizerBase
+except:
+    # This has changed in transformers v5
+    from transformers.tokenization_utils_base import PreTrainedTokenizerBase
 
 from brevitas.graph.calibrate import quantization_status_manager
 from brevitas.optim.cailey_sgd import CaileySGD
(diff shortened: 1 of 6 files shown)


tests/brevitas_examples/test_llm.py
```diff
@@ -432,6 +432,8 @@ def quantize_llm_assert_args(args, extra_args=None):
 def few_shot_eval_args(default_run_args, request):
     # Skip cases for which the LM evaluation library has not been installed
     for lib in request.param["imports"]:
+        if lib == "lm_eval":
```
pablomlago: Is the problem only in `winograde`? If so, we could remove that from the dictionary in the meantime, instead of skipping the whole test.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -16,7 +16,12 @@
 import transformers
 from transformers import Trainer
 from transformers.data.data_collator import InputDataClass
-from transformers.tokenization_utils import PreTrainedTokenizerBase
+
+try:
```
pablomlago, line 20: Is this needed with the pin to transformers?

tests/brevitas_examples/test_llm.py
Giuseppe5: I am confiding that this will be resolved soon, so this is faster than adjusting datasets and values etc.

If soon won't happen, then I agree and we can change datasets

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
Giuseppe5: No but it will be needed once we un-pin it, so might as well have it already.

Same it true for the removal of `transformers.utils.fx`. It still exists in <5. but not in >=5.

pablomlago (approved):
LGTM, consider opening an issue for unpinning transformers.

PR #1448, opened 2026-02-02
Feat (equalize): adding initial support for MixQuant

i-colbert:
## Reason for this PR

This PR extends graph equalization to enable [MixQuant](https://arxiv.org/pdf/2601.22347), which calibrates permutations to improve quantization accuracy when using block rotations. If calibrated intentionally (e.g., with mass diffusion), permutations can help balance the distribution of activation magnitudes across blocks prior to rotation. This is particularly beneficial for low-bit quantization (e.g., INT4, FP4) where outlier management is critical.

```
@article{sanjeet2026mixquant,
      title={MixQuant: Pushing the Limits of Block Rotations in Post-Training Quantization}, 
      author={Sai Sanjeet and Ian Colbert and Pablo Monteagudo-Lago and Giuseppe Franco and Yaman Umuroglu and Nicholas J. Fraser},
      year={2026},
      eprint={2601.22347},
      archivePrefix={arXiv},
      primaryClass={cs.LG},
      url={https://arxiv.org/abs/2601.22347}, 
}
```

## Changes Made in this PR

- __New Permutation Infrastructure__ (`src/brevitas/graph/permute.py`):

  - Added `GraphPermutationEqualization` class to manage permutation computation and application
  - Implemented 4 permutation strategies: `massdiff`, `zigzag`, `absmax`, and `random`
  - Created `rotate_permute_mode` context manager for unified rotation+permutation workflow

- __CLI Integration__ (`src/brevitas_examples/llm/llm_args.py`):

  - Added `--permute-fn` argument to select permutation strategy
  - Default behavior (i.e., `--permute-fn=None` ) does not apply permutations

- __Main Workflow Updates__ (`src/brevitas_examples/llm/main.py`):

  - Modified `fused_rotation_no_fx()` to support permutation mode
  - Integrated permutation computation during calibration phase
  - Added proper flow control for rotation+permutation combinations

- __Utility Functions__ (`src/brevitas/graph/utils.py`):

  - Added `find_node_for_module` helper function for graph traversal

__Context Manager Design:__ The `rotate_permute_mode` context manager encapsulates the entire workflow:

```python
with rotate_permute_mode(model, permute_fn='massdiff', ...) as rpm:
    # Calibration happens here to collect stats
    model(**calibration_data)
    # Permutations are computed and applied on exit
```

### Notable Implementation Details:

1. __Permutations are always fused__ (applied directly to weights), unlike rotations which can be parametrized for optimization
2. __Hook management__: Careful tracking of hooked modules to avoid duplicate hooks
3. __Device handling__: Activation statistics collected on CPU to avoid OOM issues
4. __SDPA region filtering__: SDPA (Scaled Dot-Product Attention) regions are excluded from permutation (at least for now) to avoid head alignment issues. Experiments suggest this is not critical.

### Expected Results

MixQuant demonstrates significant improvements over block rotations alone on Llama-3.2-1B-Instruct with W4A4 per-channel quantization. Using block rotations with `block_rotation_dim: 32` and `massdiff` permutation strategy, MixQuant achieves:

- __37% reduction__ in perplexity (27.1 → 17.0 on WikiText-2)
- __+2.2 pp improvement__ in average downstream task accuracy
- Consistent gains across all zero-shot tasks (ARC, HellaSwag, PIQA, Winogrande)

The improvements stem from better activation outlier management through channel permutations that balance magnitude distributions within rotation blocks.

| Method      | Wiki2 | ARC-C | ARC-E | HellaS | PIQA  | Wino  | Overall |
|-------------|-------|-------|-------|--------|-------|-------|---------|
| none        | 27.1  | 24.7% | 47.7% | 37.0%  | 65.7% | 52.3% | 45.5%   |
| absmax      | 20.3  | 25.3% | 46.4% | 37.6%  | 66.5% | 52.6% | 45.7%   |
| zigzag      | 17.2  | 25.6% | 51.4% | 38.6%  | 66.5% | 53.0% | 47.0%   |
| massdiff    | 17.0  | 27.6% | 51.3% | 38.7%  | 66.8% | 53.8% | 47.7%   |

**Configuration:** Both methods use [Qronos](https://arxiv.org/pdf/2505.11695) for error correction with dynamic per-row activations, MSE weight scales, and fused Hadamard rotations. See `src/brevitas_examples/papers/mixquant/llama3-mixquant-int4.yml` for the full config. You can run this as:
```shell
brevitas_ptq_llm --config=llama3-mixquant-int4.yml
```

Please use https://github.com/i-colbert/brevitas/tree/mixquant/src/brevitas_examples/papers/mixquant to reproduce the experiments from the paper.

## Testing Summary

Added `tests/brevitas/graph/test_permute.py`
- Tests with multiple permutation functions: `massdiff`, `zigzag`, `absmax`, and `random`
- Tests interop with `block_rotation_dim`, `disable_block_rotation_for_fused`, and `expansion_step`
- Tests CPU and GPU (if available)

## Risk Highlight

- [x] Documentation updates required in subsequent PR.

## Checklist

- [x] Initial CI/CD passing.


src/brevitas/graph/equalize.py
```diff
@@ -399,6 +406,177 @@ def apply_rewriters_accelerate(
...
+    def __init__(self, model, permute_fn='massdiff', **kwargs):
+        # rotate_permute_mode performs a specific transformation sequence:
+        # 1. Identify regions → 2. Apply permutations → 3. Apply rotations
+        # This sequencing requires delay_rewriters=True (defer rotation application)
+        # and return_rewriters=True (to access rotations for later application)
+        if not kwargs.pop('delay_rewriters', True) or not kwargs.pop('return_rewriters', True):
+            warnings.warn(
+                "delay_rewriters and return_rewriters must be True for rotate_permute_mode, ",
+                "overwriting provided value(s).")
+        kwargs['delay_rewriters'] = True
```
Giuseppe5: Considering that we can compute rotations without applying them, I'd try to decouple completely the GraphRotationEqualization class from the Permute class.

The main cost could be that we need to go through the region extract part of the algorithm twice, but it's something I'm willing to pay for decoupling.

This should also have implications in how we call this function in main.py, meaning that we should be able to avoid the if/else in fused_no_fx

src/brevitas_examples/llm/main.py
```diff
@@ -114,30 +116,60 @@ def fused_rotation_no_fx(model, calibration_loader, args):
...
-        disable_block_rotation_for_fused=args.disable_block_rotation_for_fused,
-        extra_state_kwargs={'scale_invariant_layers': rmsnorm_classes})
-    fx_model, rewriters = eq.apply(fx_model)
-
-    model = offload_model(model)
-    rewriters = fix_rewriter(rewriters, model, 'weight')
+    # When both rotation and permutation are enabled, use the unified context manager
+    if args.apply_permute:
+        print("Applying permutations...")
+        with rotate_permute_mode(
```
Giuseppe5: If we decouple rotation and permutations, we can first compute all rotations no matter what, and then do `rotate_permute_mode` maybe?
In that case, there should be the possibility pass the rotation rewriters to the class that it will just take care of applying them (rather than computing them)

src/brevitas/graph/equalize.py
```diff
@@ -230,6 +231,12 @@ def is_valid(self):
     def is_valid_activation_equalization(self) -> bool:
         return self.act_axis is not None
 
+    def apply_permute(self, indexes):
```
pablomlago: For me, `Region` should be a container to pass information between the region capture methods and the equalization ones. Therefore, I would try to avoid having equalization-specific methods in `Region`. I would move this logic into `_permute`, while making this a `@staticmethod` of `PermuteGraph`.

src/brevitas/graph/equalize.py
```diff
@@ -399,6 +406,177 @@ def apply_rewriters_accelerate(
     return model
 
 
+class rotate_permute_mode:
+
+    def __init__(self, model, permute_fn='massdiff', **kwargs):
```
pablomlago: Please add type hints in new/modified methods whenever possible.

src/brevitas/graph/equalize.py
```diff
@@ -464,6 +642,94 @@ def _channel_maxabs(inp: torch.Tensor, dim: int = 1) -> torch.Tensor:
...
+        block_norm[:, min_block] += torch.abs(x[:, i])
+        block_idxs[min_block].append(i)
+        # mark block as full
+        if (len(block_idxs[min_block]) == block_rotation_dim):
+            block_norm[:, min_block] = float('inf')
+    indexes = torch.tensor(block_idxs).flatten()
+    return indexes
+
+
+_permute_fn_map = {
```
pablomlago: For variables with module scope, I would rather use upper case, i.e.
```suggestion
_PERMUTE_FN_MAP = {
```

src/brevitas/graph/equalize.py
```diff
@@ -399,6 +406,177 @@ def apply_rewriters_accelerate(
...
+            warnings.warn(
+                "delay_rewriters and return_rewriters must be True for rotate_permute_mode, ",
+                "overwriting provided value(s).")
+        kwargs['delay_rewriters'] = True
+        kwargs['return_rewriters'] = True
+
+        self.rotation = GraphRotationEqualization(apply_permute=True, **kwargs)
+        self.model = model
+        self.rewriters = None
+        self.permute_fn = _permute_fn_map[permute_fn]
```
pablomlago: Consider, after replacing `GraphRotationEqualization` for `GraphPermutationEqualization` (or a similar name) to pass `permute_fn` to its constructor, to better encapsulate the permutation logic.

src/brevitas/graph/equalize.py
```diff
@@ -464,6 +642,94 @@ def _channel_maxabs(inp: torch.Tensor, dim: int = 1) -> torch.Tensor:
...
+
+_permute_fn_map = {
+    'zigzag': zigzag_permute,
+    'massdiff': massdiff_permute,
+    'absmax': absmax_permute,
+    'random': random_permute}
+
+
+@torch.no_grad()
+def _permute(region, list_of_act_val, block_rotation_dim, permute_fn, device):
```
pablomlago: I would do this method a `@staticmethod` of `PermuteGraph` (thus encapsulating most of the permutation logic in a single class), and, instead of calling `region.apply_permute`, remove that function from `Region`, as suggested above, and unpack that logic here.

src/brevitas/graph/equalize.py
```diff
@@ -464,6 +642,94 @@ def _channel_maxabs(inp: torch.Tensor, dim: int = 1) -> torch.Tensor:
     return out
 
 
+def _zigzag_sort(indexes, block_size):
```
pablomlago: Consider using a pattern like the one in [vLLM](https://github.com/vllm-project/vllm/blob/233b26ab354d5c96de29f98c17d5f2ed10773782/vllm/model_executor/layers/quantization/__init__.py#L50), to register the different permutation strategies.

Also, these methods do not need to be "private" (I would remove the underscore), and maybe I would extract them into a separate file, e.g. `permutation_equalize.py`, `permutation_utils.py`, to unclutter `equalizue.py`.

src/brevitas/graph/utils.py
```diff
@@ -92,6 +93,18 @@ def is_subseq(seq, subseq):
     return any(subseq == seq[i:len(subseq) + i] for i in range(len(seq) - len(subseq) + 1))
 
 
+def find_node_for_module(graph_model, target_module) -> Node:
```
pablomlago:
```suggestion
def find_node_for_module(graph_model, target_module) -> Optional[Node]:
```

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,354 @@
...
+from brevitas.graph.equalize import GraphRotationEqualization
+from brevitas.graph.equalize import Region
+from brevitas.graph.equalize import WalkRegionState
+from brevitas.graph.utils import find_node_for_module
+from brevitas.nn.equalized_layer import RotatedModule
+from brevitas.utils.logging import setup_logger
+
+logging = setup_logger(__name__)
+
+__all__ = ['GraphPermutationEqualization', 'rotate_permute_mode', 'get_permutation_method']
```
Giuseppe5: `get_permutation_method` does not to be in `__all__` if we are using it only internally

src/brevitas/graph/equalize.py
```diff
@@ -2128,6 +2145,9 @@ def apply(self,
                 added_regions += 1
         logging.debug(f"Adding {added_regions} sink-only regions")
 
+        # Store regions for potential use by GraphPermutationEqualization
```
Giuseppe5: The comment is too specific

src/brevitas/graph/equalize.py
```diff
@@ -2128,6 +2145,9 @@ def apply(self,
...
+        self.regions = regions
```
Giuseppe5: I'm not a huge fan of this, since we are "randomly" assigning an attribute in the middle of a function that does way too many things, but it's not "your fault" and probably we should revisit this class since it has grown too much.
I'm ok to leave this for now but maybe add a TODO to improve this

Giuseppe5: Probably all the logic above can be condensed in a method `get_region`.
We initialize `self.region` to None, and if `self.region` is None, then they are computed following this logic, and then initialize the attribute.
If `self.region` is not None, it just returns them.

`get_region` will also be used then in the context manager for permute instead of calling directly `self.region`

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,354 @@
...
+        for region in self.rotation.regions:
+            # Permutations are only applied to regions that use block rotations
+            apply_block_rotation = self.permutation.block_rotation_dim is not None
+            if self.rotation.disable_block_rotation_for_fused and (len(region.srcs) > 0):
+                apply_block_rotation = False
+            if apply_block_rotation:
+                # Check if block rotation is compatible with the current shape
+                if (region.max_shape_sinks // self.permutation.block_rotation_dim > 1) and \
+                    (region.max_shape_sinks % self.permutation.block_rotation_dim == 0):
+                    permute_regions.append(region)
```
Giuseppe5: This logic should probably live within GraphPermuteEqualization

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,354 @@
...
+            for name in region.sinks_names:
+                module = region.get_module_from_name(name)
+                if module not in self.hooked_modules:
+                    self.hooked_modules.add(module)
+                    hook_fn = partial(self.forward_stats_hook, name=name, batch_dim=batch_dim)
+                    h = module.register_forward_hook(hook_fn)
+                    self.hooks.append(h)
+
+    def extract_permute_regions(self, graph_model, regions):
+        state_impl_kwargs = {
```
Giuseppe5: Did you check if this class can benefit from `RegionWalkMixin`?
I'm happy to extend what the Mixin can do if it currently does not cover this class' use case

src/brevitas_examples/llm/main.py
```diff
@@ -114,30 +116,60 @@ def fused_rotation_no_fx(model, calibration_loader, args):
...
-        layers_to_expand=layers_to_expand,
-        block_rotation_dim=args.block_rotation_dim,
-        disable_block_rotation_for_fused=args.disable_block_rotation_for_fused,
-        extra_state_kwargs={'scale_invariant_layers': rmsnorm_classes})
-    fx_model, rewriters = eq.apply(fx_model)
-
-    model = offload_model(model)
-    rewriters = fix_rewriter(rewriters, model, 'weight')
+    # When both rotation and permutation are enabled, use the unified context manager
+    if args.apply_permute:
```
Giuseppe5: I'm not a huge fan of this massive if/else, especially since 95% of the arguments are shared between the two.
A quick solution would be to create only once a dict with all the arguments, and pass it like `**rotation_args`.

Maybe there's even a better way to do this.

In general, fused_no_fx is heading towards a refactoring because it is doing too many things (including calibration now)

src/brevitas_examples/llm/llm_args.py
```diff
@@ -392,6 +390,15 @@ def create_args_parser() -> ArgumentParser:
         default=0.5,
         type=float,
         help='If activation equalization is enabled, decide what alpha to use')
+    parser.add_argument(
+        '--apply-permute',
+        action='store_true',
+        help='Apply permutation equalization. Works with block rotations when both are enabled.')
+    parser.add_argument(
+        '--permute-fn',
+        choices=['absmax', 'massdiff', 'zigzag', 'random'],
```
Giuseppe5, line 395: Can't we fuse these two options?
```python
--permute-fn,
default = None
```

If None, do nothing.

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,354 @@
...
+        self.permutation.extract_permute_regions(model, permute_regions)
+        self.permutation.setup_permute()
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        self.permutation.apply_permute()
+        self.permutation.remove_hooks()
+
+
+class GraphPermutationEqualization:
```
Giuseppe5: For consistency, this should at least inherit from `GraphTransform` and maybe also from `RegionWalkMixin` (see the other comment for explaination)

Giuseppe5 (changes requested):

src/brevitas/graph/equalize.py
```diff
@@ -2060,6 +2080,8 @@ def find_sink(node):
             end_index = head_dim if head_dim != -1 else output_weight.shape[0]
             output_index = EqualizationIndexes(0, end_index, 0)
 
+            # NOTE: GraphPermutationEqualization.extract_permute_regions looks for these src and
```
Giuseppe5: We don't need this comment

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,387 @@
...
+        self.block_size = block_size
+        self.permute_fn = get_permutation_method(permute_fn)
+
+    def setup(self, graph_model: GraphModule, regions: List[Region]) -> GraphModule:
+        """Extract regions and setup hooks"""
+        self._extract_regions(graph_model, regions)
+        self._setup_hooks()
+        return graph_model
+
+    def forward_stats_hook(self, module, *args, name, batch_dim=0, **kwargs):
```
Giuseppe5, line 187: This function is already defined somewhere else.
We should try and avoid duplication as much as possible, if possible.

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def __init__(
+            self,
+            model: GraphModule,
+            rotation: GraphRotationEqualization,
+            block_size: int,
+            permute_fn: str = 'massdiff',
+            disable_for_fused_rotations: bool = False):
+
+        assert isinstance(rotation, GraphRotationEqualization), "Error: expected GraphPermutationEqualization instance"
```
Giuseppe5: This can also be None, right?

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,387 @@
...
+            if not self._is_compatible_region(region):
+                continue
+
+            # Create a new state for the online region
+            state = WalkRegionState(**self.full_state_kwargs)
+
+            # Add all sinks from the region to the state
+            for sink_name, sink_wrapper in region.sinks.items():
+                module = region.get_module_from_name(sink_name)
+                node = find_node_for_module(graph_model, module)
```
Giuseppe5, line 256: I am not a fan of this dance between nodes and modules. Probably we should just extend the region (including `find_srcs` and `find_sinks` to simply keep track of the node as well, so that you can also do something like `node = region.get_node_from_name(...)`.

src/brevitas_examples/llm/llm_args.py
Giuseppe5: This was marked as resolved but nothing has changed?
I assume you don't want to do it but why?

src/brevitas_examples/papers/mixquant/README.md
```diff
@@ -0,0 +1,22 @@
+# MixQuant: Pushing the Limits of Block Rotations in Post-Training Quantization
+
+📄 [Paper](https://arxiv.org/pdf/2601.22347)
```
Giuseppe5, line 3: Before we merge, I'd like to make sure results are consistent

src/brevitas_examples/llm/llm_args.py
i-colbert: The suggested was implemented... the default was changed to None and apply-permute was removed

src/brevitas_examples/papers/mixquant/README.md
i-colbert: I am saving that for the end.

src/brevitas/graph/permute.py
i-colbert: It is slightly different though, we also save the device here.

src/brevitas/graph/equalize.py
i-colbert: In case of a future refactor, this comment provides a reminder that it has a connection to `GraphPermutationEqualization` that is not obvious. The intention is to be helpful.

src/brevitas/graph/permute.py
```diff
@@ -0,0 +1,387 @@
...
+logging = setup_logger(__name__)
+
+__all__ = ['GraphPermutationEqualization', 'rotate_permute_mode']
+
+# Initialize permutation-invariant layers from scale-invariant layers
+_permute_invariant_layers = list(_scale_invariant_layers)
+_permute_invariant_layers.extend([torch.nn.GELU, torch.nn.SELU])
+
+# Try to add HuggingFace activations
+try:
```
i-colbert: Pass these in from the LLM entrypoint as extra_state_kwargs.

src/brevitas/graph/permute.py
i-colbert: This context manager assumes rotations will be applied. If it is None, then it should throw an error. Will add that now.

src/brevitas/graph/permute.py
i-colbert: Pushed shared code into `GraphTransform`.

PR #1452, opened 2026-02-09
Feat (stats): Statistics capture

pablomlago:
## Reason for this PR

Add opt-in statistics collection for GPTQ/PTQ to quantify and debug per-layer quantization error during calibration, without changing default behavior.

### What
- Introduced a generic statistics collection framework (`stats_utils.py`) with a no-op default.
- Instrumented GPTQ to emit `pre_update` and `post_update` events, to compute statistics before/after weight update.
- Added GPTQ error metrics (relative weight error and Hessian-weighted output error).
- Added `--ptq-stats` CLI flag to the LLM entrypoint.

### How
- Used a `ContextVar`-based collector to minimize changes required for statistics collection.
- Decoupled *when* stats are emitted (events) from *what* is computed (handlers).

## Testing Summary

- Manually validated GPTQ runs with and without `--ptq-stats gptq`.

## Risk Highlight

## Checklist

- [ ] Initial CI/CD passing.

### DIFF
--- a/src/brevitas/graph/gptq.py
+++ b/src/brevitas/graph/gptq.py
@@ -212,6 +212,7 @@ class GPTQ(GPxQ):
 
         # Log post-update statistics
         gptq_stats_collector.log("post_update", name=self.name, layer=self.layer, H=H_orig)
+        del H_orig
         # If stats are being collected, print statistics to DEBUG
         if gptq_stats_collector.is_active:
             logger.debug(
--- a/src/brevitas/graph/gptq.py
+++ b/src/brevitas/graph/gptq.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from copy import deepcopy
+import functools
 import math
 from typing import List
 from typing import Optional
@@ -9,6 +10,8 @@ import warnings
 
 from packaging import version
 import torch
+
+from brevitas.utils.stats_utils import StatsCollectorCtx
 
 try:
     from torch.linalg import LinAlgError
@@ -21,6 +24,23 @@ from brevitas.graph.gpxq import gpxq_mode
 from brevitas.graph.gpxq import SUPPORTED_CONV_OP
 from brevitas.graph.utils import is_conv_transposed
 from brevitas.utils.torch_utils import StopFwdException
+
+
+# TODO (pml): Simplify this function
+@torch.no_grad()
+def compute_weight_error(prefix: str, name: str, layer: torch.nn.Module, H: torch.Tensor):
+    quant_weight = layer.quant_weight().value.to(dtype=H.dtype)
+    weight = layer.weight_orig.to(dtype=H.dtype, device=quant_weight.device)
+    H = H.to(device=quant_weight.device).squeeze(0)
+    # Compute relative error between quantized and original weights
+    err = quant_weight - weight
+    weight_rel_err = torch.norm(err) / torch.norm(weight)
+    # Compute relative error weighted by the Hessian, i.e. (w-q)^T H (w-q) / w^T H w
+    out_rel_err = torch.norm(err.T @ H @ err, p='fro') / torch.norm(weight.T @ H @ weight, p='fro')
+    return {
+        name: {
+            f"{prefix}_rel_weight_err": weight_rel_err.item(),
+            f"{prefix}_rel_out_err": out_rel_err.item(),}}
 
 
 class GPTQ(GPxQ):
@@ -112,6 +132,10 @@ class GPTQ(GPxQ):
         weight = self.layer.weight.data
         dev = weight.device
 
+        # Log pre-update statistics
+        gptq_stats_collector = StatsCollectorCtx.get()
+        gptq_stats_collector.log("pre_update", name=self.name, layer=self.layer, H=self.H)
+
         # Store the original dtype of the weights
         # During computation, everything is converted to float32.
         # When the weights are updated, we cast everything back to the original dtype
@@ -167,7 +191,9 @@ class GPTQ(GPxQ):
                 f'Increasing the number of samples might fix this issue')
             return
         finally:
-            del self.H
+            # TODO (pml): Delay Hessian deletion for statistics capture
+            # del self.H
+            pass
 
         for i1 in range(0, self.columns, self.blocksize):
             i2 = min(i1 + self.blocksize, self.columns)
@@ -195,6 +221,10 @@ class GPTQ(GPxQ):
                 weight[group_index, :, perm[i2:]] -= (
                     error_block[group_index].matmul(h_inv[group_index, i1:i2,
                                                           i2:].to(dev))).to(dtype)
+
+        # Log post-update statistics
+        gptq_stats_collector.log("post_update", name=self.name, layer=self.layer, H=self.H)
+
         if hasattr(self.layer, 'offload_params'):
             self.layer.offload_params(self.layer)
 
@@ -260,6 +290,12 @@ class gptq_mode(gpxq_mode):
         # How many subblock to use during GPTQ for each layer
         self.num_blocks = num_blocks
         self.gptq_class = gptq_class
+
+        # Register the statistics to collect during GPTQ
+        gptq_stats_collector = StatsCollectorCtx.get()
+        gptq_stats_collector.on("pre_update", functools.partial(compute_weight_error, prefix="pre"))
+        gptq_stats_collector.on(
+            "post_update", functools.partial(compute_weight_error, prefix="post"))
 
     def catch_stopfwd(self, *args, **kwargs):
         try:
--- a/src/brevitas/graph/gptq.py
+++ b/src/brevitas/graph/gptq.py
@@ -11,6 +11,7 @@ import warnings
 from packaging import version
 import torch
 
+from brevitas.utils.logging import setup_logger
 from brevitas.utils.stats_utils import StatsCollectorCtx
 
 try:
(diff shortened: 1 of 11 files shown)


src/brevitas_examples/llm/main.py
```diff
@@ -607,16 +609,19 @@ def quantize_llm(args, extra_args=None):
...
-                max_accumulator_tile_size=args.gpxq_max_accumulator_tile_size)
+            stats_collector_ctx = collect_stats(
+                DictStatsCollector()) if "gptq" in args.ptq_stats else nullcontext()
+            with stats_collector_ctx:
+                apply_gptq(
+                    model,
+                    calibration_loader,
+                    act_order=args.gpxq_act_order,
+                    use_quant_activations=args.gpxq_use_quant_activations,
+                    create_weight_orig=not args.disable_create_weight_orig,
```
JP-Amboage, line 620: Given the current code, I think this has to be true anytime the stats are being logged. I'd suggest adding a validation for that.

src/brevitas/utils/stats_utils.py
```diff
@@ -0,0 +1,87 @@
...
+class BaseStatsCollector(ABC):
+
+    def __init__(self) -> None:
+        self._stats_fn: Dict[str, StatFn] = {}
+
+    def on(self, key: str, fn: StatFn) -> None:
+        self._stats_fn[key] = fn
+
+    def log(self, event: str, **payload) -> None:
+        if (fn := self._stats_fn.get(event, None)) is not None:
```
JP-Amboage: Very minor thing, but `None`  is already the default value in `.get` so this could just be `self._stats_fn.get(event)`. Also, since the values of `self._stats_fn` are always `StatFn`, I think we can avoid the `is not None`. And just rewrite the whole line as `if (fn := self._stats_fn.get(event))`. But this is just a matter of "style" so if you find the current one to be more readable or prefer it for any other reason I'm fine with that too.

src/brevitas/graph/utils.py
```diff
@@ -186,3 +186,22 @@ def remove_weight_orig(model: nn.Module):
     for name, module in model.named_modules():
         if hasattr(module, 'weight_orig'):
             del module.weight_orig
+
+
+@torch.no_grad()
+def gpxq_compute_error_stats(prefix: str, name: str, layer: torch.nn.Module,
+                             H: torch.Tensor) -> Dict[str, float]:
+    # This computation only supports nn.Linear, currently
+    assert isinstance(layer, qnn.QuantLinear), "`compute_weight_error` only supports `QuantLinear` layers"
```
JP-Amboage: Maybe, following our private discussion, replace `assert` for `Exceptions/Errors`

src/brevitas/graph/utils.py
```diff
@@ -186,3 +186,22 @@ def remove_weight_orig(model: nn.Module):
...
+    quant_weight = layer.quant_weight().value.to(dtype=H.dtype)
+    weight = layer.weight_orig.to(dtype=H.dtype, device=quant_weight.device)
+    H = H.to(device=quant_weight.device).squeeze(0)
```
JP-Amboage: perhaps the device could be a parameter to accommodate for cases where memory is an issue. For example, in cases where it is possible to run the whole algorithm (GPTQ/Qronos) in GPU memory, but there's no space for these extra weights or H matrix in GPU. In the current approach, one would have to switch the whole thing to CPU, however it could be better to still run the algorithm in GPU and just the stats in CPU.

src/brevitas/graph/gptq.py
```diff
@@ -112,6 +119,13 @@ def single_layer_update(self, percdamp=.01, c=1e4):
         weight = self.layer.weight.data
         dev = weight.device
 
+        gptq_stats_collector = StatsCollectorCtx.get()
+        H_orig = None
+        if gptq_stats_collector.is_active:
+            H_orig = deepcopy(self.H)
```
JP-Amboage: As I also mentioned in stats_utils.py, perhaps allow for using an specific device for the copy. I explained why I think this in the other comment (stats_utils.py).

src/brevitas/graph/utils.py
```diff
@@ -186,3 +186,22 @@ def remove_weight_orig(model: nn.Module):
     for name, module in model.named_modules():
         if hasattr(module, 'weight_orig'):
             del module.weight_orig
+
+
+@torch.no_grad()
+def gpxq_compute_error_stats(prefix: str, name: str, layer: torch.nn.Module,
```
JP-Amboage: Since this is for `gpxq` It may be good to have also an argument to receive `G` (the "cross correlation" matrix in Qronos) defaulting to None. Then inside the function decide which loss to compute based on `G` being None or not. This would allow this function to be directly applicable in Qronos to compute stats about the Qronos objective.

src/brevitas/graph/gptq.py
```diff
@@ -261,6 +284,13 @@ def __init__(
         self.num_blocks = num_blocks
         self.gptq_class = gptq_class
 
+        # Register the statistics to collect during GPTQ
+        gptq_stats_collector = StatsCollectorCtx.get()
+        gptq_stats_collector.on(
+            "pre_update", functools.partial(gpxq_compute_error_stats, prefix="pre"))
+        gptq_stats_collector.on(
+            "post_update", functools.partial(gpxq_compute_error_stats, prefix="post"))
```
JP-Amboage: I wonder if this could be in the parent GPXQ class to avoid duplicating code when extending this functionality to Qronos etc.

src/brevitas/graph/gptq.py
```diff
@@ -195,6 +209,15 @@ def single_layer_update(self, percdamp=.01, c=1e4):
                 weight[group_index, :, perm[i2:]] -= (
                     error_block[group_index].matmul(h_inv[group_index, i1:i2,
                                                           i2:].to(dev))).to(dtype)
+
+        # Log post-update statistics
+        gptq_stats_collector.log("post_update", name=self.name, layer=self.layer, H=H_orig)
```
JP-Amboage: I wonder if this logic (and other logging related logic) could be encapsulated in a function/method and handled through the parent class so that the functionality extends directly to other GPXQ methods.

src/brevitas_examples/llm/llm_args.py
```diff
@@ -486,6 +486,13 @@ def create_args_parser() -> ArgumentParser:
         default=1,
         help='Batch size for calibration data loader. (default: %(default)s).')
 
+    parser.add_argument(
+        '--ptq-stats',
```
JP-Amboage, line 490: Is the use of PTQ here intentional (to keep it more general)? Or should it be GPXQ?

src/brevitas/graph/gpxq.py
```diff
@@ -110,6 +117,13 @@ def __init__(
         else:
             self.model.forward = self.catch_stopfwd
 
+        # Register the statistics to collect during GPTQ
+        gptq_stats_collector = StatsCollectorCtx.get()
+        gptq_stats_collector.on(
+            "pre_update", functools.partial(gpxq_compute_error_stats, prefix="pre"))
+        gptq_stats_collector.on(
+            "post_update", functools.partial(gpxq_compute_error_stats, prefix="post"))
```
JP-Amboage, line 125: perhaps `gpxq_stats_collector` is a better name than `gptq_stats_collector` now that it's shared among GPXQ algorithms.

src/brevitas/graph/utils.py
```diff
@@ -189,16 +191,26 @@ def remove_weight_orig(model: nn.Module):
...
 @torch.no_grad()
-def gpxq_compute_error_stats(prefix: str, name: str, layer: torch.nn.Module,
-                             H: torch.Tensor) -> Dict[str, float]:
+def gpxq_compute_error_stats(
+        prefix: str,
+        name: str,
+        layer: torch.nn.Module,
+        H: torch.Tensor,
+        G: Optional[torch.Tensor] = None,
+        device: Optional[Union[str, torch.device]] = None) -> Dict[str, float]:
```
JP-Amboage, line 200: shall we have a new argument for the llm quantization script like `--gpxq-stats-device` to manage this? I'm not sure if it's better to add that now or to just leave it with the default that you wrote, and let anyone add it in the future if they actually need it.

JP-Amboage (commented):
I like the new approach as a solution to get the stats functionality to work in all gpxq algorithms without any extra code 🚀!  I just left two comments, one about a variable name, and another about the device handling. I'm closing the comments from the previous review that have been addressed. Some remain open.

src/brevitas/graph/utils.py
JP-Amboage: Addressed in the next commit.

src/brevitas/graph/utils.py
JP-Amboage: Addressed in the next commit

src/brevitas/graph/gptq.py
JP-Amboage: Addressed in the next commit

src/brevitas/graph/utils.py
JP-Amboage: Addressed in the next commit

src/brevitas/graph/gptq.py
JP-Amboage: Addressed in the next commit

src/brevitas_examples/llm/main.py
pablomlago: Added the following check under the `__init__` of `gpxq_mode`:
```
# Verify that the original weights are kept when statistics are being collected
if gpxq_stats_collector.is_active and not create_weight_orig:
    raise ValueError(
        "Statistics collection during GPxQ requires keeping original weights. Set `create_weight_orig` to True."
    )
```

src/brevitas_examples/llm/llm_args.py
pablomlago: Yes, I want to keep it to general, as this is likely to be reused for other PTQ algorithms.

src/brevitas/graph/utils.py
pablomlago: Definitely computing the statistics will increase the peak memory usage during GPxQ, but since I am expecting the statistics collection to be used sporadically, I'd rather delay adding a new CLI argument, until it is strictly needed (e.g. when hitting an OOM for a model that we need to collect debugging statistics on while running GPxQ). In the meantime, I opted for providing the minimal functionality to propagate `device` to the quantization error computation.

src/brevitas_examples/llm/main.py
JP-Amboage: Following our private discussion, this check should be performed in `src/brevitas_examples/llm
/llm_args.py`

PR #1455, opened 2026-02-12
Setup: test ONNX dynamo export with torch >= 2.7

Giuseppe5:
## Reason for this PR

New version of onnx_ir seems to not be compatible with our pinned version of onnxscript.

## Changes Made in this PR

Test ONNX dynamo export only against torch 2.7 and higher, where there are no issues.
The other solution is to pin both onnxscript _and_ onnx_ir, which is not practical.

### DIFF
--- a/requirements/requirements-export.txt
+++ b/requirements/requirements-export.txt
@@ -1,3 +1,4 @@
 onnx==1.17.0
 onnxoptimizer
+onnxscript
 qonnx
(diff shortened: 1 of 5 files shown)


nickfraser (approved):
Did you try unpinning onnxscript (in case they fixed what was broken before?)

If so, LGTM!

PR #1456, opened 2026-02-12
Feat (common/quantizer): Enable passing custom quantizers

pablomlago:
## Reason for this PR

Currently, every time we need to support a new quantization type, we need to modify our entry-point in several ways (e.g., new args, new option in the dict), and this process does not scale up with more advanced quantization schemes.

## Changes made in this PR

- Added a generic `Registry` helper to `python_utils.py` with a decorator-based API.
- Created base class for custom quantizers (`BaseQuantizer`).
- Added new `custom-quantizer` argument to the LLM entrypoint.
- Added example demonstrating custom quantizer registration using the new registry.

## Testing Summary

- Added new test in `test_llm_cases.py`.

## Risk Highlight

### DIFF
--- a/src/brevitas_examples/llm/llm_args.py
+++ b/src/brevitas_examples/llm/llm_args.py
@@ -21,6 +21,13 @@ def create_args_parser() -> ArgumentParser:
         default="facebook/opt-125m",
         help='HF model name. Default: facebook/opt-125m.')
     parser.add_argument(
+        '--custom-quantizer',
+        type=str,
+        default=None,
+        help=
+        'Override the quantization list with custom user defined quantizers. This must be a .py file with a list of seven quantizers. Default: None.'
+    )
+    parser.add_argument(
         '--dtype',
         type=str,
         default="auto",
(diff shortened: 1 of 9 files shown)


src/brevitas_examples/llm/llm_args.py
```diff
@@ -20,6 +20,13 @@ def create_args_parser() -> ArgumentParser:
         type=str,
         default="facebook/opt-125m",
         help='HF model name. Default: facebook/opt-125m.')
+    parser.add_argument(
+        '--custom-quantizer',
+        type=str,
+        default=None,
+        help=
+        'Override the quantization list with custom user defined quantizers. This must be a .py file with a list of seven quantizers. Default: None.'
```
Giuseppe5, line 28: We need to update this. I'd add a small example on how to call this with/without external file.

Giuseppe5 (approved):
Small change, then we can merge

src/brevitas_examples/llm/llm_args.py
pablomlago: Added a test case for this use-case.

PR #1457, opened 2026-02-18
Feat (bit_width): generalize restrict_impl with clamping

Giuseppe5:
## Reason for this PR

Currently there is no way of clamping the bit-width when it is a learned parameter

## Changes Made in this PR

Change the restrict_impl to also include clamping.
Also there is also a clamping class now that accounts for max possible value.

## Testing Summary

Added

### DIFF
--- a/src/brevitas/core/function_wrapper/ops_ste.py
+++ b/src/brevitas/core/function_wrapper/ops_ste.py
@@ -5,6 +5,8 @@
 ScriptModule wrappers of various functions defined in :obj:`~brevitas.function.ops_ste`.
 """
 
+from typing import Optional
+
 import torch
 
 import brevitas
@@ -90,6 +92,29 @@ class ScalarClampMinSte(brevitas.jit.ScriptModule):
     @brevitas.jit.script_method
     def forward(self, x: torch.Tensor):
         return scalar_clamp_min_ste(x, self.min_val)
+
+
+class ScalarSignedClampSte(brevitas.jit.ScriptModule):
+    """
+    ScriptModule wrapper for :func:`~brevitas.function.ops_ste.scalar_clamp_ste`.
+    """
+
+    __constants__ = ['min_val', 'max_val']
+
+    def __init__(self, min_val: float, max_val: float) -> None:
+        super(ScalarSignedClampSte, self).__init__()
+        # Verify that the minimum value is set to a non-zero value, as when min_val == 0.0,
+        # this module implements the identity but the gradient returned at x = 0.0, is zero,
+        # instead of 1.
+        assert abs(min_val) > 0.0, "min_val has to be greater than zero."
+        assert abs(max_val) > 0.0, "max_val has to be greater than zero."
+        self.min_val = abs(min_val)
+        self.max_val = abs(max_val)
+
+    @brevitas.jit.script_method
+    def forward(self, x: torch.Tensor):
+        return torch.where(x >= 0, 1., -1.).type_as(x) * scalar_clamp_ste(
+            abs_binary_sign_grad(x), self.min_val, self.max_val)
 
 
 class ScalarSignedClampMinSte(brevitas.jit.ScriptModule):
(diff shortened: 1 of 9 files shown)


src/brevitas/core/function_wrapper/ops_ste.py
```diff
@@ -92,6 +94,26 @@ def forward(self, x: torch.Tensor):
...
+class ScalarClampSte(brevitas.jit.ScriptModule):
+    """
+    ScriptModule wrapper for :func:`~brevitas.function.ops_ste.scalar_clamp_ste`.
+    """
+
+    __constants__ = ['min_val', 'max_val']
+
+    def __init__(self, min_val: float, max_val: float) -> None:
+        super(ScalarClampSte, self).__init__()
+        # Verify that the minimum value is set to a non-zero value, as when min_val == 0.0,
```
pablomlago: Comment is outdated.

src/brevitas/core/function_wrapper/ops_ste.py
```diff
@@ -92,6 +94,26 @@ def forward(self, x: torch.Tensor):
...
+        # this module implements the identity but the gradient returned at x = 0.0, is zero,
+        # instead of 1.
+        self.min_val = abs(min_val)
+        self.max_val = abs(max_val)
+
+    @brevitas.jit.script_method
+    def forward(self, x: torch.Tensor):
+        return scalar_clamp_ste(abs_binary_sign_grad(x), self.min_val, self.max_val)
```
pablomlago: Is `abs_binary_sign_grad` needed here? I would remove if possible, as the absolute value can be applied as part of `restrict_scaling_impl`.

src/brevitas/core/restrict_val.py
```diff
@@ -29,13 +30,19 @@ class _RestrictClampValue(brevitas.jit.ScriptModule):
...
-            scaling_min_val: Optional[float] = None,
+            min_val: Optional[float] = None,
+            max_val: Optional[float] = None,
             restrict_value_impl: Optional[Module] = None):
         super(_RestrictClampValue, self).__init__()
-        if scaling_min_val is not None and scaling_min_val != 0:
-            self.clamp_min_ste = ScalarSignedClampMinSte(scaling_min_val)
+        if min_val is not None:
+            if max_val is not None:
+                # When min_val and max_val are defined, we don't need to have a signed version of
```
pablomlago, line 42: Maybe it is worth having a more explicit comment, on how this module behaves. When only `min_val` is passed, this module ensures that the output does not fall in the interval `(-min_val, min_val)` (i.e. values within (-min_val, 0] are assigned to -min_val, and the symmetric behaviour in [0, min_val)). However, when `max_val` is also passed, the clamping projects the values to `[min_val, max_val]`.

src/brevitas/core/function_wrapper/ops_ste.py
```diff
@@ -92,6 +94,26 @@ def forward(self, x: torch.Tensor):
...
+    """
+
+    __constants__ = ['min_val', 'max_val']
+
+    def __init__(self, min_val: float, max_val: float) -> None:
+        super(ScalarClampSte, self).__init__()
+        # Verify that the minimum value is set to a non-zero value, as when min_val == 0.0,
+        # this module implements the identity but the gradient returned at x = 0.0, is zero,
+        # instead of 1.
+        self.min_val = abs(min_val)
```
pablomlago: I do not see why the absolute value is needed for `min_val` and `max_val`.

tests/brevitas/core/test_bit_width.py
```diff
@@ -153,3 +156,59 @@ def test_load_from_stateful_const(
             bit_width_stateful_const_tensor = bit_width_stateful_const()
             assert bit_width_stateful_const_tensor == bit_width_init_two
             assert bit_width_stateful_const_tensor == bit_width_parameter_tensor
+
+    def test_bit_width_offset_clamping_min_max(self):
```
pablomlago: This test is actually testing multiple things. Isn't it possible to divide into single tests? It is not a big deal, just mention it in case it is easy to split it.

PR #1461, opened 2026-02-24
Feat (brevitas_examples/llm): better dataset handling

Giuseppe5:
## Reason for this PR

Dataset management is not ideal, with iterator being explicitly converted to lists for no apparent reason.

This should take care of that, although it would be worth taking a more general look in the future.

We need to add 'labels' token everywhere, to avoid a potential miss-match between the forward signature in training/eval mode after the model has gone through FX tracing.

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/data.py
+++ b/src/brevitas_examples/llm/llm_quant/data.py
@@ -173,30 +173,27 @@ def get_wikitext2(
     # Add BOS token to each sequence if add_bos_token is True and the tokenizer supports this token
     if add_bos_token and tokenizer.bos_token_id is not None:
         seqlen = seqlen - 1
-        sequence_process_fn = lambda inp: torch.cat([
-            torch.tensor([[tokenizer.bos_token_id]], dtype=inp.dtype, device=inp.device), inp],
-                                                    dim=1)
+        sequence_process_fn = sequence_process_fn = lambda inp: [tokenizer.bos_token_id] + inp
     else:
         # Identity, the BOS token is not added
         sequence_process_fn = lambda inp: inp
 
-    data = tokenizer("\n\n".join(raw_dataset['text']), return_tensors='pt')
-    dataloader = []
+    input_ids = tokenizer(
+        "\n\n".join(raw_dataset['text']), return_attention_mask=False)["input_ids"]
+    tokenized_data = []
     if split == 'train':
         for _ in tqdm(range(nsamples)):
-            i = random.randint(0, data.input_ids.shape[1] - seqlen - 1)
+            i = random.randint(0, input_ids.shape[1] - seqlen - 1)
             j = i + seqlen
-            inp = sequence_process_fn(data.input_ids[:, i:j])
+            inp = sequence_process_fn(input_ids[:, i:j])
             attention_mask = torch.ones_like(inp)
-            dataloader.append({'input_ids': inp, 'attention_mask': attention_mask, 'labels': inp})
+            tokenized_data.append({'input_ids': inp})
     elif split in ['test', 'validation']:
-        nsamples = data['input_ids'].numel() // seqlen
+        nsamples = input_ids.numel() // seqlen
         for i in tqdm(range(nsamples)):
-            batch = sequence_process_fn(data['input_ids'][:, (i * seqlen):((i + 1) * seqlen)])
-            attention_mask = torch.ones_like(batch)
-            dataloader.append({
-                'input_ids': batch, 'attention_mask': attention_mask, 'labels': batch})
-    return dataloader
+            batch = sequence_process_fn(input_ids[:, (i * seqlen):((i + 1) * seqlen)])
+            tokenized_data.append({'input_ids': inp})
+    return Dataset.from_list(tokenized_data)
 
 
 def load_raw_dataset(dataset_name: str, split: str, seed: int = 42) -> Dataset:
--- a/src/brevitas_examples/llm/llm_quant/data.py
+++ b/src/brevitas_examples/llm/llm_quant/data.py
@@ -101,12 +101,6 @@ def tokenize_and_group_texts(
     )
 
 
-def _clm_dataset_to_list(row: np.ndarray,) -> Dict[str, torch.Tensor]:
-    input_ids = torch.tensor(row["input_ids"], dtype=torch.int64).unsqueeze(0)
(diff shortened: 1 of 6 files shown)


src/brevitas_examples/llm/llm_quant/data.py
pablomlago:
```suggestion
) -> datasets.Dataset:
```

src/brevitas_examples/llm/llm_quant/data.py
```diff
@@ -179,29 +172,27 @@ def get_wikitext2(
     # Add BOS token to each sequence if add_bos_token is True and the tokenizer supports this token
     if add_bos_token and tokenizer.bos_token_id is not None:
         seqlen = seqlen - 1
-        sequence_process_fn = lambda inp: torch.cat([
-            torch.tensor([[tokenizer.bos_token_id]], dtype=inp.dtype, device=inp.device), inp],
-                                                    dim=1)
+        sequence_process_fn = sequence_process_fn = lambda inp: [tokenizer.bos_token_id] + inp
```
pablomlago: Should this be?
```suggestion
        sequence_process_fn = lambda inp: [tokenizer.bos_token_id] + inp
```

src/brevitas_examples/llm/llm_quant/data.py
pablomlago:
```suggestion
        seed: int = 42) -> datasets.Dataset:
```

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -44,22 +45,59 @@
 from brevitas_examples.llm.llm_quant.data import load_raw_dataset
 
 
-class DatasetToDevice(torch.utils.data.Dataset):
+def collate_batch(examples: List[Dict[str, List[np.ndarray]]],
+                  sequence_length) -> Dict[str, torch.Tensor]:
```
pablomlago:
```suggestion
                  sequence_length: int) -> Dict[str, torch.Tensor]:
```

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -44,22 +45,59 @@
...
+    # Make sure we load only what's necessary, ie we only load a `input_ids` column.
+    assert all(list(example.keys()) == ["input_ids"] for example in examples)
```
pablomlago: Should we remove this? Maybe it is worth turning into an `if ... raise ValueError(....)`.

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -44,22 +45,59 @@
...
 
-    def __init__(self, data: List, device: Optional[Union[str, torch.device]]):
-        super().__init__()
-        self.data = data
-        self.device = device
+    input_ids = np.vstack([examples[i]["input_ids"] for i in range(len(examples))])  # (b, s)
```
pablomlago, line 54: This is a `numpy.ndarray` still? Shouldn't it be converted to `torch.Tensor`? I think `torch.vstack` should do the trick.

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -44,22 +45,59 @@
...
 
-    def __getitem__(self, idx):
-        if self.device is not None:
-            return {
-                name: recursive_to_device(val, self.device) for name, val in self.data[idx].items()}
-        else:
-            return self.data[idx]
+    result: Dict[str, np.ndarray] = {}
```
pablomlago:
```suggestion
    result: Dict[str, torch.Tensor] = {}
```

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -44,22 +45,59 @@
...
-    def __len__(self):
-        return len(self.data)
+        num_heads = normalized_config.num_attention_heads
+        if hasattr(normalized_config, "num_key_value_heads"):
+            num_kv_heads = normalized_config.num_key_value_heads
+        else:
+            num_kv_heads = num_heads
+        head_dim = normalized_config.hidden_size // num_heads
+        return partial(collate_batch_fx, num_kv_heads=num_kv_heads, head_dim=head_dim)
+    else:
```
pablomlago: No need to have `else`, as the `if` has a return statement. This could just be:
```
if require_fx:
     return partial(...)
return partial(...)
```

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -81,19 +119,17 @@ def recursive_to_device(tensor_or_iterable: Union[Iterable, torch.Tensor], devic
...
+        dataset_name: str,
+        tokenizer: Any,
+        nsamples: int = 128,
+        seqlen: int = 2048,
+        seed: int = 0,
+        split: str = "train",
+        bos_preprocessing: Optional[str] = None,
+        add_eos_token: bool = False,
+        fuse_documents: bool = True,
+        require_fx: bool = False):
```
pablomlago:
```suggestion
        require_fx: bool = False) -> datasets.Dataset:
```

pablomlago (changes requested):
Some minor comments

src/brevitas_examples/llm/llm_quant/data_utils.py
Giuseppe5: I added a conversion to torch.tensor after the `np.vstack`. Unless torch.vstack is orders of magnitude faster than np.vstack, I'd say it's fine

src/brevitas_examples/llm/llm_quant/data_utils.py
```diff
@@ -81,19 +118,15 @@ def recursive_to_device(tensor_or_iterable: Union[Iterable, torch.Tensor], devic
...
-):
+        dataset_name: str,
+        tokenizer: Any,
+        nsamples: int = 128,
+        seqlen: int = 2048,
+        seed: int = 0,
+        split: str = "train",
+        bos_preprocessing: Optional[str] = None,
+        add_eos_token: bool = False,
+        fuse_documents: bool = True):
```
pablomlago:
```suggestion
        fuse_documents: bool = True) -> datasets.Dataset:
```

pablomlago (approved):
LGTM (assuming test pass)

PR #1467, opened 2026-02-27
Feat (brevitas_examples/llm): deprecate unused MLIR export

Giuseppe5:
## Reason for this PR

Residue of a forgotten past.

We used to support a custom export flow to MLIR.
This flow only supported a specific types of model, to a specific target, and it is highly likely that it depends on specific torch/transformers versions.

This flow has not been tested in a while and it is not currently being used anymore, so it is safe to deprecate.

## Changes Made in this PR

Cleanup


nickfraser (approved):
LGTM - we can always revert if this becomes useful again.

PR #1470, opened 2026-02-27
Docs: documentation for MixQuant paper

i-colbert:
## Reason for this PR

Documentation page for MixQuant paper: https://arxiv.org/pdf/2601.22347

It's a draft because tests do not need to run, so we can save that time and resources.

block_size_activation_plots

transformer_block_with_permutations

correlation_quant_error_bound_all_models


docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,324 @@
...
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 3
+
+
+About the Algorithm
+-------------------
+
+In few-bit PTQ, activation outliers inflate the dynamic range, increasing the resolution of the quantizer and its
```
pablomlago: Is it `increasing` or `decreasing` the resolution? The bigger the outliers, the greater the gaps in the quantization grid, so intuitively, I would say that its resolution decreases.

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,324 @@
...
+
+In Brevitas, the default permutation strategy is MassDiff, the greedy calibration algorithm
+described in Algorithm 1 of the paper. Given a calibration dataset, MassDiff:
+
+1. Computes an average magnitude score for each channel
+2. Processes channels in descending order of score
+3. Assigns each channel to the block whose accumulated :math:`\ell_1` mass would increase the least
+4. Continues until all blocks reach size :math:`b`
+
+This directly targets the objective implied by the block rotation bound: balance per-block mass to reduce the
```
pablomlago: Is it possible to make this statement "the objective implied by the block rotation bound" more precise? In the end, the objective is the bound itself, which is minimized by reducing the per-block l1 norms.

docsrc/source/papers/figures/block_size_activation_plots.svg
Giuseppe5, line 1: Do we need the SVG file here? It seems to account for 300K lines of code which is a bit much

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,324 @@
...
+
+   \Phi(X W_1) W_2
+   =
+   \Phi(X W_1 P)\, P^T W_2.
+
+.. admonition:: Key idea 💡
+
+   Brevitas identifies permutation-equivariant regions and merges the calibrated permutation into adjacent linear weights before deployment. As a result, no explicit permutation operator remains in the inference graph. Only rotations that are intentionally kept online incur runtime cost.
+
+Brevitas Integration
```
Giuseppe5: Not a huge fan of this title, since we're already in Brevitas, where else should this be integrated?

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,324 @@
...
+.. code:: shell
+
+   # Try a different block size
+   brevitas_ptq_llm --config=llama3-mixquant-int4.yml --rotation-block_size=16
+
+   # Try a different permutation strategy
+   brevitas_ptq_llm --config=llama3-mixquant-int4.yml --permute-fn=zigzag
+
+
+More Experiments!
```
Giuseppe5: Not a huge fan of this either, it's non descriptive

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,369 @@
...
+        <a href="https://github.com/Xilinx/brevitas/tree/dev/src/brevitas_examples/papers/mixquant">🧪 Examples</a>
+    </div>
+
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 3
+
+
+About the Algorithm
```
Giuseppe5, line 25: I believe there are a bit too many details for being just a documentation section (for example, compare this with the equivalent Qronos section).
I'd try to base the explanations here a lot more on intuition, with pointers to the paper for more appropriate and formal description.

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,369 @@
...
+| Model | float_ppl | quant_ppl | ARC-C | ARC-E | HellaS | PIQA  | WinoG |
++=======+===========+===========+=======+=======+========+=======+=======+
+| 1B    | 11.7      | 17.0      | 27.5  | 52.5  | 38.9   | 67.1  | 52.5  |
++-------+-----------+-----------+-------+-------+--------+-------+-------+
+| 3B    | 9.9       | 13.3      | 32.9  | 61.7  | 45.2   | 71.8  | 54.5  |
++-------+-----------+-----------+-------+-------+--------+-------+-------+
+| 8B    | 6.5       | 8.5       | 43.2  | 74.0  | 51.4   | 74.9  | 59.0  |
++-------+-----------+-----------+-------+-------+--------+-------+-------+
+
+You can override other hyperparamters via CLI. For example:
```
Giuseppe5: This details about the entrypoint should not be here I think

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,369 @@
...
+The MixQuant* config specifies:
+
+- INT4 weights and activations
+- Full-vector Hadamard rotations where mergable (i.e., R1 and R2)
+- Online block Hadamard rotations with block size 32 (i.e., R3)
+- MixQuant permutations via ``massdiff`` (i.e., P3)
+- `Qronos <https://xilinx.github.io/brevitas/dev/papers/qronos.html>`_ rounding [3]
+
+
++-------+-----------+-----------+-------+-------+--------+-------+-------+
```
Giuseppe5, line 225: I would probably separate "How to run" from actual results tables

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,369 @@
...
+
+- ``--disable-block-rotation-for-fused``:
+  If ``rotation`` and ``rotation_block_size`` are specified, keep fused rotations as full-vector and only use block 
+  rotations where rotations remain online. This is a no-op if ``rotation_block_size`` is unset.
+
+
+Getting Started
+---------------
+
+Install Brevitas and the required dependencies:
```
Giuseppe5: I would also considering removing how to install torch/brevitas (but leave the recommended versions".

Giuseppe5 (changes requested):
I consider these documentation pages as a way to get a quick idea behind each paper and algorithm. The points that should be cristal clear are:
- What problem we are trying to solve
- What's the intuition behind our method

At the moment this documentation is a bit too technical for that, which is not inherently an issue but then it overlaps a lot with the paper itself. 

This page should also have a pointer to how to experiment with the paper's idea in Brevitas, but there are some assumptions which I think we can make:
- If you want to learn more about the idea, read the paper
- If you want to know how to install brevitas and its dependencies, read the appropriate part of the description
- If you want to know how to use specific part of the codebase like the LLM entrypoint, go and explore (and read the tutoriaLs/readme, if present).

I am aware that is very difficult to draw a line between some of this categories, but considering how everything is extremely interconnected, we need to try.


I would also probably refactor this (and maybe go back to the other papers too at some point with the following general structure):
- Problem + Solution
- Results
- Pointer to how to reproduce results in Brevitas

In general, I'd probably avoid mixing results and command line argument for how to generate those. We can put that in the README files which accompany anyway all the yaml files.

The last section about reproducibility, should give some tips and tricks about flags and configurations, and interesting features to experiment with.


Another different way to approach this is to have nested documentation pages for each paper, where you can have entire different sections for technical details and such, but in its current form where everything lives in a single page, I don't think that works very well.


This is extremely subjective, and I'd love to hear what @nickfraser and @pablomlago think about this too.

docsrc/source/papers/mixquant.rst
```diff
@@ -0,0 +1,255 @@
...
+balance the per-block :math:`\ell_1` mass across blocks. The default calibration algorithm,
+MassDiff (Algorithm 1 in the paper), is a greedy assignment that:
+
+1. Computes an average magnitude score for each channel
+2. Processes channels in descending order of score
+3. Assigns each channel to the block whose accumulated :math:`\ell_1` mass would increase the least
+4. Continues until all blocks reach size :math:`b`
+
+The result is a permutation that spreads high-magnitude channels evenly across blocks,
+improving outlier suppression for a fixed block size and at no inference cost.
```
Giuseppe5: I'd add a comment about why there is no inference cost, which is the fact that the permutation matrices are fused. I don't think it was mentioned before.

Giuseppe5: Ok, I guess it's mentioned in the next paragraph, but can hint at it here and then explain it later.

docsrc/source/papers/mixquant.rst
```diff
@@ -129,157 +73,131 @@ Permutation-equivariant regions
...
+forward hooks on the sink modules of each permutation-equivariant region to collect activation
+statistics during the calibration pass. On exit, it runs the calibration algorithm to produce
+one permutation per region and fuses each permutation directly into the surrounding weights.
 
-- ``--rotation``:
-  Enables rotation equalization. If not specified, then neither rotations nor permutations are applied.
-  
-  Options:
+``GraphPermutationEqualization`` handles the graph walk and permutation application. It reuses
+the same region-walk logic as used for rotation equalization to find permutation-equivariant regions.
```
Giuseppe5, line 145: I would restore the part of the documentation where you give an example of how to use this with the entrypoint (one "example") and the key CLI arguments to do that.

nickfraser (approved):
LGTM!

PR #1471, opened 2026-02-28
Fix (brevitas_examples/llm): fix to RMSNorm context manager

Giuseppe5:
## Reason for this PR

The rmsnorm_patch context manager goal is to use torch.nn.RMSNorm during tracing, and restore the original RMSNorm implementation  after dynamo.

In its current implementation, we do not swap back to the original, non-torch implementation, because of naming overload.

## Changes Made in this PR

Clarify names and make sure that we restore the correct RMSNorm class.

## Testing Summary

Added test for checking the type of RMSNorm.


nickfraser (approved):
LGTM!

PR #1472, opened 2026-02-28
Fix (rotation): fix extra compile keys in state dict

Giuseppe5:
## Reason for this PR

When compiling, the `state_dict` has extra keys with `orig_mod` prefix attached to it.
The values associated to the keys with and without `orig_mod` are shared.

Check https://github.com/pytorch/pytorch/issues/101107 for more details

## Changes Made in this PR

When we fuse rotations, we adjust the state dict in case the proxy was compiled to avoid state dict issues.

Another potential solution is to modify the `state_dict` method for the Proxy class, and make sure that the `orig_mod` keys are never added to the dict in the first place.

## Testing Summary

Added tests, took the chance to split the rotations tests in their own file


nickfraser:
Any reason for the refactor of the equalization tests? Anything you want to point out? Are there new tests? Is there increased coverage?

Initial glance seems like the refactor is unrelated to the PR itself.

Giuseppe5:
I added another test for rotation, which until now were all inside test_equalize.py

The file was becoming massive, so I took the opportunity to split into his own thing. Basically all the old rotation tests, plus one new one added with this PR, are now in the new file.

We could do this in 2 PRs if needed.

nickfraser (approved):
LGTM!

PR #1473, opened 2026-03-02
test (qonnx): Fixes to QONNX export tests when (`dynamo=True`)

nickfraser:
Up until the release of QONNX v1.0.0, our TruncQuant tests were disabled, since there was a [mismatch between behaviour](#1042 ). With the release, some changes / bugfixes need to be made before re-enabling the tests, specifically:
 - Generate the node "Trunc", not "TruncQuant"
 - Fix issue during dynamo tracing of node not being picked up
 - Remove type hints for fake quant node (not necessary, but now matches other ops)
 - Fix f-string in ONNX export test

### DIFF
--- a/src/brevitas/export/onnx/qonnx/custom_ops.py
+++ b/src/brevitas/export/onnx/qonnx/custom_ops.py
@@ -220,7 +220,7 @@ def _trunc_quant_fake(
 
 
 @onnxscript.script(qonnx_op, default_opset=qonnx_op)
-def TruncQuant(
+def Trunc(
         x: FLOAT,
         scale: FLOAT,
         zero_point: FLOAT,
@@ -245,7 +245,7 @@ def trunc_quant_wrapper(
         rounding_mode: str,
         signed: int,
         narrow: int) -> FLOAT:
-    return TruncQuant(
+    return Trunc(
         x,
         scale,
         zero_point,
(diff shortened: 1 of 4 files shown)


src/brevitas/export/onnx/qonnx/custom_ops.py
```diff
@@ -207,20 +207,20 @@ def trunc_quant(
...
     tensor_x,
-    scale: torch.Tensor,
-    zero_point: torch.Tensor,
-    input_bit_width: torch.Tensor,
-    output_scale: torch.Tensor,
-    output_bit_width: torch.Tensor,
-    rounding_mode: str,
-    signed: int,
-    narrow: int,
+    scale,
```
Giuseppe5, line 210: Not a huge fan of removing type hints (@pablomlago would be appalled), but I understand the consistency reason.

Giuseppe5 (approved):
One small comment, but in general we can merge as soon as tests pass.

src/brevitas/export/onnx/qonnx/custom_ops.py
nickfraser: I think I'll leave it with type hints removed. Ideally, we're [switching to new syntax](https://docs.pytorch.org/docs/stable/onnx_ops.html#symbolic-operators).

PR #1474, opened 2026-03-03
Feat (brevitas_examples/llm): custom trainer support

Giuseppe5:
## Reason for this PR

Currently, our Trainer only supports rotation optimization, with little possibility of customization

## Changes Made in this PR

This PR allows to fully override the Trainer, including what parameters to optimize, which loss to use for each different set of parameters, which callback to use, etc.

## Testing Summary

Added test with custom class


src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -24,8 +27,76 @@
...
+        self.schedulers = schedulers if schedulers else []
+
+    def step(self, *args, **kwargs) -> None:
+        for scheduler in self.schedulers:
+            if scheduler is not None:
+                scheduler.step(*args, **kwargs)
+
+    def get_last_lr(self) -> List[float]:
+        if not self.schedulers or self.schedulers[0] is None:
+            return []
```
Giuseppe5: This could cause an issue if the first scheduler is None, since the Trainer will try to pick this value up and it will produce an out of range error.

We could either use a placeholder, since this function is mostly used for logging, or try to store and return the actual (static) learning rate

src/brevitas_examples/llm/llm_quant/parse_utils.py
```diff
@@ -0,0 +1,85 @@
...
+from brevitas.utils.logging import setup_logger
+from brevitas_examples.common.generative.quantizers import QUANTIZERS_REGISTRY
+from brevitas_examples.llm.llm_quant.rotation_optimization import OPTIMIZER_CONFIG_REGISTRY
+from brevitas_examples.llm.llm_quant.rotation_optimization import TRAINER_REGISTRY
+from brevitas_examples.llm.llm_quant.rotation_optimization import TRAINING_ARGS_REGISTRY
+
+logging = setup_logger(__name__)
+
+
+def parse_custom_quantizer(quant_name: str) -> str:
```
pablomlago: Most of the logic of `parse_custom_quantizer` and `parse_custom_trainer` is shared. Consider refactor the common logic into a single function.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -54,35 +237,41 @@ class TrainingArguments(transformers.TrainingArguments):
...
-        teacher_logits = teacher_logits / temperature
+        student_logits = student_logits.float() / temperature
+        teacher_logits = teacher_logits.float() / temperature
 
         # Compute log probabilities for student and probabilities for teacher
         student_log_probs = F.log_softmax(student_logits, dim=-1)
         teacher_log_probs = F.log_softmax(teacher_logits, dim=-1)
-        student_log_probs = student_log_probs
+
+        if topk > 0:
```
pablomlago: The issue with this codepath is that the sliced probability distribution is not normalized, so it is possible to see negative losses, which might make it difficult to compare losses for different settings of `topk`. Consider adding the extra normalization, as in https://github.com/huggingface/trl/blob/c6d74a15c0c50a39ef9e16e4c867e144ea26d0cf/trl/experimental/distillation/distillation_trainer.py#L1124.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +343,210 @@ def mock_save_pretrained_fn(*args, **kwargs):
     return model
 
 
-def apply_rotation_optimization(
+def _prepare_train_dataset(train_dataset: Dataset) -> Dataset:
```
pablomlago: Remove this function

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +343,210 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
+    return train_dataset
+
+
+def _build_default_optimizers(model: torch.nn.Module, training_args: TrainingArguments) -> tuple:
```
pablomlago: This function should be called `_build_rotation_optimizers`. Also, the type hint for the output can be improved.

src/brevitas_examples/llm/main.py
```diff
@@ -323,11 +294,35 @@ def quantize_llm(args, extra_args=None):
...
+        if args.custom_trainer is not None:
+            custom_trainer_config_name = parse_custom_trainer(args.custom_trainer)
+            # Look up registered overrides by config name.  Each registry
+            # lookup is optional -- the user only needs to register what
+            # they want to customise.
+            if custom_trainer_config_name in TRAINER_REGISTRY.get_registered_keys():
+                custom_trainer_cls = TRAINER_REGISTRY.get(custom_trainer_config_name)
+            if custom_trainer_config_name in TRAINING_ARGS_REGISTRY.get_registered_keys():
+                custom_training_args_cls = TRAINING_ARGS_REGISTRY.get(custom_trainer_config_name)
+            if custom_trainer_config_name in OPTIMIZER_CONFIG_REGISTRY.get_registered_keys():
```
pablomlago: I do not see a reason for having a registry for optimizer configuration (the blockwise trainer allows the same level of customization without requiring to register any configuration).

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -21,11 +25,178 @@
...
 from brevitas.utils.parametrization_utils import extract_trainable_rotation_matrices
+from brevitas.utils.python_utils import Registry
+from brevitas_examples.common.accelerate_utils.accelerate import offload_model
 from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
 
+# Registries for out-of-source customization of the training process.
+# Users can register custom trainers, training argument classes, and
+# optimizer/scheduler/param configurations via a plugin .py file.
+TRAINER_REGISTRY = Registry[type](registry_name="TrainerRegistry")
+TRAINING_ARGS_REGISTRY = Registry[type](registry_name="TrainingArgsRegistry")
```
pablomlago: I do not see a reason for having a registry for training arguments. I would expect that each new Trainer will have its own arguments class associated to it and, if not, a fallback can exist to `transformers.TrainingArguments`.

pablomlago (changes requested):
I think the architecture of the custom trainer needs to be simplified.

src/brevitas_examples/llm/llm_quant/parse_utils.py
```diff
@@ -0,0 +1,80 @@
...
+import importlib
+from pathlib import Path
+
+from brevitas.utils.logging import setup_logger
+from brevitas_examples.common.generative.quantizers import QUANTIZERS_REGISTRY
+
+logging = setup_logger(__name__)
+
+
+def _load_plugin(path: str, module_name: str, file_kind: str) -> None:
```
pablomlago: This function could be parametrized with the registry, allowing to further simplify the public function `parse_custom_quantizer` and `parse_custom_trainer`.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -21,12 +25,201 @@
...
+    optimizer_setup: Optional[Any] = None
+
+
+# Single registry for out-of-source customization of the training process.
+# Users register a TrainerSetup (custom trainer, and optionally training
+# argument class and optimizer setup) under a config name via a plugin .py file.
+TRAINER_SETUP_REGISTRY = Registry[TrainerSetup](registry_name="TrainerSetupRegistry")
+
+
+class MultiOptimizer(torch.optim.Optimizer):
```
pablomlago: I would move this into a separate file, e.g, `trainer_utils.py`.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -21,12 +25,201 @@
...
+        if len(sub_states) != len(self.optimizers):
+            raise ValueError(
+                f"MultiOptimizer.load_state_dict: expected "
+                f"{len(self.optimizers)} sub-optimizer state dicts, "
+                f"got {len(sub_states)}.")
+        for optimizer, sub_state in zip(self.optimizers, sub_states):
+            optimizer.load_state_dict(sub_state)
+
+
+class MultiScheduler:
```
pablomlago: I would move this into a separate file, e.g, trainer_utils.py.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -54,35 +259,49 @@ class TrainingArguments(transformers.TrainingArguments):
...
         default=-1,
         metadata={"help": "Consider the first K logits when computing distillation loss"})
+    kl_loss_reduction: str = field(
+        default="batchmean", metadata={"help": "Reduction mode to use when computing KL loss"})
 
 
 class GeneralizedTrainer(Trainer):
 
-    def __init__(self, args: TrainingArguments = None, **kwargs) -> None:
+    def __init__(self, args: TrainingArguments = None, teacher_model=None, **kwargs) -> None:
```
pablomlago:
```suggestion
    def __init__(self, args: TrainingArguments = None, teacher_model: Optional[torch.nn.Module]=None, **kwargs) -> None:
```

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +373,205 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
         args=training_args,
         train_dataset=train_dataset,
         eval_dataset=None,
         data_collator=collate_fn,
-        optimizers=(optimizer, None))
+        optimizers=optimizers)
+
+    # Select trainer class
+    if trainer_cls is None:
+        if training_args.use_distillation_loss:
```
pablomlago: There is a slight inconsistency, since the type hint for `training_args` is `transformers.TrainingArguments`, but `use_distillation_loss` is not an attribute of this class, but an attribute of the custom `TrainingArguments`. For extra safety, I would do:
```suggestion
        if isinstance(training_args, TrainingArguments):
```
or
```suggestion
        if hasattr(training_args, 'use_distillation_loss'):
```

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -21,12 +25,201 @@
...
 from brevitas.graph.calibrate import quantization_status_manager
 from brevitas.optim.cailey_sgd import CaileySGD
 from brevitas.utils.parametrization_utils import extract_trainable_rotation_matrices
+from brevitas.utils.python_utils import Registry
+from brevitas_examples.common.accelerate_utils.accelerate import offload_model
 from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
 
 
+@dataclass
+class TrainerSetup:
```
pablomlago: I don't think this extra `dataclass` is necessary. Ideally, each `trainer_cls` should have an associated `training_args_cls` (with a fallback to `transformers.TrainingArguments`), and the way the optimizers/schedulers are handled should be harmonized with `LearnedRoundTrainer`:
```
@dataclass
class OptimizerArgs:
    target_params: Union[str, TargetParamFn] = field(
        metadata={
            "help": ("Targets to be optimized."),
            "choices": TARGET_PARAM_FN_REGISTRY.get_registered_keys(),})
    optimizer_cls: Union[str, Type[Optimizer]] = field(
        default="adam",
        metadata={"help": "The optimizer to use."},
    )
    lr: float = field(
        default=1e-3,
        metadata={"help": "Initial learning rate for the optimizer."},
    )
    optimizer_kwargs: Optional[Union[Dict, str]] = field(
        default=None,
        metadata={"help": "Extra keyword arguments for the optimizer."},
    )
    lr_scheduler_args: Optional[LRSchedulerArgs] = field(
        default=None,
        metadata={
            "help": ("Hyperparameters of learning rate scheduler for the selected"
                     "optimizer.")},
    )
```
This functionality can be probably expanded/refactored, but I do not think we should have completely separate codepaths for doing more or less the same thing.

pablomlago (changes requested):
Ideally, we should harmonize the optimizer/scheduler instantiation for `GeneralizedTrainer` and `LearnedRoundTrainer`.

src/brevitas_examples/common/learned_round/learned_round_args.py
```diff
@@ -26,7 +26,7 @@
 LR_SCHEDULER_NAMESPACES = [torch.optim.lr_scheduler]
 
 
-def _parse_optimizer_class(optimizer_str: str) -> Type[Optimizer]:
+def parse_optimizer_class(optimizer_str: str) -> Type[Optimizer]:
```
pablomlago: Consider extracting `parse_optimizer_class` and `parse_lr_scheduler_class` outside the `learned_round` namespace.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -42,6 +58,27 @@ class TrainingArguments(transformers.TrainingArguments):
             "help":
                 "Data type for CaileySGD optimizer computations. None means use parameter dtype."})
 
+    ### Multi-optimizer/scheduler args
```
pablomlago: Could we have a nested dataclass for these arguments?

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -54,35 +91,58 @@ class TrainingArguments(transformers.TrainingArguments):
...
+    # Class attributes consumed by the LLM entrypoint when this trainer is
+    # registered via ``--custom-trainer``. Subclasses may override them to
+    # customise the training arguments and optimizer/scheduler setup. When left
+    # at their defaults (``training_args_cls`` is the built-in
+    # ``TrainingArguments`` and ``optimizer_setup`` is ``None``), the default
+    # behaviour of the LLM example is used.
+    training_args_cls: Type[transformers.TrainingArguments] = TrainingArguments
+    optimizer_setup: Optional[Callable[[], List["OptimizerParamsSpec"]]] = None
+
+    def __init__(self, args: TrainingArguments = None, teacher_model=None, **kwargs) -> None:
```
pablomlago: Type hints for this signature.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -116,8 +178,13 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
         return (loss, outputs) if return_outputs else loss
 
 
-def parse_rotation_optimization_args(extra_args: Optional[List[str]] = None) -> TrainingArguments:
-    parser = transformers.HfArgumentParser(TrainingArguments)
+def parse_rotation_optimization_args(
```
pablomlago, line 74: None of the arguments of this function should be `Optional`.

src/brevitas_examples/llm/main.py
```diff
@@ -337,11 +306,36 @@ def quantize_llm(args, extra_args=None):
...
+        custom_training_args_cls = None
+
+        if args.custom_trainer is not None:
+            custom_trainer_config_name = parse_custom_trainer(args.custom_trainer)
+            # Look up the registered Trainer class by config name. The training
+            # arguments class and optimizer setup are read from the trainer's
+            # ``training_args_cls`` and ``optimizer_setup`` class attributes.
+            # When these are not defined (``None``), the defaults of the LLM
+            # example are used downstream.
+            custom_trainer_cls = TRAINER_REGISTRY.get(custom_trainer_config_name)
```
pablomlago: I do not think we should have this logic in the entrypoint. Probably this unpacking should be moved to the `rotation_optimization.py` class, and remove the arguments `optimizer_setup` and `training_args` (pass `extra_args` instead, doing the parsing inside that function) from  `apply_fine_tuning` .

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -24,7 +28,19 @@
 from brevitas.graph.calibrate import quantization_status_manager
 from brevitas.optim.cailey_sgd import CaileySGD
 from brevitas.utils.parametrization_utils import extract_trainable_rotation_matrices
+from brevitas_examples.common.accelerate_utils.accelerate import offload_model
 from brevitas_examples.common.accelerate_utils.accelerate import remove_hooks
+# Optimizer/scheduler building and trainer plumbing live in trainer_utils. They
+# are re-exported here for backward-compatible imports.
```
pablomlago: Why is backward-compatibility required here? These classes did not exist before in `dev`.

pablomlago (changes requested):
Further discussion on the structure might be required.

src/brevitas_examples/llm/llm_args.py
```diff
@@ -513,9 +536,20 @@ def fx_required(args: Namespace):
...
 def validate(args: Namespace, extra_args: Optional[List[str]] = None) -> None:
+    # --optimize-rotations is a deprecated alias for --fine-tune
     if args.optimize_rotations:
+        warn(
+            "--optimize-rotations is deprecated, use --fine-tune instead.",
+            DeprecationWarning,
+            stacklevel=2)
+        args.fine_tune = True
+    # --custom-trainer implies --fine-tune
+    if args.custom_trainer is not None:
```
Giuseppe5: Remove

src/brevitas_examples/llm/main.py
```diff
@@ -337,11 +305,20 @@ def quantize_llm(args, extra_args=None):
...
-        rot_optimization_args = parse_rotation_optimization_args(extra_args=extra_args)
-        # Load the data for rotation optimization
-        rot_calibration_dataset = get_dataset_for_model(
+    if args.fine_tune:
+        # Load custom training plugin if specified. The registered Trainer class
+        # carries its own ``training_args_cls`` class attribute (which in turn
+        # defines the optimizer setup via ``optimizer_scheduler_args``), consumed
+        # inside apply_fine_tuning.
+        custom_trainer_cls = None
+        custom_callbacks = None
```
Giuseppe5: Remove if not used

src/brevitas_examples/llm/main.py
```diff
@@ -337,11 +305,20 @@ def quantize_llm(args, extra_args=None):
...
-        # Extra arguments should be used as training arguments for rotation optimization
-        rot_optimization_args = parse_rotation_optimization_args(extra_args=extra_args)
-        # Load the data for rotation optimization
-        rot_calibration_dataset = get_dataset_for_model(
+    if args.fine_tune:
+        # Load custom training plugin if specified. The registered Trainer class
+        # carries its own ``training_args_cls`` class attribute (which in turn
+        # defines the optimizer setup via ``optimizer_scheduler_args``), consumed
+        # inside apply_fine_tuning.
+        custom_trainer_cls = None
```
Giuseppe5: Maybe we move it to when we need it

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +206,119 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
+        trainer_cls = GeneralizedTrainer if getattr(
+            training_args, 'use_distillation_loss', False) else Trainer
+
+    # Wire the teacher model whenever the selected trainer is a
+    # GeneralizedTrainer subclass and distillation loss is enabled.
+    if issubclass(trainer_cls, GeneralizedTrainer) and getattr(
+            training_args, 'use_distillation_loss', False):
+        trainer_kwargs["teacher_model"] = copy.deepcopy(model.cpu())
+
+    if callbacks is not None:
```
Giuseppe5: Remove

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +206,119 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
-        lr=training_args.learning_rate,
-        stiefel=True,
-        dtype=training_args.optimizer_dtype)
-    trainer = GeneralizedTrainer(
+
+    # Build optimizer / scheduler pair from the training args. When no
+    # optimizer_scheduler_args are provided (e.g. a custom trainer that relies on
+    # the HF Trainer's built-in optimizer), pass (None, None) so the HF Trainer
+    # uses its built-in optimizer.
+    if training_args.optimizer_scheduler_args is not None:
```
Giuseppe5: This should raise an error. If no optimizer_scheduler_args are provided, there are no parameters with requires_grad=True

tests/brevitas_examples/llm_test_plugin.py
pablomlago, line 1: Rename file, e.g. `llm_test_plugin.py` or something along those lines.

src/brevitas_examples/llm/main.py
```diff
@@ -620,21 +586,38 @@ def quantize_llm(args, extra_args=None):
...
+
+            fine_tune_extra_args = list(extra_args) if extra_args is not None else []
             if args.load_checkpoint:
-                rot_optimization_args.max_steps = 0
-            apply_rotation_optimization(
+                # Skip training when loading from a checkpoint by forcing
+                # max_steps to 0 through the training arguments. Appended last so
+                # that it overrides any user-provided --max_steps (the argument
+                # parser keeps the last value for a repeated flag).
+                fine_tune_extra_args += ["--max_steps", "0"]
```
pablomlago, line 606: Was this tested? If `max_steps` was already set, which setting is taking precedence? Probably fine-tuning could be skipped altogether if `args.load_checkpoint = True`.

src/brevitas_examples/llm/main.py
```diff
@@ -620,21 +586,38 @@ def quantize_llm(args, extra_args=None):
...
+            # carries its own ``training_args_cls`` class attribute (which in turn
+            # defines the optimizer setup via ``optimizer_scheduler_args``), consumed
+            # inside apply_fine_tuning.
+            custom_trainer_cls = None
+
+            if args.custom_trainer is not None:
+                custom_trainer_config_name = parse_custom_trainer(args.custom_trainer)
+                custom_trainer_cls = TRAINER_REGISTRY.get(custom_trainer_config_name)
+
+            fine_tune_extra_args = list(extra_args) if extra_args is not None else []
```
pablomlago: Why is the cast to `list` needed now? It should already be a  `list`.

src/brevitas_examples/llm/main.py
```diff
@@ -337,11 +305,9 @@ def quantize_llm(args, extra_args=None):
...
     validation_loader = DataLoader(dataset=validation_dataset, batch_size=1, collate_fn=collate_fn)
 
-    if args.optimize_rotations:
-        # Extra arguments should be used as training arguments for rotation optimization
-        rot_optimization_args = parse_rotation_optimization_args(extra_args=extra_args)
-        # Load the data for rotation optimization
-        rot_calibration_dataset = get_dataset_for_model(
+    if args.fine_tune:
+        # Load the data for training
+        train_dataset = get_dataset_for_model(
```
pablomlago: Given that the invoked function later is `apply_fine_tuning`, I would rename this variable for extra clarity.
```suggestion
        finetune_dataset = get_dataset_for_model(
```

src/brevitas_examples/llm/llm_quant/trainer_utils.py
```diff
@@ -0,0 +1,324 @@
...
+        # tolerates None entries until then.
+        multi_scheduler = MultiScheduler(schedulers)
+        return multi_optimizer, multi_scheduler
+    else:
+        # A None scheduler here lets the HF Trainer build its default scheduler.
+        return optimizers[0], schedulers[0]
+
+
+@dataclass
+class TrainingArguments(transformers.TrainingArguments):
```
pablomlago, line 280: This should live in the same file as `GeneralizedTrainer`.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +250,101 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
 
-def apply_rotation_optimization(
+def apply_fine_tuning(
         model: torch.nn.Module,
         tokenizer: PreTrainedTokenizerBase,
         train_dataset: Dataset,
-        training_args: TrainingArguments,
-        collate_fn: Callable) -> None:
+        collate_fn: Callable,
+        custom_trainer_cls: Optional[Type[Trainer]] = None,
```
pablomlago: I think inside this function (and maybe even in the entrypoint), we can shorten `custom_trainer_cls` to `trainer_cls`. I do not think, the `custom` prefix is providing any useful information.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +250,101 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
         train_dataset=train_dataset,
         eval_dataset=None,
         data_collator=collate_fn,
-        optimizers=(optimizer, None))
+        optimizers=optimizers)
+
+    # When no custom trainer is given but the model has trainable rotation
+    # matrices, default to RotationTrainer (CaileySGD on the rotations, expressed
+    # through the standard optimizer_scheduler_args mechanism).
+    if custom_trainer_cls is None:
```
pablomlago: Can we move this inside the `train` function of `RotationTrainer`, e.g.:
```
def train(model, *args, **kwargs):
   # perform check
   return super().train(model, *args, **kwargs)
```

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -147,41 +250,101 @@ def mock_save_pretrained_fn(*args, **kwargs):
...
+            raise RuntimeError(
+                "No Custom Trainer has been defined and no optimizable rotations are present in the model."
+            )
+        trainer_cls = RotationTrainer
+    else:
+        trainer_cls = custom_trainer_cls
+
+    # Wire the teacher model whenever the selected trainer is a
+    # GeneralizedTrainer subclass and distillation loss is enabled.
+    if issubclass(trainer_cls, GeneralizedTrainer) and getattr(
```
pablomlago, line 212: This should be handled inside the `GeneralizedTrainer` class. Maybe inside `train` before calling `super().train`?

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
```diff
@@ -22,67 +25,107 @@
...
-
-        if topk > 0:
-            teacher_logits, indices = teacher_logits.topk(topk, dim=-1, sorted=False)
-            student_log_probs = student_log_probs.gather(-1, indices).flatten(0, -2)
-
+        out_dtype = student_logits.dtype
         # Apply temperature scaling
-        student_logits = student_logits / temperature
-        teacher_logits = teacher_logits / temperature
+        student_logits = student_logits.float() / temperature
```
pablomlago: Can we have a comment on why this is necessary?

pablomlago (changes requested):
Some minor comments.

src/brevitas_examples/llm/main.py
Giuseppe5: It was tested, the comment was updated.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
Giuseppe5: No this has to be defined before we instantiate the class that calls train.

src/brevitas_examples/llm/llm_quant/rotation_optimization.py
Giuseppe5: As above, we need to decide which class we use before we call train.

PR #1482, opened 2026-03-06
fix (fx): bump version guard for value tracer

nickfraser:
Bump the `value_tracer` version to `pytorch<2.11`


Giuseppe5:
Can you add a deprecation warning?
I'm happy to say "in future versions", without being extremely specific, for now.

Giuseppe5 (approved):

PR #1486, opened 2026-03-06
ci (python): testing newer python versions

nickfraser:
Builds on https://github.com/Xilinx/brevitas/pull/1451 & https://github.com/Xilinx/brevitas/pull/1477 - test newer versions of python. Superseeds #1483.


Giuseppe5:
We could reasonably say that the PR #1451 introduces python 3.13 compatibility.
The tests that are failing are mostly related to new torch versions:
- torch.compile tests don't run well on CPU. We have exclusion criteria for torch version up to 2.7. Tests with higher torch versions might fail
- ONNX dynamo export with torch 2.9 seems to have issue

PR #1488, opened 2026-03-09
Feat (quant/float): move float quant out of experimental phase

Giuseppe5:
## Reason for this PR

Float quantizers (MX and not) have been tested and used for a while now, so we can move them out of the `experimental` category.

For now back-compatibility is ensured, but in the future it will be completely dropped.

## Changes Made in this PR

Move float quantizers out of the experimental folder, while preserving old import.


src/brevitas/quant/experimental/float.py
```diff
@@ -1,141 +1,32 @@
...
+
+warnings.warn(
+    "Importing from 'brevitas.quant.experimental.float' is deprecated. "
+    "Please use 'brevitas.quant.float' instead. "
+    "Support for importing from the old path will be removed in a future release.",
+    DeprecationWarning,
+    stacklevel=2,
+)
+
+from brevitas.quant.float import *
```
pablomlago, line 14: Should this import be removed or the following ones?

Giuseppe5: I would remove this, and have all of them listed.
Easier to understand if something breaks

pablomlago (approved):

Issue #1479, opened 2026-03-05
Broken interop between GPxQ and ConvTranspose2d


i-colbert:
**Describe the bug**

All GPxQ algorithms (GPTQ, GPFQ, Qronos, etc.) rely on modifying a local `weight` tensor that shares memory with `self.layer.weight.data`. This allows `get_quant_weights()` to see modifications. However, for `ConvTranspose2d` layers, this memory sharing is broken.

__Root cause:__ PyTorch memory behavior with transpose + flatten:

```python
if isinstance(self.layer, SUPPORTED_CONV_OP):
    if is_conv_transposed(self.layer):
        # View - same memory, changed strides
        weight = weight.transpose(1, 0) 
    # Issue is here for non-contiguous. It is only a view for contiguous (eg, Conv2d)
    weight = weight.flatten(1)  
```

You cannot:

- Transpose (changing strides to non-contiguous)
- Then flatten across non-contiguous dimensions
- And expect no copy

Flattening across permuted memory is fundamentally incompatible with stride-only views.

**Reproducibility**
- [x] Can be reproduced consistently.
- [ ] Difficult to reproduce.
- [ ] Unable to reproduce.

**To Reproduce**

```python
>>> import torch.nn as nn

# ConvTranspose2d with transpose - CREATES COPY
>>> layer = nn.ConvTranspose2d(3, 32, 3)
>>> weight = layer.weight.data
>>> weight.data_ptr()
94067644132544
>>> weight = weight.transpose(1, 0)
>>> weight.data_ptr()
94067644132544  # Same - transpose is a view
>>> weight = weight.flatten(1)
>>> weight.data_ptr()
94067752850112  # DIFFERENT - flatten copied!

# ConvTranspose2d without transpose - NO COPY
>>> layer = nn.ConvTranspose2d(3, 32, 3)
>>> weight = layer.weight.data
>>> weight.data_ptr()
94067752850112
>>> weight = weight.flatten(1)
>>> weight.data_ptr()
94067752850112  # Same - contiguous memory preserved
```

**Expected behavior**

Modifications to the working weight tensor should be visible to `get_quant_weights()`. For `ConvTranspose2d`, this currently doesn't work because the tensor is detached from layer memory.

_If known:_

 - Brevitas version: dev (c2b9e7)
 - PyTorch version: 2.6
 - Operating System / platform: Linux

**Additional context**
[...]

PR #1490, opened 2026-03-13
Fix (graph/gpxq): ConvTranspose temporarily unsupported

Giuseppe5:
## Reason for this PR

Temporary patch for #1479 

## Changes Made in this PR

Disable GPxQ for ConvTranspose

The remaining code around conv_transpose support is still left in place, so that developing a solution can start from there.


nickfraser (approved):
LGTM!

PR #1491, opened 2026-03-13
fix (stats): Mix MSE for zero-point

nickfraser:
# Summary

Fixes a correctness bug in MSE-based quantization for asymmetric (shifted) weight quantizers, refactors the MSE search for clarity/extensibility, and adds several example generative quantizers.

## Core fix: MSE for asymmetric weights (`src/brevitas/core/stats/stats_op.py`)

- **Bug fix:** `MSE.mse_search` no longer applies `torch.abs` to the search seed (`init`). Forcing `init` positive flipped the sign of the zero-point statistic (`NegativeMinOrZero`), so the search optimized a value that inference then applied with the opposite sign — breaking `ShiftedUint8Weight*FloatMSE`. `init` now keeps its sign.
- **Robust interval ordering:** since `init` (and thus `base`) can now be negative, the search endpoints are explicitly ordered via `minimum`/`maximum` so both `grid` and `fibonacci` searches receive an ascending `[xl, xr]` interval.
- **Rename:** `restrict_scale_positive` → `bipolar_search` (inverted semantics, default `False`) for clarity. Wired through:
  - `src/brevitas/quant/base.py` — MSE scale sub-injectors forward `bipolar_search`.
  - `src/brevitas/quant/solver/common.py` — `SolveRestrictScaleSign` adds `bipolar_search = not restrict_scale_positive`.
- **Pluggable base computation:** the search lower bound is now produced by a module, `mse_base_op` (the new `MSEUniformStepBase`, computing `init / mse_iters`), so the calculation can be swapped without changing `MSE`.
- **Required MSE dependencies via injection:** `mse_iters` and `mse_base_op` are now required arguments of `MSE` (no in-`__init__` defaults), with the defaults (`mse_iters = 20`, `mse_base_op = MSEUniformStepBase`) provided by the MSE sub-injectors in `src/brevitas/quant/base.py` (`MSESymmetricScaleSubInjector`, `MSEAsymmetricScaleSubInjector`, `MSEZeroPointSubInjector`). `MSE.__init__` parameters were reordered so required params precede defaulted ones.

## Tests

- `tests/brevitas_examples/test_quantize_model.py` — `test_layerwise_stats_vs_mse` is parametrized over `weight_quant_type` (`sym`/`asym`), providing regression coverage that reconstruction error of MSE ≤ stats for asymmetric weights (previously failing).
- `tests/brevitas/core/test_stats.py` — updated the MSE attribute assertion to `bipolar_search`.

## Example quantizers (`src/brevitas_examples/common/generative/`)

- Added asymmetric/MSE and group-wise quantizers: `IntWeightSymmetricGroupQuantMSE`, `ShiftedUint8WeightGroupQuantFloatMSE`, and registered `per_group` sym/asym MSE entries in `WEIGHT_QUANT_MAP`.
- Added `Fp8e4m3OCPDynamicActPerTensorFloat` and registered a `per_tensor` dynamic-stats OCP act entry in `INPUT_QUANT_MAP`.
- Fixed dynamic float activation quantizers to set `proxy_class = DynamicActFloatQuantProxyFromInjector` (and dropped the incorrect `DynamicActProxyMixin` from the FP8 per-group variants).

## Future Considerations

- **Base interval vs. bipolar search:** `MSEUniformStepBase` computes the interval lower bound as `init / mse_iters` independent of `bipolar_search`. Bipolar search only halves the *per-side* iteration count (`mse_iters // 2`); it does not change the bound, so the search interval `[init / mse_iters, init]` is identical for unipolar and bipolar searches at a given `mse_iters`. As a result, doubling `mse_iters` (e.g. to preserve the per-side iteration count under bipolar search) also moves the lower bound to `init / (2 * mse_iters)`, widening the interval — which may be unintuitive. A custom `mse_base_op` can decouple these if needed.

## Backward-compatibility note

- The `MSE` constructor argument/attribute `restrict_scale_positive` is renamed to `bipolar_search` (with inverted meaning). Any external code referencing `restrict_scale_positive` on the MSE stats impl must be updated.
- `MSE` now requires `mse_iters` and `mse_base_op` to be supplied (previously defaulted), and its `__init__` parameter order changed. All in-repo instantiation is via dependency injection (keyword-based), but any external code constructing `MSE` directly/positionally must pass these arguments and account for the new ordering.


Giuseppe5 (approved):

src/brevitas/core/stats/stats_op.py
```diff
@@ -567,7 +577,8 @@ def __init__(
             stats_reduce_dim: Optional[int] = None,
             mse_search_method: str = 'fibonacci',
             mse_iters: int = 20,
-            restrict_scale_positive: bool = True):
+            bipolar_search: bool = False,
+            mse_base_op: Optional[torch.nn.Module] = None):
```
nickfraser: Is it best to leave this as optional, or should the default be provided by the `MSESubInjector`s? If we move to SubInjectors, then `mse_iters` is _always_ defined, so it might make sense to remove the default value from MSE as well (to make a clear failure when this is not defined)

src/brevitas/core/stats/stats_op.py
```diff
@@ -584,7 +595,10 @@ def __init__(
         self.search_method = mse_search_method
         self.stats_reduce_dim = stats_reduce_dim
         self.local_loss_mode: bool = False
-        self.restrict_scale_positive = restrict_scale_positive
+        self.bipolar_search = bipolar_search
+        if mse_base_op is None:
+            mse_base_op = MSEUniformStepBase(mse_iters)
```
nickfraser: Similar to above, if this is provided by the `MSESubInjectors`, then this could be removed.

nickfraser (commented):
One design decision needs to be reconsidered.

src/brevitas/core/stats/stats_op.py
Giuseppe5: I think I prefer having both the class default argument and the subinjector to be consistent between them, and avoid using None if possible.
This means changing this to have `MSEUniformStepBase` as default instead of None, and remove the `if` clause a few lines below; also we would have the subinjectors setting`mse_base_op = MSEUniformStepBase` explicitly.

Another possible solution, to avoid changing too many lines, is to just change this class default, and do not change the injector (so we follow the default set by the class), which is a pattern that we also have in some cases.
I'm not a huge fan of this, since it makes more difficult to understand what class we are using without instantiating them and inspecting them.

nickfraser: I also prefer what you're saying, but you didn't comment on the `mse_iters` issue: if I make the change you suggest, then either:
 - I need to set `mse_iters = 20` as the default in both classes (annoying to maintain)
 - Remove the default value of `mse_iters` in both, and provide this through subinjectors instead.
 
 Do you have a preference? I think I prefer moving `mse_iters` to the subinjectors too.

src/brevitas/core/stats/stats_op.py
nickfraser: Resolved in [1fd0008](https://github.com/Xilinx/brevitas/pull/1491/commits/1fd00080b9239e6a3284708cce5f174acbcd134f)

src/brevitas/core/stats/stats_op.py
nickfraser: Resolved in [1fd0008](https://github.com/Xilinx/brevitas/pull/1491/commits/1fd00080b9239e6a3284708cce5f174acbcd134f).

Giuseppe5 (approved):

PR #1493, opened 2026-03-16
test (ex/llm): remove redundant mock from lighteval tests

nickfraser:
Probably should've been included in #1492. The previous update made the patch in the lighteval tests redundant.

A local tests showed that the tests take approximately the same time to run - showing the update still had the right affect (limit the size of the test).


tests/brevitas_examples/test_llm.py
pablomlago, line 6: Remove unused import

pablomlago (approved):
LGTM (assuming test pass)

PR #1458, opened 2026-02-18
Feat (data): compute perplexity at float32

i-colbert:
## Reason for this PR

Decouples model precision from evaluation precision. Allows running perplexity evaluation at a different precision (e.g., `float32`) than the model execution precision (e.g., `bfloat16`) for more accurate evaluation metrics.

## Changes Made in this PR

Added `dtype` parameter to `compute_perplexity()` in `eval.py`

## Testing Summary

No existing tests directly use `compute_perplexity`. The LLM tests currently use float32, so no updates needed.

## Risk Highlight

- [ ] This PR includes code from another work (please detail).
- [ ] This PR contains API-breaking changes.
- [ ] This PR depends on work in another PR (please provide links/details).
- [ ] This PR introduces new dependencies (please detail).
- [ ] There are coverage gaps not covered by tests.
- [ ] Documentation updates required in subsequent PR.

## Checklist

- [ ] Code comments added to any hard-to-understand areas, if applicable.
- [ ] Changes generate no new warnings.
- [ ] Updated any relevant tests, if applicable.
- [x] No conflicts with destination `dev` branch.
- [x] I reviewed my own code changes.
- [ ] Initial CI/CD passing.
- [ ] 1+ reviews given, and any review issues addressed and approved.
- [ ] Post-review full CI/CD passing.

PR #1494, opened 2026-03-17
Feat (graph/equalize): LayerNorm affine merge at float64

Giuseppe5:
## Reason for this PR

It is safer to perform operations like merging of affine parameters of LN at float64

## Changes Made in this PR

As described above, plus tests

### DIFF
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -2286,7 +2286,7 @@ class MergeLnAffine(GraphTransform, RegionWalkMixin):
 
                 for name in region.sinks:
                     module = region.get_module_from_name(name)
-                    if not _is_supported_ln_shape(layernorm_module, module):
+                    if not _is_supported_ln_shape(layernorm_module, module.in_features):
                         _raise_merge_ln_error(layernorm_module, module)
 
                     scale_bias = id(module) not in scaled_biases
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1830,11 +1830,6 @@ def _is_supported_ln_shape(layer_norm, channels):
     ) == channels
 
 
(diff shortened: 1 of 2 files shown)


tests/brevitas/graph/test_ln_affine_merge.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
```
i-colbert, line 1: 2026?

src/brevitas/graph/equalize.py
```diff
@@ -1813,28 +1813,53 @@ def fuse_parametrizations(model: nn.Module) -> nn.Module:
...
+        f"{next_module.__class__.__name__}. LayerNorm/RMSNorm affine parameters must be channelwise "
+        f"and the sink must be Linear.")
 
-    # Merge bias, new_bias includes the bias of next_module by going through its fwd
-    if hasattr(layer_norm, 'bias'):
-        inp = layer_norm.bias.data.view(view_shape)
-        new_bias = next_module(inp)
-        _replace_bias(next_module, new_bias)
+
+def _merge_ln(layer_norm, next_module, scale_bias_by_weight):
```
i-colbert, line 1840: For consistency, I'd suggest following the same pattern we used in `compute_perplexity` (see #1458), where we added `dtype=torch.float32` directly to the function signature so callers could override it for backward compatibility. It seems like we should do the same here by exposing `dtype=torch.float64` as an input parameter instead of hard‑coding it.

Giuseppe5: I would consider this less of a choice and more of a proper bug-fix where, although I understand the issue of backward compatibility, we need to accept that we're fixing something that might change past behaviour.

PR #1496, opened 2026-03-20
Feat (loss): support for high precision loss

Giuseppe5:
## Reason for this PR

This loss is sensitive to precision; when adding lots of element, computing at (b)float16 could cause overflow and incorrect computation

## Changes Made in this PR

Support for optional high-precision loss computation


nickfraser (approved):
LGTM!

PR #1505, opened 2026-04-04
Feat (utils): replace weights with quantized ones

Giuseppe5:
## Reason for this PR

In certain instances, like learned round, the weight tensors has extra parameters attached to it that could make export somewhat complicated.

## Changes Made in this PR

We perform a destructive replacement of the original weight tensor with its quantized counterparts. This allows for easier exports in many scenarios.

There is an optional flag to keeep track of the original weights.

After merging, we reset the quantizers, including setting rounding mode to _round_, which is what is most commonly supported during export process.

What is missing:
- [ ] Integration in the LLM entrypoint.
- [ ] Handling the WeightWithInputQuantizer (i.e., the one required by A2Q and A2Q+)

## Testing Summary

New tests added.

### DIFF
--- a/src/brevitas/nn/utils.py
+++ b/src/brevitas/nn/utils.py
@@ -98,25 +98,23 @@ def merge_quant_weights(
 
     hooks: List[RemovableHandle] = []
     proxy_list: List[WeightQuantProxyFromInjectorBase] = []
+    module_tensor_id_mapping: Dict = {}
 
     def hook(module: WeightQuantProxyFromInjectorBase, args: Tuple[Any, ...], output: Any) -> None:
-        module_tensor_id_mapping: Dict = {}
         input_tensor = args[0]
         with torch.no_grad():
             for m in module.tracked_module_list:
                 # We match the module based on its weights and the ID of the tensor to quantize
                 if id(m.weight.data) == id(input_tensor.data):
-                    # We track how many modules have been converted
-                    module_tensor_id_mapping[m] = [output.value.data]
-
-            if len(module_tensor_id_mapping) == len(module.tracked_module_list):
-                for quant_module, new_weights in module_tensor_id_mapping.items():
                     quant_module.weight.data = output.value.data
-                proxy_list.append(module)
-            else:
-                warnings.warn(
-                    "Could not match all the modules to their weights, skipping quantizer")
+                    # We track how many modules have been converted
+                    if module not in module_tensor_id_mapping:
+                        module_tensor_id_mapping[module] = 1
+                    else:
+                        module_tensor_id_mapping[module] += 1
+            proxy_list.append(module)
 
+    
     # Register Proxy hooks
     for module in model.modules():
         if not isinstance(module, WeightQuantProxyFromInjectorBase):
@@ -133,10 +131,13 @@ def merge_quant_weights(
         for h in hooks:
             h.remove()
         hooks.clear()
+    
 
     # Reset quantizers from LEARNED_ROUND to ROUND
     with torch.no_grad():
         for module in proxy_list:
+            if module_tensor_id_mapping[module] < len(module.tracked_module_list):
+                raise RuntimeError("Not all weights associated to this quantizer were replaced")
             _reset_quantizer(module)
 
 
--- a/src/brevitas/nn/utils.py
+++ b/src/brevitas/nn/utils.py
@@ -108,15 +108,15 @@ def merge_quant_weights(
                 if id(m.weight.data) == id(input_tensor.data):
                     # We track how many modules have been converted
                     module_tensor_id_mapping[m] = [output.value.data]
-            
+
             if len(module_tensor_id_mapping) == len(module.tracked_module_list):
-                for quant_module, new_weights in module_tensor_id_mapping:
+                for quant_module, new_weights in module_tensor_id_mapping.items():
                     quant_module.weight.data = output.value.data
                 proxy_list.append(module)
             else:
                 warnings.warn(
                     "Could not match all the modules to their weights, skipping quantizer")
-            
+
     # Register Proxy hooks
     for module in model.modules():
         if not isinstance(module, WeightQuantProxyFromInjectorBase):
--- a/src/brevitas/nn/utils.py
+++ b/src/brevitas/nn/utils.py
@@ -99,8 +99,7 @@ def merge_quant_weights(
     hooks: List[RemovableHandle] = []
     module_tensor_id_mapping: Dict = {}
 
-    def hook(
-            module: WeightQuantProxyFromInjectorBase, args: Tuple[Any, ...], output: Any) -> None:
+    def hook(module: WeightQuantProxyFromInjectorBase, args: Tuple[Any, ...], output: Any) -> None:
         input_tensor = args[0]
         with torch.no_grad():
             for m in module.tracked_module_list:
--- a/src/brevitas/nn/utils.py
+++ b/src/brevitas/nn/utils.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from typing import Any
 from typing import Dict
 from typing import List
 from typing import Tuple
@@ -10,10 +11,8 @@ from torch.nn import Parameter
 from torch.utils.hooks import RemovableHandle
 
 from brevitas import config
-from brevitas.core.function_wrapper.learned_round import LearnedRoundSte
 from brevitas.inject.enum import FloatToIntImplType
 from brevitas.inject.enum import ScalingImplType
-from brevitas.quant_tensor import _unpack_quant_tensor
 from brevitas.utils.torch_utils import compute_channel_view_shape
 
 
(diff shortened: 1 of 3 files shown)


src/brevitas/nn/utils.py
```diff
@@ -60,6 +69,100 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+        model.eval()
+        with merge_learned_round(model):
+            model(sample_input)
+        # Weights are now merged and rounding mode is ROUND.
+
+    Args:
+        model: A model containing quantised layers with learned round quantisers.
+    """
+
+    def __init__(self, model: torch.nn.Module, disable_quant: bool = True) -> None:
```
nickfraser: - add destructive/non-destructive optional argument
 - don't just replace rounding implementation, replace the full quantizer with a simple quantizer (const and/or parameter scale)

nickfraser (changes requested):
See comments

src/brevitas/nn/utils.py
```diff
@@ -60,6 +70,117 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
         state_dict[keys_map[old_key]] = state_dict.pop(old_key)
 
 
+class merge_quant_weights:
```
pablomlago: I do not see a reason why this should be a context manager, specially considering you need to add a check to verify that one one forward pass is run. For me, this should be a function that accept a sample and runs the forward within the function.

src/brevitas/nn/utils.py
```diff
@@ -60,6 +70,117 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+    """Context manager that merges quantized weights into model weights.
+
+    This could be useful for example with Learned Round.
```
pablomlago, line 78: From what I see, this is the main purpose of the context manager. What else is this context manager intended to cover? Also, this docstring does not reflect the fact that the scales are also updated to `PARAMETER_FROM_STATS`.

src/brevitas/nn/utils.py
```diff
@@ -60,6 +70,117 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+
+    def __init__(
+            self,
+            model: torch.nn.Module,
+            disable_quant: bool = True,
+            preserve_original_weights: bool = False) -> None:
+        self._model = model
+        self._hooks: List[RemovableHandle] = []
+        self._module_tensor_id_mapping = {}
+        self.disable_quant = disable_quant
```
pablomlago: Is there a reason why some attributes are public and other private? E.g., `_model` and `disable_quant`.

src/brevitas/nn/utils.py
```diff
@@ -60,6 +70,117 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        for hook in self._hooks:
+            hook.remove()
+        self._hooks.clear()
+
+        with torch.no_grad():
+            for module in self._module_tensor_id_mapping:
+                self._reset_quantizer(module)
+
+    def change_scale_impl_type(self, proxy) -> None:
```
pablomlago:
```suggestion
    def change_scale_impl_type(self, proxy: WeightQuantProxyFromInjectorBase) -> None:
```

src/brevitas/nn/utils.py
```diff
@@ -60,6 +70,117 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+            scaling_impl_type=ScalingImplType.PARAMETER_FROM_STATS)
+        proxy.init_tensor_quant()
+        proxy.load_state_dict(state_dict, strict=False)
+        config.IGNORE_MISSING_KEYS = ignore_missing_key
+        config.REINIT_ON_STATE_DICT_LOAD = reinit_on_state_dict
+
+    @staticmethod
+    def _reset_quantizer(proxy) -> None:
+        """Switch a weight quant proxy from LearnedRound back to standard Round."""
+        reinit_on_state_dict = config.REINIT_ON_STATE_DICT_LOAD
```
pablomlago: This pattern of overriding values in `config` and then restoring to the original values appears multiple times. Can we extract this common functionality? E.g.:
```
from contextlib import contextmanager
@contextmanager
def override_config(**overrides):
    old = {}
    try:
        for k, v in overrides.items():
            old[k] = getattr(config, k)
            setattr(config, k, v)
        yield
    finally:
        for k, v in old.items():
            setattr(config, k, v)
```
and then use it like:
```
with override_config(
        REINIT_ON_STATE_DICT_LOAD=False,
        IGNORE_MISSING_KEYS=True,
    ):
```

tests/brevitas/nn/test_merge_quant_weights.py
```diff
@@ -0,0 +1,152 @@
...
+from tests.conftest import SEED
+
+IN_FEATURES = 8
+OUT_FEATURES = 16
+
+LEARNED_ROUND_OPTIONS = [
+    LearnedRoundImplType.HARD_SIGMOID, LearnedRoundImplType.SIGMOID, LearnedRoundImplType.IDENTITY]
+
+
+def _insert_learned_round(model, learned_round_param):
```
pablomlago: I think there is no need to create a new function. Use `insert_learned_round_quantizers`.

tests/brevitas/nn/test_merge_quant_weights.py
```diff
@@ -0,0 +1,152 @@
...
+    model.eval()
+
+    x = torch.randn(4, IN_FEATURES)
+    with pytest.raises(RuntimeError), merge_quant_weights(model):
+        for _ in range(3):
+            model(x)
+
+
+@pytest.mark.parametrize("learned_round_param", LEARNED_ROUND_OPTIONS)
+def test_merge_quant_weights_reset(learned_round_param):
```
pablomlago: This test could be merged into the previous one.

tests/brevitas/nn/test_merge_quant_weights.py
```diff
@@ -0,0 +1,152 @@
...
+    x = torch.randn(4, IN_FEATURES)
+    with merge_quant_weights(model):
+        model(x)
+    assert isinstance(
+        model.weight_quant.tensor_quant.scaling_impl, ParameterFromStatsFromParameterScaling)
+    assert model.weight_quant.rounding_mode == "ROUND"
+
+
+@pytest.mark.parametrize("learned_round_param", LEARNED_ROUND_OPTIONS)
+def test_merge_quant_weights_forward_equivalence(learned_round_param):
```
pablomlago, line 87: Probably this test can also be merged into the previous one.

src/brevitas/nn/utils.py
```diff
@@ -1,9 +1,19 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from typing import Dict
```
pablomlago, line 5: Please remove unused imports.

pablomlago (changes requested):

src/brevitas/nn/utils.py
Giuseppe5: The learned round case is the most obvious one, but in general, merging the quant weight into the weights is a good way to save inference time, if needed.
I suspect it could make other exports easier (e.g., vLLM, where we export the original weights and then we quantize them at init time)

src/brevitas/nn/utils.py
Giuseppe5: I'll be more consistent, thanks for spotting.

tests/brevitas/nn/test_merge_quant_weights.py
Giuseppe5: That function is heavily tied to the class that it is attached to, unless I'm missing something?

src/brevitas/nn/utils.py
Giuseppe5: I'm going to open this as an issue and do it specifically in its own PR

tests/brevitas/nn/test_merge_quant_weights.py
pablomlago: It's a standalone method, it's not a method of the class.

src/brevitas/nn/utils.py
```diff
@@ -60,6 +69,106 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+            ``m.weight_orig`` before overwriting (default ``False``).
+    """
+    # Imported here to avoid a circular import
+    from brevitas.proxy.parameter_quant import WeightQuantProxyFromInjectorBase
+
+    hooks: List[RemovableHandle] = []
+    module_tensor_id_mapping: Dict = {}
+
+    def hook(
+            module: 'WeightQuantProxyFromInjectorBase', args: Tuple[Any, ...], output: Any) -> None:
```
pablomlago: Why not have the type hint with the class, instead of the literal `'WeightQuantProxyFromInjectorBase'`?

src/brevitas/nn/utils.py
```diff
@@ -60,6 +69,106 @@ def rename_state_dict_by_postfix(old_postfix, new_postfix, state_dict):
...
+
+    # Reset quantizers from LEARNED_ROUND to ROUND
+    with torch.no_grad():
+        for module in module_tensor_id_mapping:
+            _reset_quantizer(module)
+
+
+def _change_scale_impl_type(proxy) -> None:
+    """Change the scaling implementation type to PARAMETER_FROM_STATS."""
+    reinit_on_state_dict = config.REINIT_ON_STATE_DICT_LOAD
```
pablomlago, line 144: Can we extract this pattern of setting the values of `config` and restoring the original values into a context manager?

pablomlago (approved):
LGTM, feel free to ignore the following change.

PR #1508, opened 2026-04-10
Feat (ex/llm): custom quantizer plugin allows to modify quant model

Giuseppe5:
## Reason for this PR

In certain scenarios it is useful to have fine-grained control over what happens on your model once quantization has been  applied.

This could be useful to share quantizers in scenarios like exporting to vLLM, where for certain quantization configuration it is mondatory to have Q/K/V quantizers shared, as well as Gate Proj/Up Proj.

## Changes Made in this PR

Extend the current mechanism to define custom quantizer so that it also allows to define a custom function to modify the quant model.

If inheriting from BaseQuantizer, it is possible to only override one of the two methods while preserving the default behaviour (no-op) for the other one.

## Testing Summary

Added tests

### DIFF
--- a/src/brevitas_examples/common/generative/quantizers.py
+++ b/src/brevitas_examples/common/generative/quantizers.py
@@ -76,6 +76,10 @@ class BaseQuantizer:
                 quantizers_dict[key] = value
         return quantizers_dict
 
+    @classmethod
+    def modify_quantized_model(cls: "BaseQuantizer", model: nn.Module) -> nn.Module:
+        return model
+
 
 # Registry for custom quantizers
 QUANTIZERS_REGISTRY = Registry[Type[BaseQuantizer]](registry_name="QuantizersRegistry")
--- a/src/brevitas_examples/common/generative/quantizers.py
+++ b/src/brevitas_examples/common/generative/quantizers.py
@@ -77,7 +77,7 @@ class BaseQuantizer:
         return quantizers_dict
 
     @classmethod
-    def modify_quantized_model(cls: "BaseQuantizer", model: nn.Module) -> nn.Module:
+    def post_process_quant_model(cls: "BaseQuantizer", model: nn.Module) -> nn.Module:
         return model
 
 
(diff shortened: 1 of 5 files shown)


tests/brevitas_examples/llm_example_quantizer.py
```diff
@@ -7,3 +9,22 @@
...
+        return model
+
+
+@Registry.register(QUANTIZERS_REGISTRY, "example_quant_and_model_adjuster")
+class ExampleQuantAndModelAdjuster(BaseQuantizer):
+    weight_quant = Int8WeightPerTensorFloat.let(bit_width=4)
+
+    @classmethod
+    def modify_quantized_model(cls, model: nn.Module) -> nn.Module:
+        model.example_quant_and_model_adjuster_applied = True
```
nickfraser, line 31: Update test example to tie weight quantizers, like what is required for vLLM.

src/brevitas_examples/common/generative/quantizers.py
```diff
@@ -76,6 +76,10 @@ def override_quantizers_dict(
                 quantizers_dict[key] = value
         return quantizers_dict
 
+    @classmethod
+    def modify_quantized_model(cls: "BaseQuantizer", model: nn.Module) -> nn.Module:
```
pablomlago: I would change the name to `post_process_quantizers` or something along those lines.

src/brevitas_examples/llm/main.py
```diff
@@ -4,7 +4,7 @@
 from contextlib import nullcontext
 from copy import deepcopy
 import functools
-import importlib
+import importlib.util
```
pablomlago: Why is this required?

pablomlago (approved):
Some minor changes

src/brevitas_examples/llm/main.py
Giuseppe5: I guess it is a residue, I'll remove

PR #1509, opened 2026-04-10
Feat (quantize): functional quantization

Giuseppe5:
### 1. Motivation

This PR introduces a new `functional_quantization_mode` context manager that intercepts torch function calls at runtime via `TorchFunctionMode`, quantizes their positional arguments using lazily-created Brevitas quantizers, and cleans up all state on exit. No module replacement is needed. The model's original graph structure is preserved, and fused kernels remain fused.

This is useful for MoE quantization, as well as SDPA quantization.

For the LLM example pipeline, the old two-phase flow (Phase 1: insert `ScaledDotProductAttention` modules; Phase 3: wrap calibration in functional mode routing to those modules) is replaced by a single step that passes SDPA quantizer classes directly to the new context manager, including query, key, and value quantizers.

#### 2.1 Core implementation (`src/brevitas/graph/quantize.py`)

Three new classes are added inside the `torch_version >= 1.12` block:

- **`_WeightQuantHolder`**: A minimal `nn.Module` that holds a `.weight` attribute, enabling `WeightQuantProxyFromInjector` to be instantiated through the standard `QuantProxyMixin` path.
- **`_QuantParametrization`**: A `torch.nn.utils.parametrize`-compatible module that wraps a weight quant proxy and applies it during the forward pass.
- **`functional_quantization_mode`**: The main context manager (subclass of `TorchFunctionMode`).

#### 2.2 Interface

`functional_quantization_mode` accepts three arguments:

| Argument | Type | Description |
|---|---|---|
| `model` | `nn.Module` | The model whose forward pass will be intercepted. |
| `quant_map` | `Dict[Callable, spec]` | Mapping from torch functions to quantizer specifications. |
| `enabled` | `bool` | Whether quantization is active (default `True`). |

The `quant_map` values (`spec`) can take several forms:

**Single quantizer class** — quantize only the first positional argument:

```python
# Quantize the input (first arg) to F.linear with Int8 activation quantization.
# The weight (second arg) is left unquantized.
quant_map = {F.linear: Int8ActPerTensorFloat}
```

**Tuple of quantizer classes** — one per positional argument, any element may be `None`:

```python
# Quantize both input and weight of F.linear.
# First arg gets an activation quantizer; second arg is an nn.Parameter,
# so it gets a persistent weight parametrization via torch.nn.utils.parametrize.
quant_map = {F.linear: (Int8ActPerTensorFloat, Int8WeightPerTensorFloat)}
```

```python
# Quantize all three SDPA inputs: query, key, and value.
quant_map = {
    F.scaled_dot_product_attention: (q_quant, k_quant, v_quant)
}
```

```python
# Skip query quantization (None), quantize key and value.
quant_map = {
    F.scaled_dot_product_attention: (None, k_quant, v_quant)
}
```

```python
# Quantize both operands of torch.bmm (both are non-parameter tensors).
quant_map = {torch.bmm: (Int8ActPerTensorFloat, Int8ActPerTensorFloat)}
```

```python
# Disable quantization at runtime without changing the quant_map.
with functional_quantization_mode(model, quant_map, enabled=False):
    model(x)  # runs without quantization
```

#### 2.3 How arguments are quantized

Each positional argument is handled according to its position and nature:

| Position | Tensor type | Behavior |
|---|---|---|
| 0 (first arg) | Any tensor | A `QuantIdentity` activation quantizer is created and applied. |
| >= 1 | `nn.Parameter` | A weight quant proxy is created via `_WeightQuantHolder` and registered as a persistent `torch.nn.utils.parametrize` parametrization on the owning module. The parametrization persists across all forward passes within the context manager and is removed on exit. |
| >= 1 | Regular tensor | A `QuantIdentity` activation quantizer is created. If no explicit quantizer class is provided for this position, falls back to the first non-`None` class in the tuple. |
| >= 1 | `QuantTensor` | Skipped (already quantized). |
| Any | `None` in spec | Skipped (no quantizer created for this argument). |

#### 2.4 Lifecycle

```python
with functional_quantization_mode(model, quant_map) as ctx:
    # __enter__:
    #   1. Builds a parameter-ownership map (data_ptr -> (module, param_name))
    #   2. Attaches pre/post forward hooks to every module for module-stack tracking
    #   3. Attaches a top-level reset hook to clear per-module call counters after each forward
    #   4. Activates TorchFunctionMode interception

    model(calibration_data)
    # During forward:
    #   - Hooks push/pop the current module onto a stack
    #   - __torch_function__ intercepts calls in quant_map
    #   - Per-(module, func) call counter identifies which invocation this is
    #   - Quantizers are lazily created on first encounter, registered as submodules
    #     with _fq_ prefix, and reused on subsequent forwards

# __exit__:
#   1. Deactivates TorchFunctionMode
#   2. Removes all forward hooks
#   3. Removes all weight parametrizations (leaves original weights intact)
```

#### 2.5 Quantizer naming scheme

Quantizers are registered as submodules on the model with deterministic names:

```
_fq_{module_name}_{func_name}_{call_index}          # first arg (arg 0)
_fq_{module_name}_{func_name}_{call_index}_arg1      # second arg
_fq_{module_name}_{func_name}_{call_index}_arg2      # third arg
_fq_{module_name}_{func_name}_{call_index}_wq        # weight parametrization (arg 1)
_fq_{module_name}_{func_name}_{call_index}_arg2_wq   # weight parametrization (arg >= 2)
```

Dots in `module_name` (e.g., `model.layers.0.self_attn`) are replaced with underscores because `nn.Module.add_module` does not allow dots.

#### 2.6 LLM pipeline changes (`src/brevitas_examples/llm/main.py`)

- Removed the Phase 1 block that inserted `ScaledDotProductAttention` modules into the model graph.
- Removed the `ScaledDotProductAttention` import.
- The Phase 3 calibration context manager now extracts `q_scaled_quant`, `k_transposed_quant`, and `v_quant` directly from `quantizers_dict` and passes them as a 3-tuple:

```python
sdpa_q_quant = quantizers_dict.get('q_scaled_quant')
sdpa_k_quant = quantizers_dict.get('k_transposed_quant')
sdpa_v_quant = quantizers_dict.get('v_quant')

sdpa_quant_map = {
    torch.nn.functional.scaled_dot_product_attention:
        (sdpa_q_quant, sdpa_k_quant, sdpa_v_quant)
}
quantization_cm = functional_quantization_mode(model, sdpa_quant_map)
```

#### 3.1 Unit tests (`tests/brevitas/graph/test_functional_quantization_mode.py`)

27 tests covering all aspects of the context manager:

| Category | Tests | What is verified |
|---|---|---|
| **Basic operation** | `test_context_manager_basic`, `test_multiple_forward_passes`, `test_disabled_mode` | Context manager runs, multiple forwards work, `enabled=False` is a no-op. |
| **Hook lifecycle** | `test_hooks_removed_on_exit` | All hooks are removed when the context manager exits. |
| **Quantizer creation** | `test_quantizers_created`, `test_quantizers_registered_as_submodules` | Quantizers are lazily created, registered as `_fq_`-prefixed submodules. |
| **Counter management** | `test_counters_reset_after_forward` | Per-module call counters reset after each complete forward pass. |
| **Multi-module** | `test_two_linear_layers`, `test_multiple_calls_same_module` | Separate quantizers per module; correct counting when a module calls the same function multiple times. |
| **Passthrough** | `test_non_quantizable_function_passthrough` | Functions not in `quant_map` are unaffected. |
| **Output correctness** | `test_output_differs_from_unquantized` | Quantization changes the output (not a no-op). |
| **Single-class spec** | `test_single_quantizer` | `{func: class}` format only quantizes the first arg. |
| **Weight parametrization** | `test_second_quantizer_with_parameter`, `test_second_quantizer_creates_weight_quant_proxy`, `test_parametrization_removed_on_exit`, `test_parametrization_persistent_across_forwards`, `test_two_linear_with_second_quantizer`, `test_parametrization_uses_quant_parametrization_class` | Parameter args get `torch.nn.utils.parametrize` registrations; parametrization persists across forwards; removed on exit; uses `_QuantParametrization`. |
| **Non-parameter second arg** | `test_second_arg_non_parameter_uses_explicit_quant_type` | Non-parameter tensors (e.g., `torch.bmm` operands) get activation quantizers. |
| **QuantTensor skip** | `test_second_arg_already_quant_tensor_skipped` | Already-quantized tensors are not re-quantized. |
| **None quantizers** | `test_none_first_quantizer_skips_first_arg`, `test_none_first_with_explicit_second_quantizer` | `None` entries skip quantization of that argument while still quantizing others. |
| **SDPA 2-arg** | `test_sdpa_quantization_two_args` | Query + key quantization with a 2-tuple. |
| **SDPA 3-arg** | `test_sdpa_quantization_three_args` | Query + key + value quantization with a 3-tuple; verifies `_arg2` quantizer is created. |
| **SDPA selective** | `test_sdpa_none_query_quantize_key`, `test_sdpa_none_query_quantize_key_and_value` | None query with key-only or key+value quantization. |
| **Nested modules** | `test_nested_module_dots_in_name` | Dots in module names are safely handled. |

#### 3.2 Existing tests

- **41 tests** in `tests/brevitas/graph/test_quantize.py` continue to pass unchanged.
- **3 LLM functional SDPA tests** in `tests/brevitas_examples/test_llm.py` pass, updated to verify both `_arg1` (key) and `_arg2` (value) `QuantIdentity` submodules are created with the expected properties (unsigned, groupwise, correct group size).


Giuseppe5:
Add support for lambda-based quantizers (using the current module we're in, name and index of current quantizer that is being registered), and check whether this works or not with gradient checkpointing.
Check if triton call can be intercepted.

PR #1515, opened 2026-04-28
Feat (brevitas_examples/llm): support for FSDP2 trainer

Giuseppe5:
## Summary

Adds FSDP2 support for Brevitas LLM quantization workflows with fine-tuning. The change enables distributed training through Hugging Face Accelerate, restores a full trained model on rank 0 for post-training work, and keeps evaluation compatible with models that must be dispatched across multiple GPUs.
It also fixes parameter tracking in quantization proxies so they continue to observe the current model parameters after FSDP replaces them during sharding.

### FSDP2 Fine-Tuning

- Add FSDP-aware LLM fine-tuning through accelerate launch and Hugging Face Trainer.
- Support FSDP2 only, with explicit runtime validation:
- PyTorch >= 2.6
- Accelerate >= 1.14, < 2
- Transformers >= 5.15.1, < 6
- Validate that the selected trainer is a GeneralizedTrainer subclass and that Accelerate initialized FSDP2.
- Keep optimizer and scheduler construction compatible with FSDP2 DTensor parameter remapping.
- Prepare the distillation teacher through the FSDP2 path when distillation is enabled.
- Gather the trained full state dict after FSDP training and load it into a rank-0 model for evaluation, checkpointing, and export.
- End distributed training cleanly and terminate nonzero ranks before rank-0 post-training work.

### Live Quantizer Parameter Tracking

FSDP changes parameter identity during model preparation. It may replace a module's original nn.Parameter with an FSDP-managed parameter or DTensor-backed representation. A parameter list created before FSDP wrapping therefore becomes stale: it continues to hold references to the old parameters rather than the parameters that the model actually uses after wrapping.
This is particularly important for Brevitas parameter quantization proxies. Their tracked_parameter_list is used to locate the weights or biases associated with a quantizer. Returning a normal list such as:
[m.weight for m in self.tracked_module_list if m.weight is not None]
captures the parameter objects at the moment the property is evaluated. If FSDP subsequently replaces module.weight or module.bias, any retained list no longer follows the module's current parameter.
This PR replaces those eager lists with _TrackedParameterList:

```python
class _TrackedParameterList(Sequence):
    ...
```
The sequence stores weak references to the tracked modules and resolves weight or bias dynamically on every access.
This gives the proxy the following properties:
- It always returns the module's current parameter, including after FSDP replaces parameters.
- It does not hold obsolete parameter objects alive.
- It does not create strong-reference cycles between quantizer proxies and tracked modules.
- It retains the existing sequence-style behavior expected by callers: len(...), indexing, and iteration continue to work.
- It handles modules whose tracked parameter is absent or has become None.

The change is applied consistently to tracked weight and bias properties in ParameterQuantProxyFromInjector and its relevant subclasses.

### Evaluation and Multi-GPU Dispatch

- Add an FSDP-aware benchmark entrypoint that invokes main.py through accelerate launch.
- Add --job-folder result handoff so rank 0 can write results.json for the benchmark parent process.
- Extend GPU memory-map calculation to accept explicit device IDs.
- Preserve the rank-0 post-FSDP model as a multi-GPU dispatched model through offload_model.
LightEval previously always created its own Accelerate backend and called accelerator.prepare(model.to(device)). That is invalid for a model already dispatched by offload_model, because it can re-wrap or relocate a model that is intentionally distributed across GPUs.
The LightEval path now disables its internal Accelerate backend when the supplied model is already multi-device dispatched:
- FSDP post-training evaluation runs on rank 0 using the reconstructed model distributed across all local GPUs.
- Single-process, non-FSDP multi-GPU evaluation also retains its existing offload_model dispatch instead of collapsing to one GPU.
- Non-FSDP jobs launched with one process per GPU retain LightEval's normal Accelerate data-parallel evaluation behavior.
lm_eval requires no matching change. It receives the pre-initialized model directly, does not call Accelerator.prepare, and does not move the supplied model. It already relies on Brevitas' dispatch hooks for multi-GPU execution.

### Usage

Launch an FSDP2 job with an Accelerate configuration that specifies distributed_type: FSDP and fsdp_version: 2:
```bash
accelerate launch \
  --config_file path/to/accelerate_fsdp2_config.yaml \
  src/brevitas_examples/llm/main.py \
  --config path/to/llm_config.yaml \
  --fine-tune \
```
Run an FSDP benchmark:
```bash
python -m brevitas_examples.llm.benchmark.llm_fsdp_benchmark \
  --config path/to/benchmark.yaml \
  --accelerate-config path/to/accelerate_fsdp2_config.yaml \
  --gpus 0,1,2,3,4,5,6,7 \
  --num-gpus-per-process 8 \
  --results-folder ./results
```

### Notes
- FSDP activation checkpointing and Transformers gradient_checkpointing must not both be enabled.
- FSDP2 fine-tuning currently requires a GeneralizedTrainer subclass.
- Full validation requires a multi-GPU environment with the required FSDP2 dependency versions, model access, and evaluation datasets.


amd-tchouhan:
FSDP2 is to be launched directly with `brevitas_examples/llm/main.py` as:
`accelerate launch main.py --config <config.yaml>`

Before launching experiments, I set up `accelerate config` which looks like the followig:
```compute_environment: LOCAL_MACHINE
debug: false
distributed_type: FSDP
downcast_bf16: 'no'
enable_cpu_affinity: false
fsdp_config:
  fsdp_activation_checkpointing: true
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_cpu_ram_efficient_loading: false
  fsdp_offload_params: false
  fsdp_reshard_after_forward: true
  fsdp_state_dict_type: SHARDED_STATE_DICT
  fsdp_version: 2
machine_rank: 0
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

Note that `fsdp_activation_checkpointing` above and `gradient_checkpointing` in the yaml file cannot be set to true simultaneously.

PR #1518, opened 2026-05-11
Fix (brevitas_examples/llm): Better handling of thinking models

Giuseppe5:
## Reason for this PR

Lighteval does not handle well model with thinking enabled.
Similarly, there is an issue in how we handle `pad_token`, since they are always over-written to `pad_token`.

Now lighteval is pinned and it should be unpinned when they fix the issue linked below.

## Changes Made in this PR

This PR does two things:
- preserves `pad_token`
- Disables chat_template, which also disables thinking.

### MXFP8 Qwen 1.7B

Before Fix:

 'arc:challenge:0_acc': 32.08
 'arc:easy:0_acc': 43.45
 'hellaswag_lm_eval:0_acc': 42.83
 'winogrande:0_acc': 55.16

After Fix:

 'arc:challenge:0_acc': 39.76
 'arc:easy:0_acc': 71.42
 'hellaswag_lm_eval:0_acc': 46.06
 'winogrande:0_acc': 56.99

### BF16 Qwen3 0.7B

Before Fix:
 'arc:easy:0_acc':  38.67
'gsm8k:0_extractive_match': 12.59

After Fix:
 'arc:easy:0_acc':  60.48
'gsm8k:0_extractive_match': 52.16

### BF16 Llama 3.2 1B:

Before/After Fix:
 'arc:easy:0_acc':  66.41

### BF16 Llama 3.2 1B Instruct:

Before/After Fix:
'gsm8k:0_extractive_match': 35.94

cc: @pablomlago @i-colbert

### DIFF
--- a/src/brevitas_examples/llm/eval_lighteval.py
+++ b/src/brevitas_examples/llm/eval_lighteval.py
@@ -38,7 +38,9 @@ from lighteval.pipeline import ParallelismManager
 from lighteval.pipeline import Pipeline
 from lighteval.pipeline import PipelineParameters
 from lighteval.tasks.lighteval_task import LightevalTaskConfig
+from lighteval.tasks.prompt_manager import PromptManager
 from lighteval.tasks.requests import Doc
+from lighteval.tasks.requests import SamplingMethod
 from torch import nn
 from transformers import AutoTokenizer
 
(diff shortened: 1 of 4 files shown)


Giuseppe5:
Related issue:
https://github.com/huggingface/lighteval/issues/855

src/brevitas_examples/llm/eval_lighteval.py
```diff
@@ -129,6 +132,78 @@ def piqa_harness(line, task_name: str = None):
...
+    This subclass inspects ``doc.sampling_methods`` and routes accordingly:
+
+    * ``LOGPROBS`` / ``PERPLEXITY`` → plain-text formatting (no chat template).
+    * ``GENERATIVE`` → chat template with ``enable_thinking=False``.
+
+    For non-reasoning models the ``enable_thinking`` kwarg is silently ignored by Jinja2,
+    so this is safe to use unconditionally.
+    """
+
+    def prepare_prompt(self, doc: Doc) -> str:
```
pablomlago, line 160: I would try to keep the original code path as much as possible, except in the offending case, i.e.:
```
    def prepare_prompt(self, doc: Doc) -> str:
        if  SamplingMethod.GENERATIVE in doc.sampling_methods and self.use_chat_template:
            return self._prepare_chat_template_no_thinking(doc)
        return super().prepare_prompt(doc)
```

src/brevitas_examples/llm/eval_lighteval.py
```diff
@@ -129,6 +132,78 @@ def piqa_harness(line, task_name: str = None):
...
+        is_generative = SamplingMethod.GENERATIVE in doc.sampling_methods
+        if is_generative and self.use_chat_template:
+            return self._prepare_chat_template_no_thinking(doc)
+        else:
+            # For loglikelihood / perplexity tasks, always use plain text so
+            # that no thinking block or chat framing interferes with the
+            # probability computation over continuation tokens.
+            return self._prepare_plain_text(doc)
+
+    def _prepare_chat_template_no_thinking(self, doc: Doc) -> str:
```
pablomlago, line 170: A possible alternative that avoids duplicating the code:
```
def _prepare_chat_template_no_thinking(self, doc: Doc) -> str:
    """Format using the chat template with thinking mode explicitly disabled."""
    orig_apply = self.tokenizer.apply_chat_template
    try:
        self.tokenizer.apply_chat_template = partial(orig_apply, enable_thinking=False)
        return self._prepare_chat_template(doc)
    finally:
        self.tokenizer.apply_chat_template = orig_apply
```

pablomlago (changes requested):
Some changes might be required to avoid duplication.

src/brevitas_examples/llm/eval_lighteval.py
Giuseppe5: This is good

src/brevitas_examples/llm/eval_lighteval.py
Giuseppe5: This is not good. Calling super does not guarantee to call the plain_text version for log-likelihood tasks

pablomlago (approved):
LGTM (assuming test pass)

PR #1522, opened 2026-05-26
Feature/PWPolyF

ollycassidy13:
FINN needs a Brevitas-side PyTorch activation module that mirrors PWPolyF hardware behavior for common floating-point non-linear activations and can export a graph marker that FINN can lower to its RTL This PR adds a FINN-compatible piecewise-polynomial activation path for GELU, SiLU, sigmoid, and tanh, with QONNX export support through a `finn.pwpolyf::PWPolyF` custom-op marker.

The new activation:
  - Supports `gelu`, `silu`, `sigmoid`, and `tanh`.
  - Splits the FP32 input range into FINN/PWPolyF-style exponent and mantissa-derived segments.
  - Fits configurable per-segment polynomial coefficients.
  - Evaluates the approximation with Horner's method.
  - Applies the expected clamp/pass-through behavior for `|x| >= 8`.
  - Registers coefficients and clamp values as buffers so they are carried with the module.


src/brevitas/export/onnx/qonnx/manager.py
```diff
@@ -104,6 +109,7 @@ def export_onnx(
             onnx_export_kwargs[key][QONNX_DOMAIN_STRING] = QONNX_DOMAIN_VERSION
         else:
             onnx_export_kwargs[key] = {QONNX_DOMAIN_STRING: QONNX_DOMAIN_VERSION}
+        cls.solve_dynamo_export(onnx_export_kwargs)
```
nickfraser: Remove this.

src/brevitas/export/onnx/qonnx/manager.py
```diff
@@ -76,6 +76,11 @@ class QONNXManager(ONNXBaseManager):
         BrevitasFloatQuantFn,
         BrevitasQuantLSTMCellFn]
 
+    @classmethod
+    def solve_dynamo_export(cls, export_kwargs):
```
nickfraser: Remove this.

nickfraser (changes requested):
Sorry, but this PR seems to be mostly vibe-coded and needs a LOT of work before it would be accepted. I will hold off from doing a full review before asking the following:
 - do you really want to upstream this into Brevitas, rather than just using it as an out-of-source extension?
 - if yes, are you willing to take the time to rearchitect this?

If the answer is 'no' to either of the above - please close this PR.

PR #1526, opened 2026-06-09
fix (export/onnx): Update ONNX dynamo export for PT2.9 API

nickfraser:
# Support the PyTorch ≥2.9 dynamo ONNX exporter for QONNX and standard QCDQ

Adds support for the new `torch.export`-based ("dynamo") ONNX exporter for both the QONNX
and standard QCDQ export flows, while keeping the legacy TorchScript exporter working.
PyTorch flipped `torch.onnx.export` to default to `dynamo=True` in 2.9, so this also keeps
Brevitas's ONNX export working on recent PyTorch.

## Summary of changes

- **PT ≥2.9 API compatibility & routing** (`export/__init__.py`, ONNX managers): version-aware
  routing so bare `export_qonnx`/`export_onnx_qcdq` follow PyTorch's own default
  (`_DEFAULT_DYNAMO` for ≥2.9), never forward the `dynamo` kwarg on PyTorch <2.5 (where it
  doesn't exist), and dispatch to the correct (TorchScript vs dynamo) manager.
- **Unified op-emission pattern** (`onnx/function.py`, `onnx/handler.py`): a new `DynamoFn`
  base mirrors `torch.autograd.Function`, and `ONNXExportOpMixin.export_op(...)` selects the
  TorchScript backend (`g.op`) while tracing or the dynamo backend
  (`torch.onnx.ops.symbolic`) otherwise. Each op is now a `*TorchScriptFn` / `*DynamoFn` /
  `*Op` triple.
- **QONNX export** (`qonnx/function.py`, `qonnx/handler.py`, `qonnx/manager.py`): migrated to
  the triple pattern and **removed `qonnx/custom_ops.py`**, dropping the `onnxscript`
  wrappers, `torch.library.custom_op`s and the dynamo `custom_translation_table`. QONNX
  export no longer depends on `onnxscript`.
- **Standard QCDQ export** (`standard/function.py`, `standard/qcdq/handler.py`,
  `standard/qcdq/manager.py`): same triple pattern; new `StdQCDQONNXDynamoManager`;
  `assert_ge_zero` is gated to TorchScript tracing; unsupported-under-dynamo configurations
  raise a clear `RuntimeError` instead of failing cryptically deep in the trace.
- **Shared mixin** (`onnx/manager.py`): `ONNXDynamoExportMixin` de-duplicates the dynamo
  `set_export_mode` (QuantTensor disabling) and the export preamble (version assert, optimize
  warning) across the QONNX and QCDQ dynamo managers.
- **TruncQuant / avg-pool dynamo fix** (`common/handler/qcdq.py`): `signed`/`bit_width` are
  resolved to concrete values in `prepare_for_export`, avoiding data-dependent guards under
  `torch.export`.
- **Tests**: dynamo coverage added to the ONNX Runtime tests (`qcdq_dynamo`, `qonnx_dynamo`),
  FP8 + QONNX execution enabled, a small opset-parsing helper plus merged export/execute
  logic in the ORT helper, a latent test-only `rounding` extraction bug fixed, and new
  error-path tests (`tests/brevitas/export/test_onnx_qcdq_dynamo.py`).

## Risks & behavior changes

- **Default exporter flip (user-facing):** on PyTorch ≥2.9, bare `export_qonnx(...)` /
  `export_onnx_qcdq(...)` now use the dynamo exporter (matching PyTorch). For QCDQ
  configurations that the dynamo path does not support, this raises — callers must pass
  `dynamo=False` (legacy TorchScript) or use a supported configuration.
- **QONNX dynamo is broadly supported; QCDQ dynamo is a subset** (see Limitations).
- **`onnxscript` is no longer _directly_ required** for QONNX export.
- **Test-only:** non-dynamo QCDQ round-weight cases now also exercise the Q-node
  (`QuantizeLinear`) weight-export path. The library default (`export_weight_q_node=False`)
  is unchanged.

## Limitations

- **QCDQ + dynamo** does not support (these fail fast with a clear error, and are skipped in
  the tests): **quantized bias**, **A2Q / input-aware decoupled weights**,
  **integer-initializer weight export** (`export_weight_q_node=False`), **non-round weight
  rounding**, and **LSTM**. Root causes: integer export relies on `Tensor.data_ptr()` which
  is unsupported on `FakeTensor` under `torch.export`, and `QuantizeLinear` only implements
  round-to-nearest-even. TBD for details. QONNX export is
  unaffected (its `Quant`/`FloatQuant` ops operate on the float tensor and carry a rounding
  attribute).
- The dynamo paths require **PyTorch ≥ 2.8** (for `torch.onnx.ops`).
- Under QCDQ dynamo, weights are exported as `QuantizeLinear`+`DequantizeLinear` from the
  float weight (a different graph shape than the integer-initializer + `DequantizeLinear`
  form produced by the legacy path).

## Validation

QONNX export unit tests, FP8 export unit tests, the new dynamo error-path tests, and the
ONNX Runtime sweeps across `qcdq`, `qcdq_dynamo`, `qonnx`, `qonnx_dynamo` (including
avg-pool and LSTM where applicable) pass, with unsupported configurations skipped with
explicit reasons. Currently, the tests have only been run locally, since the CI only tests PyTorch<2.8, which isn't supported.


src/brevitas/export/onnx/qonnx/function.py
```diff
@@ -22,11 +26,27 @@ def symbolic(g, x, scale, zero_point, bit_width, narrow_range, signed, rounding_
 
     @staticmethod
     def forward(ctx, x, scale, zero_point, bit_width, narrow_range, signed, rounding_mode):
-        x = bipolar_quant(x, scale)
+        x = binary_sign(x) * scale
         return x
 
 
-class BrevitasQuantFn(Function):
+class BrevitasBinaryQuantDynamoFn(DynamoFn):
```
Giuseppe5, line 33: I would probably separate the Dynamo classes in a separate file, and group together all the "general" classes (like `BrevitasBinaryQuantOp` below); maybe have all of them defined at the end of this file or in a new file? Mostly because they are 3 lines and easy to miss when editing.

We can leave the torchscript here as this is the legacy file, and see what happens (maybe we'll stop supporting it at some point).

nickfraser: I prefer to keep these close together, so that it clear that they need to change together if the are any API changes in newer PyTorch versions.

src/brevitas/export/common/handler/qcdq.py
```diff
@@ -50,6 +50,12 @@ def itemize_quantize_scalar_params(self):
         pass
 
     def assert_ge_zero(self, *args):
+        # This is a workaround to trick the TorchScript tracer into believing all
+        # return values are used. Under torch.export (dynamo) the comparison
+        # becomes a data-dependent guard and is both unnecessary and unsupported,
+        # so we skip it.
+        if not torch.jit.is_tracing():
+            return
```
nickfraser: A bit "clunky", consider putting everything under: `if torch.jit.is_tracing():`

src/brevitas/export/onnx/qonnx/function.py
```diff
@@ -44,18 +64,35 @@ def symbolic(g, x, scale, zero_point, bit_width, narrow_range, signed, rounding_
...
-            bit_width,
-            int(narrow_range),
-            signed=int(signed),
-            rounding_mode=rounding_mode)
-        return y
+        float_to_int_impl = solve_float_to_int_impl_from_enum(rounding_mode)
+        quant = IntQuant(
+            float_to_int_impl=float_to_int_impl(),
+            tensor_clamp_impl=TensorClamp(),
+            input_view_impl=Identity(),  #TODO: Update this when QONNX support Groupwise export
```
nickfraser: Typo fix.

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +87,64 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
                 handler._export_q_node = export_weight_q_node
+
+
+class StdQCDQONNXDynamoManager(StdQCDQONNXManager):
+    run_onnx_passes = False  # Skip the optimization step from onnxoptimizer. False required to keep ONNX metadata
+    onnx_passes = ["eliminate_unused_initializer"]
+    custom_fns = []
+
+    @classmethod
+    def set_export_mode(cls, model: Module, enabled: bool):
```
nickfraser: Looks like shared logic with `QONNXDynamoManager`. Consider moving to a shared mixin?

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +87,64 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+            model._brevitas_return_quant_tensor_state = return_quant_tensor_state
+        else:
+            enable_quant_tensor = partial(_override_create_quant_tensor, state=False)
+            model.apply(enable_quant_tensor)
+            QuantizationStatusManager.restore_return_quant_tensor(
+                model, model._brevitas_return_quant_tensor_state)
+            del model._brevitas_return_quant_tensor_state
+
+    @classmethod
+    def _validate_dynamo_supported(cls, module: Module):
```
nickfraser: Do these options also fail during QONNX export? If so, should this exist as a shared Mixin?

tests/brevitas/export/test_onnx_fp8.py
```diff
@@ -6,24 +6,24 @@
...
-def test_simple_fp8_export():
+def test_simple_fp8_export(request, qcdq_export_fn):
     if torch_version < version.parse('2.1.0'):
         pytest.skip(f"OCP FP8 types not supported by {torch_version}")
 
     model = qnn.QuantLinear(3, 16, weight_quant=Fp8e4m3OCPWeightPerTensorFloat)
-    outfile = f'weight_fp8.onnx'
-    export_onnx_qcdq(model, torch.randn(1, 3), outfile, export_weight_q_node=True)
+    outfile = f'weight_fp8_{request.node.callspec.id}.onnx'
+    qcdq_export_fn(model, torch.randn(1, 3), outfile, export_weight_q_node=True, opset_version=19)
```
nickfraser: Why does the opset version need to be specified here?

tests/brevitas/export/test_onnx_fp8.py
```diff
@@ -42,25 +42,25 @@ def test_qonnx_simple_fp8_export(request, qonnx_export_fn):
...
-def test_fp8_export_activation():
+def test_fp8_export_activation(request, qcdq_export_fn):
     if torch_version < version.parse('2.1.0'):
         pytest.skip(f"OCP FP8 types not supported by {torch_version}")
 
     model = qnn.QuantLinear(3, 16, input_quant=Fp8e4m3OCPActPerTensorFloat)
-    outfile = f'act_fp8.onnx'
-    export_onnx_qcdq(model, torch.randn(1, 3), outfile, export_weight_q_node=True)
+    outfile = f'act_fp8_{request.node.callspec.id}.onnx'
+    qcdq_export_fn(model, torch.randn(1, 3), outfile, export_weight_q_node=True, opset_version=19)
```
nickfraser: Why does the opset version need to be specified here?

tests/brevitas/export/test_onnx_fp8.py
```diff
@@ -42,25 +42,25 @@ def test_qonnx_simple_fp8_export(request, qonnx_export_fn):
...
+def test_fp8_export_export_activation(request, qcdq_export_fn):
     if torch_version < version.parse('2.1.0'):
         pytest.skip(f"OCP FP8 types not supported by {torch_version}")
 
     model = qnn.QuantLinear(
         3, 16, weight_quant=Fp8e4m3OCPWeightPerTensorFloat, input_quant=Fp8e4m3OCPActPerTensorFloat)
-    outfile = 'weight_act_fp8.onnx'
-    export_onnx_qcdq(model, torch.randn(1, 3), outfile, export_weight_q_node=True)
+    outfile = f'weight_act_fp8_{request.node.callspec.id}.onnx'
+    qcdq_export_fn(model, torch.randn(1, 3), outfile, export_weight_q_node=True, opset_version=19)
```
nickfraser: Why does the opset version need to be specified here?

tests/brevitas_ort/common.py
```diff
@@ -144,11 +144,11 @@ def is_brevitas_ort_close(
         computed_out = brevitas_output
         scale = 1.
 
-    if tolerance is not None and export_type == 'qcdq':
+    if tolerance is not None and export_type in ('qcdq', 'qcdq_dynamo'):
         tolerance = tolerance * scale  # Float Output, tolerance is +/- output scale
 
     if export_type == 'qonnx':
-        exported_model = export_qonnx(model, input_t, export_path=export_name)
+        exported_model = export_qonnx(model, input_t, export_path=export_name, dynamo=False)
```
nickfraser: Any reason why `qonnx_dynamo` is not supported?

tests/brevitas_ort/test_quant_module.py
```diff
@@ -22,7 +22,7 @@
 
 
 @parametrize_with_cases('model', cases=QuantWBIOLCases)
-@pytest.mark.parametrize('export_type', ['qcdq', 'qonnx'])
+@pytest.mark.parametrize('export_type', ['qcdq', 'qcdq_dynamo', 'qonnx'])
```
nickfraser: Should we also test `'qonnx_dynamo'`?

tests/brevitas_ort/test_quant_module.py
```diff
@@ -86,6 +98,9 @@ def test_ort_wbiol(model, export_type, current_cases):
 @parametrize_with_cases('model', cases=QuantAvgPoolCases)
 @requires_pt_ge('1.10')
 def test_ort_avgpool(model, current_cases):
+    # NOTE: TruncAvgPool / TruncQuant export is not supported under the dynamo
+    # (torch.export) QCDQ path: the handler branches on the (traced) `signed`
+    # tensor, which triggers a data-dependent guard. Kept torchscript-only.
```
nickfraser: Should we instead fix the handler? Signed should be fixed at export time, even if it is propagated through a `QuantTensor`. Can we cache this and convert to a boolean to fix the tracing issue?

How does this work with QONNX export? We should probably copy that design pattern if it solves this problem already.

nickfraser (commented):
See below - several issues to address:

src/brevitas/export/common/handler/qcdq.py
nickfraser: Fixed in [b1580d9](https://github.com/Xilinx/brevitas/pull/1526/commits/b1580d9e84a6a4bca3623789c1a9fa2735d072a8).

src/brevitas/export/onnx/qonnx/function.py
nickfraser: Fixed in Fixed in [b1580d9](https://github.com/Xilinx/brevitas/pull/1526/commits/b1580d9e84a6a4bca3623789c1a9fa2735d072a8).

tests/brevitas_ort/test_quant_module.py
```diff
@@ -39,14 +39,29 @@ def test_ort_wbiol(model, export_type, current_cases):
...
+        # The dynamo (torch.export) QCDQ path only supports configs that don't rely on
+        # data_ptr()-keyed integer export (i.e. quantized bias / integer-initializer
+        # weights). In the WBIOL cases bias is quantized for everything except the fp8
+        # and dynamic-activation quantizers, so we limit dynamo coverage to those.
+        if torch_version < parse('2.8'):
+            pytest.skip('QCDQ dynamo export requires PyTorch >= 2.8')
+        if 'fp8' not in quantizer and 'dynamic' not in quantizer:
+            pytest.skip('QCDQ dynamo export does not support quantized bias (data_ptr export).')
+        # Integer weight export (data_ptr) is unsupported under dynamo; force Q-node weights.
+        export_q_weight = True
```
nickfraser: Instead of setting `export_q_weight = True`, we should skip this case (with appropriate message), since `QuantizeLinear` doesn't handle different round modes at the moment.

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Addressed in [114a573](https://github.com/Xilinx/brevitas/pull/1526/commits/114a57323a58fbbce8dc0885268be5115f18e774).

tests/brevitas_ort/common.py
nickfraser: `qonnx_dynamo` tests added in [8e58a88](https://github.com/Xilinx/brevitas/pull/1526/commits/8e58a88ed43edd364ce02eca82cf87e8b7db17b9).

tests/brevitas_ort/test_quant_module.py
nickfraser: qonnx_dynamo tests added in [8e58a88](https://github.com/Xilinx/brevitas/pull/1526/commits/8e58a88ed43edd364ce02eca82cf87e8b7db17b9).

tests/brevitas_ort/test_quant_module.py
nickfraser: Updated to cache the required values to constants in `self.prepare_for_export` in [745c341](https://github.com/Xilinx/brevitas/pull/1526/commits/745c341830ef748c396e5d704b2b95cfc9f25c06).

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,36 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+        # Integer weight/bias export relies on `data_ptr()`, which is incompatible with
+        # torch.export (FakeTensor). Surface a clear error for the configurations that
+        # would otherwise fail cryptically deep inside the trace.
+        for m in module.modules():
+            if isinstance(m, DecoupledWeightQuantWithInputProxyFromInjector) and m.is_quant_enabled:
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` does not support input-aware decoupled "
+                    "weight quantization (e.g. A2Q): integer weight export uses data_ptr(), "
+                    "which is unsupported under torch.export.")
+            if isinstance(m, BiasQuantProxyFromInjector) and m.is_quant_enabled:
```
nickfraser, line 96: Need to investigate if this is really necessary, or only when `export_weight_q_node=False`

src/brevitas/export/onnx/manager.py
```diff
@@ -214,3 +221,32 @@ def export(
...
+
+
+class ONNXDynamoExportMixin:
+    # Shared configuration and helpers for the torch.export (dynamo) ONNX managers.
+    run_onnx_passes = False  # keep ONNX metadata; skip the onnxoptimizer step
+    onnx_passes = ["eliminate_unused_initializer"]
+    custom_fns = []
+
+    @classmethod
+    def set_export_mode(cls, model: Module, enabled: bool):
```
nickfraser, line 234: Add comment about moving this to a generic `DynamoExportMixin` class, but need to be careful about possible circular dependency.

tests/brevitas_ort/common.py
```diff
@@ -161,9 +163,19 @@ def is_brevitas_ort_close(
...
+        elif export_type == 'qcdq_dynamo':
+            export_onnx_qcdq(
+                model,
+                input_t,
+                export_path=export_name,
+                export_weight_q_node=export_q_weight,
+                opset_version=onnx_opset,
+                dynamo=True,
+                optimize=True)
         elif export_type == 'qonnx_opset14':
```
nickfraser: Consider creating a private helper method that detects `opset<Version>` substring and adds that argument as necessary. Could be leveraged for the earlier comments too.

tests/brevitas_ort/test_quant_module.py
```diff
@@ -39,14 +39,29 @@ def test_ort_wbiol(model, export_type, current_cases):
...
     if 'per_channel' in quantizer and 'asymmetric' in quantizer:
         pytest.skip('Per-channel zero-point is not well supported in ORT.')
     if 'QuantLinear' in impl and 'asymmetric' in quantizer:
         pytest.skip('ORT execution is unreliable and fails randomly on a subset of cases.')
     if 'dynamic' in quantizer and ((o_bit_width != "o8" or i_bit_width != "i8") or
-                                   export_type != "qcdq"):
+                                   export_type not in ("qcdq", "qcdq_dynamo")):
         pytest.skip('Dynamic Act Quant supported only for 8bit and QCDQ export')
-    if export_type == 'qonnx' and 'fp8' in quantizer:
+    if export_type in ('qonnx', 'qonnx_dynamo') and 'fp8' in quantizer:
```
nickfraser: Double-check if latest QONNX version support FP8 export. If so, let's enable these tests again.

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Follow-up question in later review.

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Will create an issue after merging this PR.

tests/brevitas_ort/test_quant_module.py
nickfraser: Addressed in [6d80dd2](https://github.com/Xilinx/brevitas/pull/1526/commits/6d80dd21d5deaf0647b9cb7634ee79145244a418).

src/brevitas/export/onnx/manager.py
nickfraser: Addressed in [1c9a817](https://github.com/Xilinx/brevitas/pull/1526/commits/1c9a8179c290a5113b6c17c353f030d2625c638f).

tests/brevitas_ort/common.py
nickfraser: Addressed in [12d6180](https://github.com/Xilinx/brevitas/pull/1526/commits/12d618046d6dc158e073a27ae33a202bd3809da8).

tests/brevitas_ort/test_quant_module.py
nickfraser: Addressed in [5c98fee](https://github.com/Xilinx/brevitas/pull/1526/commits/5c98fee4aede72abaa07a8053c5ccced6d4d0531).

tests/brevitas/export/test_onnx_fp8.py
nickfraser: DRY implemented in [5c98fee](https://github.com/Xilinx/brevitas/pull/1526/commits/5c98fee4aede72abaa07a8053c5ccced6d4d0531).

tests/brevitas/export/test_onnx_fp8.py
nickfraser: DRY implemented in [5c98fee](https://github.com/Xilinx/brevitas/pull/1526/commits/5c98fee4aede72abaa07a8053c5ccced6d4d0531).

tests/brevitas/export/test_onnx_fp8.py
nickfraser: DRY implemented in [5c98fee](https://github.com/Xilinx/brevitas/pull/1526/commits/5c98fee4aede72abaa07a8053c5ccced6d4d0531).

src/brevitas/export/common/handler/qcdq.py
```diff
@@ -50,11 +50,16 @@ def itemize_quantize_scalar_params(self):
...
     def assert_ge_zero(self, *args):
-        for a in args:
-            bools = a >= 0.
-            if isinstance(bools, torch.Tensor):
-                bools = bools.all()
-            assert bools
+        # This is a workaround to trick the TorchScript tracer into believing all
+        # return values are used. Under torch.export (dynamo) the comparison
+        # becomes a data-dependent guard and is both unnecessary and unsupported,
+        # so we only run it while tracing.
```
nickfraser: Rewrite comment to be more concise.

src/brevitas/export/common/handler/qcdq.py
```diff
@@ -797,17 +802,42 @@ def prepare_for_export(self, module: TruncQuantProxyFromInjector):
         if module.is_quant_enabled:
             assert not self.export_fake_quantized, "Activation quantization does not support fake quantization export"
             self.validate(module)
+            # Signedness and bit-width are propagated through the input QuantTensor; reading
+            # them during tracing yields data-dependent `.item()` calls that torch.export
+            # (dynamo) cannot guard on. Resolve everything that branches on them here
+            # (before tracing) into concrete values / precomputed kwargs, so that
+            # symbolic_execution only applies tensor ops. This mirrors how the activation
+            # and weight QCDQ handlers precompute their quantize/clip/dequantize kwargs.
```
nickfraser: Rewrite comment to be more concise.

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,36 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
                 handler._export_q_node = export_weight_q_node
+
+
+class StdQCDQONNXDynamoManager(ONNXDynamoExportMixin, StdQCDQONNXManager):
+
+    @classmethod
+    def _validate_dynamo_supported(cls, module: Module, export_weight_q_node: bool):
+        # Integer weight/bias export relies on `data_ptr()`, which is incompatible with
+        # torch.export (FakeTensor). Surface a clear error for the configurations that
+        # would otherwise fail cryptically deep inside the trace.
```
nickfraser: Remove comment about "Surface clear error"

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,36 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+        for m in module.modules():
+            if isinstance(m, DecoupledWeightQuantWithInputProxyFromInjector) and m.is_quant_enabled:
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` does not support input-aware decoupled "
+                    "weight quantization (e.g. A2Q): integer weight export uses data_ptr(), "
+                    "which is unsupported under torch.export.")
```
nickfraser: Remove sentences about `data_ptr`. Leave this for an issue which describes the limitation.

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,36 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+            if isinstance(m, BiasQuantProxyFromInjector) and m.is_quant_enabled:
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` does not support quantized bias: integer "
+                    "bias export uses data_ptr(), which is unsupported under torch.export.")
```
nickfraser: Remove sentences about `data_ptr`. Leave this for an issue which describes the limitation.

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,36 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+            if (not export_weight_q_node and isinstance(m, WeightQuantProxyFromInjector) and
+                    m.is_quant_enabled):
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` requires `export_weight_q_node=True` for "
+                    "quantized weights: integer weight export uses data_ptr(), which is "
```
nickfraser: Remove sentences about `data_ptr`. Leave this for an issue which describes the limitation.

src/brevitas/export/onnx/manager.py
```diff
@@ -214,3 +221,37 @@ def export(
...
+        # NOTE: this QuantTensor-disabling logic is not ONNX-specific and could be hoisted
+        # into a generic (backend-agnostic) DynamoExportMixin shared with other
+        # torch.export-based exporters. Deferred for now because such a mixin must import
+        # QuantizationStatusManager (brevitas.graph.calibrate) and _override_create_quant_tensor
+        # (brevitas.export.inference.manager) without introducing an import cycle.
```
nickfraser: Make more concise.

tests/brevitas_ort/test_quant_module.py
```diff
@@ -22,32 +22,54 @@
...
     o_bit_width = case_id.split('-')[-6]
     i_bit_width = case_id.split('-')[-4]
     onnx_opset = 14
     export_q_weight = False
 
-    if rounding == 'round':
+    # Round weights can be faithfully exported as a Q-node (QuantizeLinear); non-round
+    # (floor) weights can only be represented via integer-initializer export. A2Q
+    # (accumulator-aware) weight quant always requires integer-initializer export, so it is
+    # excluded from Q-node export.
```
nickfraser: Make more concise.

src/brevitas/export/common/handler/qcdq.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/common/handler/qcdq.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/onnx/manager.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

tests/brevitas_ort/test_quant_module.py
nickfraser: Addressed in [e115b70](https://github.com/Xilinx/brevitas/pull/1526/commits/e115b702e225ab5dd59c0cb91210c4667d4d795b).

src/brevitas/export/common/handler/qcdq.py
```diff
@@ -50,11 +50,14 @@ def itemize_quantize_scalar_params(self):
...
 
     def assert_ge_zero(self, *args):
-        for a in args:
-            bools = a >= 0.
-            if isinstance(bools, torch.Tensor):
-                bools = bools.all()
-            assert bools
+        # Workaround to make the TorchScript tracer treat all return values as used.
+        # Skipped under torch.export (dynamo), where it would be a data-dependent guard.
+        if torch.jit.is_tracing():
```
Giuseppe5, line 55: In other places, we use a different, internal API to check for this.
Unfortunately it was not documented back then the reason of that vs this one.
I'd rather keep this, but I'm flagging the inconsistency for future works.

src/brevitas/export/onnx/standard/qcdq/manager.py
```diff
@@ -76,3 +80,32 @@ def change_weight_export(cls, export_weight_q_node: bool = False):
...
+        # Integer weight/bias export relies on `data_ptr()`, which is incompatible with
+        # torch.export (FakeTensor).
+        for m in module.modules():
+            if isinstance(m, DecoupledWeightQuantWithInputProxyFromInjector) and m.is_quant_enabled:
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` does not support input-aware decoupled "
+                    "weight quantization (e.g. A2Q).")
+            if isinstance(m, BiasQuantProxyFromInjector) and m.is_quant_enabled:
+                raise RuntimeError(
+                    "QCDQ export with `dynamo=True` does not support quantized bias.")
```
Giuseppe5, line 98: Until opset 19, QuantizeLinear could only quantize to int8, for this reason we would rather export a pre-quantized bias (e.g., quantized to INT16).
[Starting from opset 21](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#quantizelinear-21), QuantizeLinear also supports int16. This means that we can also think about having a q_node boolean flag for bias, similar to what we do for weights.

I'll leave this for reference but it would be a future work, if the need arise.

tests/brevitas_ort/test_quant_module.py
```diff
@@ -22,32 +22,52 @@
...
-@pytest.mark.parametrize('export_type', ['qcdq', 'qonnx'])
+@pytest.mark.parametrize('export_type', ['qcdq', 'qcdq_dynamo', 'qonnx', 'qonnx_dynamo'])
 @requires_pt_ge('1.10')
 def test_ort_wbiol(model, export_type, current_cases):
     cases_generator_func = current_cases['model'][1]
     case_id = get_case_id(cases_generator_func)
-    rounding = case_id.split('-')[0]
-    impl = case_id.split('-')[
-        -3]  # Inverse list of definition, 'export_type' is -1, 'impl' is -2, etc.
+    # case_id has the form
```
Giuseppe5, line 30: Thanks

src/brevitas/export/onnx/standard/qcdq/handler.py
```diff
@@ -26,20 +26,20 @@
...
+from ..function import DynamicQuantizeLinearOp
+from ..function import IntClipOp
+from ..function import QuantizeLinearOp
 
 
 class StdDQCastONNXMixin(DQCastMixin, ABC):
 
     def dequantize_fn(self, x, scale, zero_point, axis):
-        return DequantizeLinearFn.apply(x, scale, zero_point, axis)
+        return self.export_op(DequantizeLinearOp, x, scale, zero_point, axis)
```
Giuseppe5, line 40: The method `export_op` has not been defined by any of the parent classes (this applies also to the other classes).
We might want to have have a separate Mixin (abstract class) that defined this as abstract method. If we inherit from that Mixin higher enough in the chain, it should be a minimal change that could solve the issue for everyone afterwards.

Giuseppe5 (approved):
Only one minor conceptual change, the rest are comments to keep track of potential future work or past inconsistencies. 

The compatibility between dynamo QCDQ and A2Q is trickier  than I anticipated, but that also requires a separate discussion and PR.

src/brevitas/export/common/handler/qcdq.py
nickfraser: Shall I make an issue flagging this?

src/brevitas/export/onnx/standard/qcdq/manager.py
nickfraser: Shall I make an issue flagging this as well?

src/brevitas/export/common/handler/qcdq.py
Giuseppe5: Yes please, thank you!

src/brevitas/export/onnx/standard/qcdq/manager.py
Giuseppe5: Yes, I don't believe it's worth adding this change here since it involves a bit of work to handle int vs int16 vs int8 which we currently do not do.

src/brevitas/export/onnx/standard/qcdq/handler.py
nickfraser: Actually, a concrete implementation is provided in the ABC, [`ONNXBaseHandler`](https://github.com/nickfraser/brevitas/blob/e115b702e225ab5dd59c0cb91210c4667d4d795b/src/brevitas/export/onnx/handler.py#L141).

Do you want me to add an abstract class in `BaseHandler`? I personally wasn't sure that this approach would be required (or even make sense) in other export flows. Or alternatively, I could create an issue to note that we should investigate this as future work.

Giuseppe5: The issue is that in this particular example, `StdDQCastONNXMixin` is calling `self.export_op` which is not defined by this class, nor by any of the other class from which `StdDQCastONNXMixin` inherits.

Usually, we try to have at least abstracmethods for everything that might be defined later on through inheritance and we are breaking this pattern here.

My suggestion would be something like:

```python
class ExportOpMixin(ABC):
    
     # The function signature is tricky since it changes wildly based on who calls this
     # so we keep it simple
     @abstractmethod
     def export_op(self):
          pass

```
Then we would have `StdDQCastONNXMixin` or anyone other class in its MRO to also inherit from that Mixin.

The main thing I'm not sure is, where to inject that Mixin so that it covers all the use cases without causing failures or unnecessary clutter in the MRO.

nickfraser: Resolved in [6216e2e](https://github.com/Xilinx/brevitas/pull/1526/commits/6216e2e4de56220f41354cd025dc6e64478739f3). The only open question remains: should we move the concrete implementation to this class?

nickfraser: I moved the implementation as well in [808c98e](https://github.com/Xilinx/brevitas/pull/1526/commits/808c98e5cbb1f343a5d9c3ba0d5c78223e1f7fe4). Please revert it if you prefer to keep it in`ONNXBaseHandler`.

PR #1533, opened 2026-06-17
Fix (brevitas_examples/llm): patch dynamo FX  with torch 2.10+

Giuseppe5:
## Reason for this PR

In torch 2.10 and above, `torch._dynamo.export` stops creating `call_module` and flattens everything to `call_function`.

## Changes Made in this PR

This PR does two things:
- uses `install_free_tensors_for_export` to restore `call_module` in the FX graph
- Patches a torch function that is broken in torch 2.10 and torch 2.11 (fixed in torch 2.12)


Tested locally with:
```python
import torch
import torch.nn as nn
import torch._dynamo.config as dynamo_config
import inspect, textwrap                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
import torch._dynamo.output_graph as og                                                                                                                                                                                                                                                                                                                                 
from torch._dynamo.variables import CONSTANT_VARIABLE_NONE                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                        
_orig = og.OutputGraph._get_stack_values_to_restore                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                        
def _patched(self, tx, stack_pops):                                                                                                                                                                                                                                                                                                                                     
    stack_values, meta = _orig(self, tx, stack_pops)                                                                                                                                                                                                                                                                                                                    
    stack_values = [CONSTANT_VARIABLE_NONE if v is None else v for v in stack_values]                                                                                                                                                                                                                                                                                   
    return stack_values, meta                                                                                                                                                                                                                                                                                                                                           
                                                                                                         
# This is needed only for torch 2.10 and torch 2.11                                                                                                                                                                                                                                         
og.OutputGraph._get_stack_values_to_restore = _patched         

class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.module1 = nn.Linear(3, 16)
    
    def forward(self, x):
        return self.module1(x)

model = Model()
with dynamo_config.patch(install_free_tensors_for_export=False):                                                                                                                                                                                                                                                                                                        
         fx_export, _ = torch._dynamo.export(model, torch.randn(1, 3))
print(fx_export.graph.print_tabular())


```


Related to https://github.com/pytorch/pytorch/issues/174481

### DIFF
--- a/src/brevitas_examples/llm/main.py
+++ b/src/brevitas_examples/llm/main.py
@@ -85,7 +85,7 @@ except:
     logging.debug("Shark-AI not installed, cannot export to Shark")
 
 
-def _maybe_patch_dynamo_export_for_torch_2_10_2_11():
+def _maybe_patch_dynamo_export():
     # torch._dynamo.export in torch 2.10 and 2.11 can crash with
     # "'NoneType' object has no attribute 'is_tensor'" when a raw None is left on
     # Dynamo's symbolic stack: OutputGraph.compile_subgraph calls x.is_tensor()
--- a/src/brevitas_examples/llm/main.py
+++ b/src/brevitas_examples/llm/main.py
@@ -12,11 +12,13 @@ import sys
 import warnings
 
 import numpy as np
+from packaging import version
 import torch
 from torch.utils.data import DataLoader
 from transformers import AutoModelForCausalLM
 from transformers import AutoTokenizer
 
+from brevitas import torch_version
 from brevitas.export.inference.manager import quant_inference_mode
 from brevitas.export.onnx.standard.qcdq.manager import StdQCDQONNXManager
 from brevitas.graph import load_quant_model_mode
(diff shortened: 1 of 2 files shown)


src/brevitas_examples/llm/main.py
```diff
@@ -83,6 +85,47 @@
     logging.debug("Shark-AI not installed, cannot export to Shark")
 
 
+def _maybe_patch_dynamo_export():
```
pablomlago: Is it possible for this code to live under `brevitas_examples/common`? It would avoidclutter in the LLM entrypoint and I imagine that it could be reused in other entrypoints.

pablomlago (changes requested):
A minor comment regarding where this code should live:

PR #1534, opened 2026-06-19
Setup: unpin transformers

Giuseppe5:
# Support latest `transformers` (5.x) in the LLM examples

## Motivation

The LLM examples (`src/brevitas_examples/llm`) were only compatible with
`transformers <= 4.57.x`. With `transformers` unpinned, every LLM test in
`tests/brevitas_examples/test_llm.py` and `test_llm_data.py` failed against
`transformers 5.x` — in most cases the test suite could not even be *collected*.

This PR makes the LLM examples work with the latest `transformers` (validated on
`5.12.1`) while keeping compatibility with `>=4.52` wherever the upstream
libraries allow it. It covers all four LLM test jobs: `-m llm`, `-m onnx_export`,
`-m few_shot` (both `lm_eval` and `lighteval`), and `test_llm_data.py`.

## Summary of failures and fixes

The breakages fall into two buckets: **transformers 5.x API changes** in our own
code, and **third-party dependency incompatibilities**.

### 1. transformers 5.x API changes (brevitas source)

| File | Change | Why it was needed |
|------|--------|-------------------|
| `llm_quant/data.py` | `tokenizer.batch_encode_plus(...)` → `tokenizer(...)` | `batch_encode_plus` was removed in 5.x. Calling the tokenizer directly with a list of texts is the equivalent API and also works on 4.x. |
| `llm_quant/rotation_optimization.py` | `Trainer(tokenizer=...)` → `Trainer(processing_class=...)` | The `tokenizer` argument to `Trainer` was renamed to `processing_class` (added in 4.46, removed in 5.x). |
| `main.py` | Version-guarded `torch_dtype` → `dtype` kwarg in `from_pretrained` | `torch_dtype` was renamed to `dtype` in 4.56 and removed in 5.x. We select the kwarg name based on the installed version so `>=4.52` keeps working. |
| `llm_quant/prepare_for_quantize.py` | Enable `generation_config.use_cache=True` before wrapping with `TorchExportableModuleForDecoderOnlyLM`, restore on exit | 5.x's exportable module now *asserts* that caching is enabled in the constructor. This affects **every FX path** (`weight_equalization`, `act_equalization=fx`, `rotation=fx`, `quant_sdpa=fx`, `ln_affine_merge`, `convert_layernorm_to_rmsnorm`). We force-enable it only for the duration of tracing (the cache is disabled on the unwrapped model during the trace itself) and restore the original value afterwards to avoid leaking state into later generation/eval.

### 2. Removal of the `optimum` dependency
------------------------------------------------------------------------------------------------------------------------------------------------------------------
**_This has been reverted, so that optimum is installed for ONNX export and transformers < 5.0 is used.
The following description and the relevant commit are left here for the future, if we want to get rid of optimum onnx._**
------------------------------------------------------------------------------------------------------------------------------------------------------------------

`optimum` was used in two places, both of which broke on `transformers 5.x`:

- **Top-level import (`NormalizedConfigManager`)** in `data_utils.py` and
  `run_utils.py`. The installed `optimum` imports `is_tf_available` from
  `transformers.utils`, which was removed in 5.x, so the import failed at module
  load time and **blocked collection of the entire LLM test module**.
  `NormalizedConfigManager` was only being used to read a handful of standard HF
  config fields, so it was replaced with direct config attribute access
  (`num_attention_heads`, `num_key_value_heads`, `head_dim`/`hidden_size`, with
  `get_text_config()` support for nested/multimodal configs). The unused
  `modify_dataloader` helper in `run_utils.py` was removed along with its import.

- **ONNX export (`onnx_export_from_model`)** in `main.py` — see the next section.

After these changes, **`optimum` is no longer imported anywhere** in the LLM
examples, so it was dropped from `requirements-llm.txt` and from the
`optimum[onnxruntime]` install in the `noxfile.py` export job.

#### Why optimum is no longer needed for ONNX export

It's worth being explicit here, because at first glance removing optimum from the
ONNX path looks risky.

In the old code, optimum's role in the export was actually thin. The
quantization-aware part of the export was already done by **brevitas**, not
optimum:

```python
with brevitas_proxy_export_mode(model, export_manager=export_manager):
    onnx_export_from_model(model, ..., task="text-generation-with-past")
```

- `brevitas_proxy_export_mode` is what swaps the quant proxies into their QCDQ
  (`QuantizeLinear`/`DequantizeLinear`) export handlers and emits the quantized
  ONNX nodes. This is the only genuinely brevitas-specific logic, and it is
  unchanged.
- `onnx_export_from_model`, *once inside that context*, was essentially a
  convenience wrapper around `torch.onnx.export`: it assembled the model inputs
  (including KV-cache for `text-generation-with-past`), chose the opset and
  dynamic axes, and wrote `model.onnx`.

So everything optimum contributed here is reproducible directly, and the new code
calls `torch.onnx.export` ourselves inside `brevitas_proxy_export_mode`.

The **forcing reason** to remove it: `optimum 2.x` moved ONNX export into a
separate `optimum-onnx` package, and `optimum-onnx 0.1.0` hard-pins
`transformers<4.58.0`. Installing it silently downgrades `transformers` back to
`4.57.6`, which is incompatible with the goal of this PR. Since optimum provided
no irreplaceable functionality in this path, removing it both unblocks
`transformers 5.x` and drops a heavyweight, fragile dependency.

#### The new native ONNX export

The replacement in `main.py` performs a direct `torch.onnx.export` inside
`brevitas_proxy_export_mode`, with four adaptations required by torch / transformers 5.x:

1. **Device consolidation.** When the model has been offloaded with `accelerate`
   (the default in the eval path), parameters can live on CPU while the cached
   activation-quantization metadata needed for QCDQ export lives on the execution
   device (e.g. `cuda:0`). `model.to(...)` does not move those cached
   (non-buffer) tensors, which caused cross-device errors during tracing. We
   detect the device of the cached activation tensors and align the whole model
   to it before exporting.
2. **Explicit attention mask.** transformers 5.x's masking code calls
   `torch.diff` to detect packed sequences whenever `attention_mask is None`.
   `torch.diff` is unsupported by the TorchScript ONNX tracer. Passing an
   explicit `attention_mask` of ones skips that path.
3. **`opset_version=14`.** Required to export `scaled_dot_product_attention`,
   which the 5.x attention implementations use (it's only supported from opset
   14).
4. **`use_cache=False` during export** so the traced graph has a simple
   `input_ids -> logits` signature that ONNX can represent (restored afterwards).

**Trade-off / follow-up:** the native export currently traces a plain
`input_ids -> logits` graph and drops the `text-generation-with-past`
(KV-cache I/O) graph that optimum produced. This is sufficient for the test
(which only checks that `model.onnx` loads) and for static QCDQ inspection, but
if a downstream consumer specifically needs a with-past decoder graph, that would
need to be re-added.

### 3. `few_shot` dependency upgrades

- **`lm_eval`**: the previously installed `0.4.9.2` references
  `transformers.AutoModelForVision2Seq`, which was removed in 5.x (renamed to
  `AutoModelForImageTextToText`). Upgrading to `lm_eval >= 0.4.10` (validated with
  `0.4.12`) resolves this. The `noxfile.py` `lm_eval` job now pins
  `lm_eval>=0.4.10`.
- **`lighteval`**: the pinned `0.13` already imports and runs fine on
  `transformers 5.x`; no change required.

## Test changes

- **`tests/brevitas_examples/test_llm_data.py`**: `DummyTokenizer.__call__` now
  accepts either a list of strings (batch) or a single string, mirroring the
  behaviour of `PreTrainedTokenizerBase.__call__` that replaces the removed
  `batch_encode_plus`. The internal `batch_encode_plus` helper is kept so the
  expected-value tables are unchanged.
- **`tests/brevitas_examples/test_llm_cases.py`**: updated 6 golden `quant_ppl`
  values for the `rotation_optimization` cases. These tests use a deliberately
  tight tolerance (`atol=2.0`) on absolute perplexities (~30k), and the numerics
  of the 5.x model propagate through the rotation-optimization training, causing
  a small (~3–35) but deterministic drift. The float perplexities and all layer
  counts are unchanged; only the optimized quantized perplexities were
  re-baselined. (Note: like the pre-existing values, these are
  hardware/transformers-version specific.)

## Packaging changes

- `requirements/requirements-llm.txt`: removed `optimum` (no longer used; it was
  forcing `transformers<5.0`).
- `noxfile.py`:
  - export job no longer installs `optimum[onnxruntime]`;
  - `lm_eval` job pins `lm_eval>=0.4.10`.

## Compatibility

- All brevitas source fixes are compatible with `transformers >= 4.52`
  (`tokenizer(...)`, direct config attribute access, `processing_class`, and the
  version-guarded `dtype` kwarg all work on 4.x and 5.x).
- The only raised lower bound is `lm_eval>=0.4.10`.
- `optimum` is dropped entirely; the ONNX export no longer depends on it for any
  `transformers` version.

### DIFF
--- a/requirements/requirements-llm-onnx-export.txt
+++ b/requirements/requirements-llm-onnx-export.txt
@@ -1,5 +1,8 @@
 # Optional dependencies for ONNX QCDQ export of LLMs (brevitas_examples/llm).
-# `optimum` (via `optimum-onnx`) provides `onnx_export_from_model`. We install the
-# highest available version; note that `optimum-onnx` currently constrains
-# `transformers` to < 4.58, so this extra is not compatible with transformers >= 5.0.
+# `optimum` (via `optimum-onnx`) provides `onnx_export_from_model`. This workflow is
+# not compatible with transformers >= 4.58: newer optimum splits the ONNX exporter into
+# `optimum-onnx` (which pins transformers < 4.58), while older optimum (pre-split) still
+# imports symbols removed in transformers 5.x. We therefore pin transformers < 4.58
+# explicitly so the resolver reliably selects a compatible optimum + transformers pair.
 optimum[onnxruntime]
+transformers<4.58
--- a/requirements/requirements-llm-onnx-export.txt
+++ b/requirements/requirements-llm-onnx-export.txt
@@ -1,8 +1,5 @@
-# Optional dependencies for ONNX QCDQ export of LLMs (brevitas_examples/llm).
-# `optimum` (via `optimum-onnx`) provides `onnx_export_from_model`. This workflow is
-# not compatible with transformers >= 4.58: newer optimum splits the ONNX exporter into
-# `optimum-onnx` (which pins transformers < 4.58), while older optimum (pre-split) still
-# imports symbols removed in transformers 5.x. We therefore pin transformers < 4.58
-# explicitly so the resolver reliably selects a compatible optimum + transformers pair.
+# Optional ONNX QCDQ export dependencies for brevitas_examples/llm.
+# optimum-onnx (pulled in by optimum[onnxruntime]) pins transformers < 4.58, so we
+# declare that bound explicitly to get a deterministic resolution.
 optimum[onnxruntime]
 transformers<4.58
--- a/requirements/requirements-llm-onnx-export.txt
+++ b/requirements/requirements-llm-onnx-export.txt
@@ -1,0 +1,5 @@
+# Optional dependencies for ONNX QCDQ export of LLMs (brevitas_examples/llm).
+# `optimum` (via `optimum-onnx`) provides `onnx_export_from_model`. We install the
+# highest available version; note that `optimum-onnx` currently constrains
+# `transformers` to < 4.58, so this extra is not compatible with transformers >= 5.0.
+optimum[onnxruntime]
--- a/src/brevitas_examples/common/dynamo_utils.py
+++ b/src/brevitas_examples/common/dynamo_utils.py
@@ -32,11 +32,8 @@ def dynamo_export_ctx():
     # transformers 5.x checks `arg_name in func.__code__.co_varnames` inside model
     # forward wrappers. torch._dynamo < 2.7 cannot trace __contains__ on a code
     # object descriptor, raising Unsupported. Fail early with a clear message.
-    try:
-        import transformers as _transformers
-        _tr_ver = version.parse(_transformers.__version__)
-    except ImportError:
-        _tr_ver = version.parse("0")
+    import transformers as _transformers
+    _tr_ver = version.parse(_transformers.__version__)
 
     if torch_version < version.parse("2.7") and _tr_ver >= version.parse("5.0"):
         raise RuntimeError(
--- a/src/brevitas_examples/common/dynamo_utils.py
+++ b/src/brevitas_examples/common/dynamo_utils.py
@@ -7,13 +7,9 @@ from brevitas import torch_version
 
 
 def patch_dynamo_export():
-    # torch._dynamo.export in torch 2.10 and 2.11 can crash with
-    # "'NoneType' object has no attribute 'is_tensor'" when a raw None is left on
-    # Dynamo's symbolic stack: OutputGraph.compile_subgraph calls x.is_tensor()
-    # over those stack values. Upstream fixed this in 2.12 by replacing raw None
-    # on the stack with ConstantVariable(None) (pytorch/pytorch#169325). We
-    # backport that behaviour out-of-source for 2.10/2.11 only by coercing raw
-    # None to a ConstantVariable when the stack values are gathered.
+    # torch 2.10/2.11 crash with "'NoneType' has no attribute 'is_tensor'" when a
+    # raw None is left on the dynamo stack (fixed in 2.12 via pytorch#169325).
+    # Backport: coerce raw None to ConstantVariable(None) when gathering stack values.
     if not (version.parse('2.10') <= torch_version < version.parse('2.12')):
         return
     from torch._dynamo.output_graph import OutputGraph
@@ -33,14 +29,9 @@ def patch_dynamo_export():
 
 
 def dynamo_export_ctx():
-    # Guard: torch < 2.7 combined with transformers >= 5.0 is incompatible for
-    # FX-based dynamo export.  In transformers 5.x the @merge_with_config_defaults
-    # decorator accesses func.__code__.co_varnames inside the model forward wrapper
-    # (``if arg_name in func.__code__.co_varnames``).  torch._dynamo on torch < 2.7
-    # cannot trace __contains__ on a GetSetDescriptorVariable (the co_varnames code
-    # object descriptor), raising torch._dynamo.exc.Unsupported.  This was fixed in
-    # torch 2.7.  Raise a clear RuntimeError early rather than surfacing an obscure
-    # dynamo traceback.
+    # transformers 5.x checks `arg_name in func.__code__.co_varnames` inside model
+    # forward wrappers. torch._dynamo < 2.7 cannot trace __contains__ on a code
+    # object descriptor, raising Unsupported. Fail early with a clear message.
     try:
         import transformers as _transformers
         _tr_ver = version.parse(_transformers.__version__)
@@ -54,11 +45,7 @@ def dynamo_export_ctx():
             f"Found torch {torch_version}, transformers {_transformers.__version__}. "
             f"Please upgrade torch to >= 2.7.")
 
-    # From torch 2.10 onwards, torch._dynamo.export inlines built-in nn modules
-    # (install_free_tensors_for_export=True) instead of emitting call_module
-    # nodes. Setting install_free_tensors_for_export=False routes them back
-    # through the specialized NNModuleVariable path, restoring the pre-2.10 graph
-    # structure. The flag does not exist before torch 2.10.
+    # torch >= 2.10 inlines nn modules by default; restore call_module behaviour.
     if torch_version >= version.parse('2.10'):
         import torch._dynamo.config as dynamo_config
         return dynamo_config.patch(install_free_tensors_for_export=False)
--- a/src/brevitas_examples/common/dynamo_utils.py
+++ b/src/brevitas_examples/common/dynamo_utils.py
@@ -1,9 +1,129 @@
-from contextlib import nullcontext
+from contextlib import contextmanager
 import functools
 
 from packaging import version
 
 from brevitas import torch_version
+
+
+def _patch_transformers_co_varnames():
+    """
+    Patch transformers 5.x's `merge_with_config_defaults` decorator so its
+    wrapper does not access ``func.__code__.co_varnames`` at call time.
+
+    In transformers 5.x a ``@merge_with_config_defaults`` wrapper checks
+    ``if arg_name in func.__code__.co_varnames:`` on every call.  On torch
+    ≤ 2.6, torch._dynamo cannot trace the ``__contains__`` operation on a
+    ``GetSetDescriptorVariable`` (the co_varnames descriptor), which causes
+    ``torch._dynamo.exc.Unsupported`` for every FX-path test.  The fix is to
+    precompute the ``arg_index`` mapping *once* at decoration time (outside the
+    wrapper) so the wrapper only operates on plain Python values that dynamo
+    can handle.
+
+    Returns True if the patch was applied, False if it was not needed.
+    """
+    try:
+        import transformers.utils.generic as tug
+    except ImportError:
+        return False
+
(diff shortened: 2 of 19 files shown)


Giuseppe5:
The 6 re-baselined `quant_ppl` values all belong to the **`rotation_optimization`** cases, which *train* the rotation matrices before measuring perplexity.
Transformers 5.x slightly changed the model's forward numerics (notably the attention/masking implementation); those tiny differences alter the gradients, hence the trained rotations, hence the final quantized perplexity.

Giuseppe5:
it looks like wikitext2 breaks, must investigate, and add tests.

src/brevitas_examples/llm/main.py
```diff
@@ -181,22 +183,44 @@ def set_seed(seed):
...
-        # Local import to allow for optional install
-        from optimum.exporters.onnx import onnx_export_from_model
+        # Optional dependency; constrains transformers < 4.58 (see llm_onnx_export extra).
+        try:
+            from optimum.exporters.onnx import onnx_export_from_model
+        except ImportError as e:
+            raise ImportError(
+                "ONNX QCDQ export of LLMs requires `optimum`, which is an optional "
+                "dependency. Install it with `pip install brevitas[llm_onnx_export]` "
+                "(note: this currently constrains `transformers` to < 4.58).") from e
```
nickfraser: Slight concern: I think this failure triggers at export time - we might also want to add something to `validate` so that we "fail early" as well?

src/brevitas_examples/llm/main.py
```diff
@@ -181,22 +183,44 @@ def set_seed(seed):
...
 
+        # Align the model to the device of the cached activation scales. When offloaded
+        # with accelerate, parameters may be on CPU but cached scales on the execution
+        # device; .to() does not move non-buffer tensors, so we derive the target device
+        # from the scales rather than model.parameters().
+        export_device = None
+        for m in model.modules():
+            cached_act = getattr(m, "_cached_act", None)
+            if cached_act is not None and getattr(cached_act, "scale", None) is not None:
+                export_device = cached_act.scale.device
```
nickfraser: Is this something we might do with other export flow? I.e., is it worth putting this code into a helper?

nickfraser (commented):
Only a few comments. Let me know when the Wikitext2 fix comes in - it might be worth having another set of eyes on that.

src/brevitas_examples/llm/llm_args.py
```diff
@@ -545,6 +545,14 @@ def validate(args: Namespace, extra_args: Optional[List[str]] = None) -> None:
         args.fine_tune = True
     if not args.fine_tune:
         assert extra_args is None or len(extra_args) == 0, f"The following unknown arguments were passed: {[extra_arg for extra_arg in extra_args if extra_arg.startswith('--')]}"
+    if args.export_target == 'onnx_qcdq':
+        try:
+            from optimum.exporters.onnx import onnx_export_from_model
+        except ImportError as e:
+            raise ImportError(
+                "ONNX QCDQ export of LLMs requires `optimum`, which is an optional "
+                "dependency. Install it with `pip install \"brevitas[llm,export]\"`.") from e
```
nickfraser: Is this install command right? Isn't it: `pip install brevitas[export,llm,llm-onnx-export]`? Or some variant?

src/brevitas_examples/llm/main.py
```diff
@@ -179,9 +181,19 @@ def set_seed(seed):
     torch.random.manual_seed(seed)
 
 
+def align_model_to_cached_act_device(model):
```
nickfraser: Does this imply that there is an underlying issue with how accelerate is interacting with our scale? If so, if it worth considering a fix for that underlying issue?

This looks like a workaround (which might still be the right thing for this PR).

src/brevitas_examples/common/dynamo_utils.py
```diff
@@ -33,11 +29,23 @@ def _get_stack_values_to_restore(self, tx, stack_pops):
...
+    # transformers 5.x checks `arg_name in func.__code__.co_varnames` inside model
+    # forward wrappers. torch._dynamo < 2.7 cannot trace __contains__ on a code
+    # object descriptor, raising Unsupported. Fail early with a clear message.
+    try:
+        import transformers as _transformers
+        _tr_ver = version.parse(_transformers.__version__)
+    except ImportError:
+        _tr_ver = version.parse("0")
+
+    if torch_version < version.parse("2.7") and _tr_ver >= version.parse("5.0"):
```
nickfraser, line 38: As far as I understand this code: if I want to trace a simple linear layer on PyTorch 2.8, I'll get an error if transformers>=5.0 happens to be installed.

Maybe we should separate concerns (pytorch version, transformers version) here and consider having 2 separate functions for this, if transformers modules require special attention.

requirements/requirements-llm-onnx-export.txt
```diff
@@ -0,0 +1,5 @@
+# Optional ONNX QCDQ export dependencies for brevitas_examples/llm.
+# optimum-onnx (pulled in by optimum[onnxruntime]) pins transformers < 4.58, so we
+# declare that bound explicitly to get a deterministic resolution.
+optimum[onnxruntime]
```
nickfraser, line 4: There's a section in the PR doc about removing this requirement. Do we want to do that in this PR?

tests/brevitas_examples/llm/test_llm.py
```diff
@@ -143,6 +144,19 @@ def wrapper_main(
                    mock_load_raw_dataset):
             # Validate the arguments before running the entrypoint
             validate_args(parser, args)
+
+            # torch < 2.7 + transformers >= 5.0: FX export is unsupported (see dynamo_export_ctx).
+            # Assert the guard fires, then skip.
+            use_fx = fx_required(args) or args.rotation == 'fused_no_fx'
```
nickfraser: Possible unrelated issue to this PR, but if `fx_required(args)` is sufficient to know if FX is required, then shouldn't we fix that function? Or is there some reason not to that I don't recall?

nickfraser (commented):
A few comments that (might) be worth addressing. Also, looks like there's merge conflicts with `master` - if there's any serious conflicts, then maybe this needs more attention.

requirements/requirements-llm-onnx-export.txt
Giuseppe5: No, the thing is that there is a work-around to get rid of this requirement but the final ONNX graph we'd get with the workaround would be different from one we get through optimum-onnx export.
In particular, the ONNX graph with optimum-onnx is technically more precise, since it includes both phases of the LLM lifecycle, whilst the one we'd get without optimum-onnx would not (but it would still represent where the quantization nodes are placed).

I am not sure at the moment whether we need the entire LLM lifecycle represented in the ONNX graph, so I am leaving this decision for another PR.

src/brevitas_examples/common/dynamo_utils.py
Giuseppe5: No, maybe the description has to be adjusted.
torch 2.8 is the minimum version required to trace entire models on transformers 5.0.

If you try to trace a simple linear layer, there is no issue with any torch version. 

So we cannot really separate concerns.

tests/brevitas_examples/llm/test_llm.py
Giuseppe5: `fx_required` maybe is a bit of a misnomer. It indicates whether the HF model is used in its FX form for forward passes, ptq calibration, etc.

`fused_no_fx` also performs an FX transformation which is then thrown away, but for the purpose of the test it is still relevant.

src/brevitas_examples/llm/llm_args.py
Giuseppe5: Fixed

src/brevitas_examples/llm/main.py
Giuseppe5: The issue was that the act_cache might be populated by someone else and then left here stale and read during export inadvertently.

Now optimum-onnx will use an export path similar to vLLM, where we enter export_mode and cache is re-populated.

Giuseppe5: A limitation, not addressed in this PR since it is out of scope, is that ONNX export performs cpu-based inference; the issue is that `accelerate`'s offload_model could break the ONNX tracing. It is going to be addressed in the future.

src/brevitas_examples/common/dynamo_utils.py
nickfraser: Oh sorry - I see this is already a transformers-only workaround.

tests/brevitas_examples/llm/test_llm.py
nickfraser: Maybe consider 2 helpers for these 2 cases - if there is something that makes sense. Please go ahead!

nickfraser (approved):
Please double-check that SpinQuant works on some example, then merge 👍

PR #1535, opened 2026-06-26
CI: Test broader set of torch versions

Giuseppe5:


Giuseppe5:
#1536 should deal with the failures on on torch 2.11 and 2.12

PR #1536, opened 2026-06-28
Fix (tests/permute): use symbolic_trace instead of dynamo

Giuseppe5:
## Reason for this PR

For some reasons, we use `_dynamo.export` for some tests, however this is not (very) compatible with latest torch versions, unless with flags and patching.

For testing purposes, we have always used `symbolic_trace`, so we stick to that for consistency whilst we figure out how to proceed with FX graphs moving forward.

`_dynamo.export` is to be preferred to `symbolic_trace` only when dealing with big models and complex set of operators.


pablomlago (approved):

PR #1537, opened 2026-06-29
Feat (gguf): broaden format support and reproduce canonical recipes

Giuseppe5:
## Reason for this PR

This PR extends Brevitas GGUF support to add: (1) K-quants like Q2_K, Q3_K, etc, (2) nested scale
quantization, and (3) recipe-specific mixed precision. Users can now evaluate a GGUF recipe in
Brevitas and export the same quantized model.

## Changes Made in this PR

- Added GGUF quantizers from Q2_K through Q8_0. The K-quant implementations model nested scales
  and zero points. MSE scale search improves alignment with llama.cpp.
- Added registered quantizer plugins and file-based recipe plugins. Llama 3.2 1B and 3B recipes
  apply the required mixed-precision tensor layouts.
- Added `gguf:<type>` export targets and target validation. The exporter reads cached quantizer
  metadata, preserves the selected GGUF tensor types, and accepts an exact output path.
- Added GGUF usage, configuration, and validation documentation in `src/brevitas_examples/papers/gguf`.
- Removed Q6_K quantizer from `src/brevitas_examples/llm/gguf_export/quant.py` so that `q6_k_quant_block` only handles packing, just like the other implementations.

## Testing Summary

The table compares Brevitas exports with fresh `llama-quantize` outputs for
Llama-3.2-1B-Instruct. Size values use decimal MB.

| Recipe | Brevitas Size (MB) | Ref. Size (MB) |
|---|---:|---:|
| Q8_0 | 1321.1 | 1321.1 |
| Q6_K | 1021.8 | 1021.8 |
| Q5_K_M | 911.5 | 911.5 |
| Q5_K | 892.6 | 892.6 |
| Q4_1 | 831.7 | 831.7 |
| Q4_K_M | 807.7 | 807.7 |
| Q4_K_S | 775.6 | 775.6 |
| Q4_0 | 773.0 | 773.0 |
| Q4_K | 770.9 | 770.9 |
| Q3_K_L | 732.5 | 732.5 |
| Q3_K_M | 690.8 | 690.8 |
| Q3_K_S | 641.7 | 641.7 |
| Q2_K | 580.9 | 580.9 |

All 13 pairs have matching tensor names, quantization types, and shapes. File sizes differ by at
most 352 bytes because of GGUF metadata and header padding.

The reference flow uses an importance matrix when the llama.cpp recipe requires one. The Q5_K row
uses the llama.cpp Q5_K_S recipe. The Q4_K row uses Q4_K_S with
`--pure --token-embedding-type q6_k`. These options reproduce the Brevitas tensor layouts.

Interoperability with [Qronos](https://xilinx.github.io/brevitas/v0.13.0/papers/qronos.html) was also verified. The table below reports `llama-perplexity` results for GGUF models exported by Brevitas. Round-to-nearest (RTN) provides a baseline where each weight is rounded to the nearest supported value on a fixed grid.

| Recipe | Algorithm | 1B | 3B |
|---|---|---:|---:|
| Q2_K | RTN | 30.60 | 15.04 |
| Q2_K | Qronos | 18.17 | 12.60 |

The Qronos results above use the provided configurations for Llama-3.2-1B-Instruct and Llama-3.2-3B-Instruct.

## Risk Highlight

- [x] This PR includes code from another work. The GGUF layouts and mixed-precision rules adapt
  behavior from llama.cpp. The recipe files include the applicable MIT license.
- [x] This PR contains API-breaking changes. `--quantize-first-last-layer` replaces
  `--quantize-last-layer`. The `save_quantized_as_gguf` function signature also changes.
- [x] This PR depended on work in PR #1585 to support injecting a custom `scale_shift_zero_point_impl` into `ParameterFromStatsFromParameterZeroPoint `

### Follow-up TODOs

1. Generalize the nested quantizer cache in Brevitas core, either by adding a new nested cache or a new tensor subclass. This would render the out-of-source caching infrastructure redundant.
2. Add a registry for out-of-source `InferenceManager` handlers. This registry can provide a supported extension path for the GGUF handler.
3. Add activation quantization emulation. The common configuration uses dynamic 8-bit quantization, but activation quantization formats can vary by backend.


src/brevitas_examples/llm/gguf_export/convert.py
```diff
@@ -431,7 +439,9 @@ def quantize(self, name, data, data_qtype, old_dtype, bid):
 
         # TODO: Generalize this to have a map between GGUF quant type
         # and our preprocessing for quant_modules
-        if data_qtype == gguf.GGMLQuantizationType.Q4_K:
+        # Q4_K and Q5_K share the same super-block structure (8 sub-blocks of 32,
+        # 6-bit nested scales/mins, fp16 d/dmin); only the code bit-width differs.
+        if data_qtype in (gguf.GGMLQuantizationType.Q4_K, gguf.GGMLQuantizationType.Q5_K):
```
i-colbert: Does a registry make sense here, as we add support for more types?

Giuseppe5: Yes if we start adding support for new data-type, it could be worth thinking about a structure that wires data-type with export handling

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,438 @@
...
+    @property
+    def gguf_qtype(self):
+        return self.quant_injector.gguf_qtype
+
+
+class _GGUFGroupwiseIntWeightInferenceHandler(GroupwiseIntWeightInferenceHandler):
+    handled_layer = GGUFGroupwiseWeightQuantProxyFromInjector
+
+
+InferenceManager.handlers.append(_GGUFGroupwiseIntWeightInferenceHandler)
```
Giuseppe5: Let's add a comment here to mark that this is temporary

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,453 @@
...
+    @enabled.setter
+    def enabled(self, value):
+        if not value:
+            self.value = None
+            self.scale = None
+        self._enabled = value
+
+
+# TODO: temporary workaround to cache hierarchical quantizers for GGUF export; should revisit
+class _GGUFCachedIOGroupwiseInt(_CachedIOGroupwiseInt):
```
Giuseppe5: Let's do the following changes:
- Caching classes and export override in a separate file called brevitas_export_extension.py
- Proxies in a new file called brevitas_proxy_extension.py

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,453 @@
...
+    def scale(self):
+        # (possibly quantized) scale for each sub-group
+        return self.retrieve_attribute('scale_')
+
+    @property
+    def zero_point(self):
+        # (possibly quantized) zero-point for each sub-group; see Q5_KWeightQuant for example
+        # NOTE: for K-quants this no longer matches `quant_weight.zero_point_`; simple types are
+        # unaffected, since they fall through to the branch below.
+        zero_point = self.retrieve_attribute('zero_point')
```
Giuseppe5: There could be a way for the Caching class to expose an attribute called `zero_point_` which actually contains `zero_point`, so that the proxy shouldn't be aware of this difference. Mostly aesthetic though.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,453 @@
...
+        out, scale_of_scale, *_ = self.float_to_int_impl(x)
+        if self.gguf_scaling_cache.enabled:
+            int_threshold = self.int_scaling_impl(self.bit_width_impl())
+            self.gguf_scaling_cache.scale = (scale_of_scale / int_threshold).detach()
+        if self.scale_dequantized_shape is not None:
+            out = out.view(self.scale_dequantized_shape)
+        return out
+
+
+class _GGUFCachedScaleShiftQuantZeroPoint(_ScaleShiftQuantZeroPoint):
```
Giuseppe5, line 163: name is a bit misleading, since it is not caching or using anything that is cached

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,453 @@
...
+#   Q4_K -> 8 sub-blocks of 32, asymmetric: 6-bit unsigned scales + 6-bit
+#           unsigned mins, fp16 super-block d / dmin.
+#   Q3_K -> 16 sub-blocks of 16, symmetric: 6-bit signed scales, single fp16
+#           super-block d (no min / zero-point).
+#   Q2_K -> 16 sub-blocks of 16, asymmetric: 4-bit unsigned scales + 4-bit
+#           unsigned mins, fp16 super-block d / dmin.
+# ---------------------------------------------------------------------------
+
+
+class _GGUFCachedQuantRestrictValue(QuantRestrictValue):
```
Giuseppe5, line 138: name is a bit misleading, since it is not caching or using anything that is cached

src/brevitas_examples/llm/gguf_export/export.py
```diff
@@ -16,35 +16,52 @@
...
+            yield m
+
+
+def _enable_gguf_export_caching(model):
+    """Enable metadata-only weight caching on each GGUF proxy."""
+    prior_settings = {}
+    for wq in _gguf_proxies(model):
+        prior_settings[wq] = (
+            wq.cache_inference_quant_weight, wq.cache_inference_quant_weight_metadata_only)
+        wq.cache_inference_quant_weight = False
```
Giuseppe5: This should not be necessary, same below

src/brevitas_examples/llm/gguf_export/base_quantizers.py
i-colbert: It has its own cache called `gguf_scaling_cache` and it adds that there.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
i-colbert: It has its own cache called `gguf_zero_point_cache` and populates it there.

src/brevitas_examples/llm/gguf_export/export.py
i-colbert: It may be necessary actually. The cache needs to be cleared to be updated (unless we also make a change to Brevitas core), and `quant_inference_mode` enables metadata-only weight caching by default without clearing the cache on exit. Clearing the weight cache avoid the possibility of having stale metadata.

src/brevitas_examples/llm/main.py
```diff
@@ -492,9 +483,16 @@ def quantize_llm(args, extra_args=None):
             quantizer_name = parse_custom_quantizer(args.custom_quantizer)
             custom_quantizer = QUANTIZERS_REGISTRY.get(quantizer_name)
             quantizers_dict = custom_quantizer.override_quantizers_dict(quantizers_dict)
+        # The embedding (first layer) is quantized only when first/last-layer
+        # quantization is requested. Its weight quantizer is the same one used for
+        # the linear layers, so a custom quantizer that branches on the module
+        # name/type (e.g. embedding -> Q6_K) is honoured.
```
Giuseppe5: This part referring to Q6_K is kind of out of place here, the rest of the comment can be shorten but valid.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,363 @@
...
+from brevitas.quant.shifted_scaled_int import ShiftedUint8WeightPerChannelFloat
+from brevitas.quant.shifted_scaled_int import ShiftedUint8WeightPerTensorFloat
+from brevitas_examples.llm.gguf_export.proxy import GGUFGroupwiseWeightQuantProxyFromInjector
+
+# GGML block geometry: K-quant super-block spans QK_K elements; Q4_0/Q4_1/Q8_0 span QK.
+QK = 32
+QK_K = 256
+
+# K-quant sub-block geometry and nested scale bit widths
+Q6_K_GROUP_SIZE = 16  # 16 sub-blocks of 16 per 256-element super-block
```
pablomlago, line 66: I would try to give some structure to these variables, as it would be desirable to group those associated to the same format.

A nested dictionary could suffice, but something fancier like
```
from dataclasses import dataclass

@dataclass(frozen=True)
class KQuantConfig:
    group_size: int
    sub_scale_bit_width: int
    sub_zp_bit_width: int | None = None


K_QUANT_CONFIG = {
    "Q6_K": KQuantConfig(16, 8),
    "Q5_K": KQuantConfig(32, 6, 6),
    "Q4_K": KQuantConfig(32, 6, 6),
    "Q3_K": KQuantConfig(16, 6),
    "Q2_K": KQuantConfig(16, 4, 4),
}
```
could also be done.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+class GGUFQ4_0WeightQuant(_GGUFBaseQuantMixin, Int8WeightPerChannelFloat):
+    """Signed symmetric 4-bit group quantizer with one fp16 scale per 32-element block."""
+    gguf_qtype = gguf.GGMLQuantizationType.Q4_0
+    group_size = QK
+    bit_width = 4
+
+
+# ---------------------------------------------------------------------------
```
pablomlago, line 123: This format of comment is not used elsewhere in the codebase. I would make this comment a docstring of some class, e.g. `_GGUFShiftedBaseKQuantMixin`.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,363 @@
...
+#   Q3_K -> 16 sub-blocks of 16, symmetric: 6-bit signed scales, single fp16
+#           super-block d (no min / zero-point).
+#   Q2_K -> 16 sub-blocks of 16, asymmetric: 4-bit unsigned scales + 4-bit
+#           unsigned mins, fp16 super-block d / dmin.
+# ---------------------------------------------------------------------------
+
+
+class _GGUFCachedQuantRestrictValue(QuantRestrictValue):
+
+    def __init__(
```
pablomlago, line 140: Please add type hints for clarity.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,363 @@
...
+            return zero_point_shape
+        return None
+
+
+class _GGUFSignedBaseKQuantMixin(__GGUFBaseKQuantMixin, MSESymmetricScale):
+    """Base quantizer for signed symmetric K-quants with nested scales."""
+    signed = True
+    restrict_scaling_type = RestrictValueType.SIGNED_FP
+    scaling_stats_op = StatsOp.SIGNED_MAX
+    restrict_scale_positive = False
```
pablomlago, line 227: Isn't this resolved from `SolveRestrictScaleSign`? I would check whether this is a base class of the quantizer. If so, there is no need to declare this attribute here.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
```diff
@@ -0,0 +1,363 @@
...
+        return this.zp_quant.rescaling_int_quant
+
+    @value
+    def zero_point_dequantized_shape(scaling_per_output_type, zero_point_shape):
+        if scaling_per_output_type == ScalingPerOutputType.GROUP:
+            return zero_point_shape
+        return None
+
+
+class _GGUFSignedBaseKQuantMixin(__GGUFBaseKQuantMixin, MSESymmetricScale):
```
pablomlago, line 222: I would be cautious of the inheritance order when ` MSESymmetricScale` is used. My intuition is that this class you be the first one, i.e.
`class _GGUFSignedBaseKQuantMixin(MSESymmetricScale, __GGUFBaseKQuantMixin)`
See https://github.com/Xilinx/brevitas/pull/1468 for a related issue.

src/brevitas_examples/llm/gguf_export/convert.py
```diff
@@ -421,58 +435,46 @@ def quantize(self, name, data, data_qtype, old_dtype, bid):
...
-        quant_data = quant_weight.int()
-        scale = quant_weight.scale_ if hasattr(quant_weight, 'scale_') else quant_weight.scale
-        zp = quant_weight.zero_point_ if hasattr(
-            quant_weight, 'zero_point_') else quant_weight.zero_point
-        _, quant_data = self.modify_tensors(quant_data, name, bid)[0]
 
-        # TODO: Generalize this to have a map between GGUF quant type
-        # and our preprocessing for quant_modules
-        if data_qtype == gguf.GGMLQuantizationType.Q4_K:
+        def _modify_if_tensor(value):
```
pablomlago, line 439: I would change the name of this function (and that of `modify_tensors`) to something more self-explanatory.

pablomlago (commented):
Minor comments from my first pass:

src/brevitas_examples/llm/gguf_export/base_quantizers.py
pablomlago: There are several enumerations similar to this one, and it might be desirable to group/unify them for easier maintainability.

src/brevitas_examples/llm/gguf_export/base_quantizers.py
i-colbert: Thanks for flagging this. `_GGUFBaseKQuantMixin` does not define MSE attributes so the quantizer resolves to MSE. I haven't changed the inheritance order here as it may have other unintended consequences. I did add tests for the instantiated module though.

src/brevitas_examples/llm/gguf_export/convert.py
i-colbert: Thanks for the suggestion. `modify_tensors` is a general callback inherited from the llama.cpp conversion scripts. Each model can use it to rename, reshape, reorder, split, combine, or skip tensors.

A rename is therefore not a local change. It would also affect the inheritance and tests. I suggest keeping the established name, but welcome alternative function names.

src/brevitas_examples/llm/gguf_export/export.py
```diff
@@ -58,55 +76,88 @@ def _resolve_model_name(name_or_path: str) -> str:
...
-    tokenizer.save_pretrained(tmp_work_dir)
-    config.save_pretrained(tmp_work_dir)
-    if getattr(model, 'generation_config', None) is not None:
-        model.generation_config.save_pretrained(tmp_work_dir)
-
-    with torch.no_grad():
-        hparams = ModelBase.load_hparams(tmp_work_dir)
-        model_architecture = hparams["architectures"][0]
-        try:
+    # TODO: every tensor now carries its own qtype via
```
Giuseppe5: When are you planning to address this? Before/after we merge this?

src/brevitas_examples/llm/gguf_export/convert.py
```diff
@@ -421,58 +435,46 @@ def quantize(self, name, data, data_qtype, old_dtype, bid):
...
-                quant_data,
-                data_qtype,
-                scale,
-                zp,
-                wmin_m=zp,
-                d_scale=scale_scale,
-                d_wmin_m=scale_zp)
+        weight_quant = module.weight_quant
 
+        if isinstance(weight_quant, GGUFGroupwiseWeightQuantProxyFromInjector):
```
Giuseppe5: What happens when data_qtype is not defined? Does that happen for Q4_0?

src/brevitas_examples/llm/gguf_export/export.py
```diff
@@ -58,55 +76,88 @@ def _resolve_model_name(name_or_path: str) -> str:
...
+        fname_out = Path(export_path)
+        fname_out.parent.mkdir(parents=True, exist_ok=True)
+    else:
+        fname_out = Path(export_path)
+        fname_out.mkdir(parents=True, exist_ok=True)
+
+    tmp_work_dir = Path(tempfile.mkdtemp(prefix='brevitas_gguf_export_'))
+    is_training = model.training
+    prior_gguf_cache_settings = dict()
+    try:
```
Giuseppe5: Why is there a try with no except? In general I prefer explicit if conditions if possible

PR #1538, opened 2026-06-30
Fix (fx): push support for value_tracer

Giuseppe5:
## Reason for this PR

Push support for value tracer to include latest 2.12 torch version.


nickfraser (approved):
LGTM!

PR #1553, opened 2026-07-15
Docs: add v0.13.0 to the version selector

Giuseppe5:
When releasing a new version, we need to update the `versions.json`.

The `versions.json` in `docs/dev` is the groud truth from which the version selector is generated, with the idea that other versions might be dropped/added in the future, while `dev` should _always_ be there.

`versions.json` in other subfolders (e.g., 'versions.json' in `/docs/v0.11.0`) are ignored, so we do not have to update them (I hope).


nickfraser (approved):
LGTM!

PR #1555, opened 2026-07-20
Fix (brevitas_examples/custom_trainer): tensorboard compatibility

### LINKED ISSUE #1554: LLM Custom Trainer with Tensorboard
**Describe the bug**

When training rotations with our LLM entrypoint, if there is tensorboard installed, it will cause an error during the serialization of the arguments, because we are passing a callable in one of the arguments of TrainingArguments.

If tensorboard is not installed, this error does not appear, which is why we missed it so far.

This also occurs if writing a custom trainer following the same example of RotationTrainer or the one we have in the tests.

**Expected behavior**

Trainer to work also with tensorboard installed.

Giuseppe5:
## Reason for this PR
Fixes #1554

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/trainer_utils.py
+++ b/src/brevitas_examples/llm/llm_quant/trainer_utils.py
@@ -33,6 +33,33 @@ from brevitas_examples.common.trainer_utils import parse_optimizer_class
 # these under ``get_param_fn``.
 ParamsFn = Callable[[torch.nn.Module, "transformers.TrainingArguments"], List[torch.nn.Parameter]]
 
+
+def _json_safe(value: Any) -> Any:
+    """Return a JSON-serializable deep copy of ``value``.
+
+    ``TrainingArguments`` may carry values that ``json.dumps`` cannot handle --
+    most notably the ``get_param_fn`` callables and the optimizer/scheduler class
+    objects stored inside ``optimizer_scheduler_args``. This helper recursively
+    converts such values into serializable equivalents:
+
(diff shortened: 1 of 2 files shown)


src/brevitas_examples/llm/llm_quant/trainer_utils.py
```diff
@@ -33,6 +33,33 @@
...
+    converts such values into serializable equivalents:
+
+    * ``dict`` / ``list`` / ``tuple`` are traversed recursively.
+    * callables (functions, classes, lambdas, ``functools.partial``) are replaced
+      by their ``__name__`` when available, otherwise ``str(value)``.
+    * JSON-native scalars (``str``, ``int``, ``float``, ``bool``, ``None``) are
+      returned unchanged.
+    * anything else falls back to ``str(value)``.
+    """
+    if value is None or isinstance(value, (str, int, float, bool)):
```
pablomlago: I'm not a big fan of having this enumeration. In the end, the motivation for this PR was to handle `callables` mainly. 

A more generic proposal would be:
```
def _replace_callables(value: Any) -> Any:
    if callable(value):
        return getattr(value, "__name__", str(value))
    if isinstance(value, Mapping):
        return {k: _replace_callables(v) for k, v in value.items()}
    if isinstance(value, Sequence) and not isinstance(value, (str, bytes)):
        return type(value)(_replace_callables(v) for v in value)
    return value

# Inside the dataclass
def to_dict(self):
    return _replace_callables(super().to_dict())
```

pablomlago (approved):
LGTM, assuming tests pass. Are you planning to create a dedicated test or did you verify the fix locally?

PR #1556, opened 2026-07-20
Setup: adjust torch & python versions

Giuseppe5:
## Reason for this PR
- Minimum torch version is now 1.13
    - Remove unused guards for torch < 1.13
- Update periodic tests to support most recent torch versions (until 2.13) and python version (up to 3.14) 

NB: tests/brevitas_examples/test_pretrained_accuracy.py, tests were skipped if torch version was greater than torch 1.6, so we haven't tested those in a while.. Re-enabling them here and see what happens.

### DIFF
--- a/.github/workflows/gen_github_actions.py
+++ b/.github/workflows/gen_github_actions.py
@@ -73,7 +73,6 @@ PYTEST_MATRIX_EXTRA_REDUCED = od([('jit_status', [
 ALL_SUPPORTED_PYTHON_VERSIONS = ('3.10', '3.11', '3.12', '3.13', '3.14')
 
 ALL_SUPPORTED_PYTORCH_VERSIONS = (
-    '1.12.1',
     '1.13.1',
     '2.0.1',
     '2.1.1',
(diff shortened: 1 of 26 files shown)


.github/workflows/gen_github_actions.py
```diff
@@ -66,7 +70,7 @@
     'jit_disabled',])])
 
 # Data shared betwen Nox sessions and Github Actions, formatted as tuples
-ALL_SUPPORTED_PYTHON_VERSIONS = ('3.9', '3.10', '3.11', '3.12')
+ALL_SUPPORTED_PYTHON_VERSIONS = ('3.10', '3.11', '3.12', '3.13', '3.14')
 
 ALL_SUPPORTED_PYTORCH_VERSIONS = (
     '1.12.1',
```
nickfraser: Remove `1.12.1` from `ALL_SUPPORTED_PYTORCH_VERSIONS`

.github/workflows/gen_github_actions.py
```diff
@@ -83,16 +87,33 @@
...
-    '2.12.1')
+    '2.12.1',
+    '2.13.0')
 
 ALL_SUPPORTED_EXCLUSION_LIST = generate_exclusion_list([
-    [['python_version', [
-        '3.9',]], ['pytorch_version', [
-            '2.9.0',]]],
     [['python_version', [
         '3.11',]], ['pytorch_version', ['1.12.1', '1.13.1']]],
```
nickfraser: Also remove `1.12.1` in the exclusion lists.

nickfraser (commented):
Will do more thorough review later, just spotted these.

PR #1557, opened 2026-07-21
Test (brevitas_examples): restore skipped accuracy test

Giuseppe5:
## Reason for this PR
Restore the failing pretrained test doing the following:
- Have a limited set of samples to check during inference
- check against a new, known reference (computed with torch 2.7.1)
- Creates a "fast" version where we check against a limited subset of samples (5 batches, 100 samples per batch), and a "long" test version that checks against the entire dataset


tests/brevitas_examples/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+
+    A file lock plus a sentinel file guarantee that only the first process
+    performs the (serial) download while the others wait and then reuse the
+    cached copy.
+    """
+    try:
+        import torchvision  # noqa: F401
+
+        from brevitas_examples.bnn_pynq.trainer import MirrorMNIST  # noqa: F401
+    except ImportError:
```
nickfraser: Is this actually needed? I.e., does the `bnn_pynq_datasets` fixture _actually_ get loaded during tests when it is not used?

tests/brevitas_examples/test_pretrained_accuracy.py
```diff
@@ -1,17 +1,53 @@
...
-from tests.marker import requires_pt_lt
+
+# Number of samples to evaluate. With the default batch size of 100 this
+# corresponds to 5 batches, i.e. the smallest subset used to generate the
+# reference values.
+NUM_SAMPLES = 500
+
+# Local file storing the reference Prec@1 values, generated with the highest
+# torch version tested in .github/workflows/examples_pytest.yml (2.7.1) and
+# checked against all the other versions.
```
nickfraser, line 19: Can you add a link to the release(es) which contained the earlier references here? Otherwise, we may loses track to what the original, golden reference was.

noxfile.py
```diff
@@ -127,17 +127,21 @@ def tests_brevitas_examples_cpu(session, pytorch, jit_status):
...
-        '--ignore-glob',
-        'tests/brevitas_examples/*llm*',
-        '--ignore-glob',
-        'tests/brevitas_examples/*vision*',
-        '--ignore-glob',
-        'tests/brevitas_examples/*diffusion*',
-        'tests/brevitas_examples')
+    if session.posargs:
+        # Allow targeting specific tests/paths, e.g. to (re)generate references.
+        session.run('pytest', '-n', 'logical', *session.posargs)
```
nickfraser: Can you document the PR with when/how this is used?

nickfraser (approved):
Some minor comments, otherwise, LGTM!

tests/brevitas_examples/conftest.py
Giuseppe5: Everytime we call `pytest -m ...`, yes all the conftest.py files are loaded at collection time.
If we only call specific testing files (e.g., `pytest /tests/path/to/subfolder`), then we would not load this.

I would still keep this check, just to be safe.

noxfile.py
Giuseppe5: Never, removing it

tests/brevitas_examples/conftest.py
Giuseppe5: Also, it gets called because `autouse=True`. We could remove that.

Technically, we also install torchvision always, so this try/except is useless from that point of view.

tests/brevitas_examples/test_pretrained_accuracy.py
Giuseppe5: https://github.com/Xilinx/brevitas/releases/tag/bnn_pynq-r0

PR #1560, opened 2026-07-23
Fix (nn/utils): correct pointers matching during quant weight merging

Giuseppe5:
## Reason for this PR

Root cause: In merge_quant_weights (src/brevitas/nn/utils.py), the hook matched a weight to its quantized output via `id(m.weight.data) == id(input_tensor.data)`. This was unreliable because:
1. `args[0]` is `self.weight[slice(None)]` (a view created in mixin/parameter.py:71-75), not self.weight itself.
2. `tensor.data` returns a fresh Tensor wrapper on every access, so comparing `id()` of two `.data` accesses depends on CPython GC timing 

## Changes Made in this PR

- Added `_same_storage(a, b)` comparing untyped_storage().data_ptr() (with a .storage() fallback for older PyTorch). This is stable and correctly identifies a tensor and its view as sharing storage.
- Replaced the `id(...)` check with _same_storage(m.weight, input_tensor).
- Only append a proxy to proxy_list when at least one weight matched (matched flag).
- Hardened the reset loop with `module_tensor_id_mapping.get(module, 0)` so any genuinely-unmatched proxy raises the intended `RuntimeError` instead of a raw `KeyError`.

### DIFF
--- a/src/brevitas_examples/llm/llm_quant/run_utils.py
+++ b/src/brevitas_examples/llm/llm_quant/run_utils.py
@@ -28,6 +28,7 @@ from torch.utils._pytree import tree_map
 from transformers import AutoConfig
 
 from brevitas.fx.value_tracer import ValueProxy
+from brevitas.utils.torch_utils import same_storage
 
 
 def modify_dataloader(model_name_or_path, data, dtype):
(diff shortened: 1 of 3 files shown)


nickfraser (approved):
LGTM!

src/brevitas_examples/llm/llm_quant/run_utils.py
```diff
@@ -95,11 +96,14 @@ def __torch_dispatch__(self, func, types, args=(), kwargs=None):
 # We rely on the fact that two versions of the same model (eager vs FX) might have different modules id (id(fx_module) != id (eager_module))
 # However, the underlying tensors are still shared, so we can recostruct the mapping between the two
 # modules.
+# NOTE: two versions of the same model might also wrap the shared tensors in distinct Parameter
+# objects (id(fx_param) != id(eager_param)), so we match on shared storage rather than on id.
 def fix_rewriter(rewriters, old_model_ref, tensor_name):
     for r in rewriters:
-        tensor_id = id(r.old_module_instance.weight)
+        old_tensor = getattr(r.old_module_instance, tensor_name)
```
nickfraser, line 103: Worth adding a guard that `tensor_name` is actually a tensor type? Otherwise I assume this fails inside `same_storage` for a missing attribute.

nickfraser (approved):
One small comment, otherwise LGTM!

src/brevitas_examples/llm/llm_quant/run_utils.py
Giuseppe5: Done, thanks for spotting.

PR #1563, opened 2026-07-27
fix (ci/deps): Fix CI for `torch==2.13`, lift dependency to `setuptools<81`

nickfraser:
# Fix CI PyTorch 2.13.0 install and rescope the setuptools pin

## Summary

The periodic CI matrix could not install `torch==2.13.0`: the nox install
command produced an unresolvable pip request, and once that was corrected a
conflict surfaced between torch's `setuptools>=77.0.3` requirement and Brevitas'
`setuptools<70.0` pin. This PR fixes the nox install command to request
unambiguous CPU wheels, drops Brevitas' own reliance on `pkg_resources` and
`distutils`, and rescopes the setuptools pin to `setuptools<81` to satisfy the
full torch range (a shadow dependency of `torch.utils.cpp_extension`).

### nox CPU wheel resolution (`noxfile.py`)

For non-macOS runners on torch >= 2.4.0, the install commands requested a bare
`torch=={ver}` / `torchvision=={ver}` while pointing pip at both the PyTorch CPU
index and PyPI. Both a plain wheel and a `+cpu` local-version wheel then
satisfied the same version specifier, and combined with torchvision's exact
pinned-torch dependency this yielded `ResolutionImpossible` for 2.13.0. The
requests now pin the explicit `+cpu` local version for both packages, matching
the existing pre-2.4.0 branch and leaving a single candidate for the resolver.

### Drop Brevitas' own pkg_resources / distutils usage (`src/brevitas`)

torch 2.13.0 depends on `setuptools>=77.0.3`, which conflicted with the previous
`setuptools<70.0` pin. That pin existed to keep two legacy imports working, so
both are replaced with stdlib equivalents:

- `src/brevitas/config.py`: `distutils.util.strtobool` is replaced with a local
  `strtobool` helper (`distutils` was removed from the stdlib in Python 3.12 and
  from setuptools' shim in later releases). It returns `bool` and still raises
  `ValueError` on invalid input, preserving `env_to_bool` behavior. The now
  redundant `bool(...)` cast in `env_to_bool` is dropped, since the helper
  already returns `bool` (the old `distutils` version returned `int`).
- `src/brevitas/__init__.py`: the deprecated `pkg_resources`
  (`get_distribution` / `DistributionNotFound`) version lookup is replaced with
  `importlib.metadata` (`version` / `PackageNotFoundError`). The import is
  aliased to avoid clashing with the existing `packaging.version`.

### Rescope the setuptools pin (`requirements/requirements.txt`)

The `<70.0` pin is replaced with `setuptools<81` rather than removed, because
`torch.utils.cpp_extension` (imported at `src/brevitas/__init__.py`) has an
unconditional top-level `import setuptools` for every supported torch version,
and torch under-declares it: undeclared for torch <= 2.4.1, declared only for
`python_version >= "3.12"` in 2.5.1-2.10.0, and declared (`<82`, then
`>=77.0.3`) from 2.11.0 onward. Listing setuptools ensures it is installed for
the under-declared versions. The `<81` cap keeps `pkg_resources` (removed in
setuptools 81) available for torch < 2.2 (1.13.1, 2.0.1, 2.1.1), whose
`cpp_extension` also imports it. No lower bound is needed (torch 2.13.0 supplies
its own `>=77.0.3`), and no torch extra provides setuptools.

### DIFF
--- a/src/brevitas/config.py
+++ b/src/brevitas/config.py
@@ -20,7 +20,7 @@ def strtobool(val: str) -> bool:
 
 
 def env_to_bool(name: str, default: bool) -> bool:
-    return bool(strtobool(os.environ.get(name, "{}".format(default))))
+    return strtobool(os.environ.get(name, "{}".format(default)))
 
 
 REINIT_ON_STATE_DICT_LOAD = env_to_bool('BREVITAS_REINIT_ON_STATE_DICT_LOAD', True)
--- a/src/brevitas/config.py
+++ b/src/brevitas/config.py
@@ -2,13 +2,21 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from contextvars import ContextVar
-from distutils.util import strtobool
 import os
 
 try:
     from torch.jit import _enabled
 except ImportError:
     from torch.jit._state import _enabled
+
+
+def strtobool(val: str) -> bool:
+    val = val.lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return True
+    if val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    raise ValueError(f"invalid truth value {val!r}")
 
 
 def env_to_bool(name: str, default: bool) -> bool:
--- a/.github/workflows/base_periodic.yml.template
+++ b/.github/workflows/base_periodic.yml.template
@@ -1,14 +1,8 @@
 name: &{name}
 
-# TEMPORARY (sanity-check): run the full support matrix on PRs like the reduced
-# workflows to validate the torch 2.13.0 fix. Revert to the schedule before merge.
 on:
-  pull_request:
-    types:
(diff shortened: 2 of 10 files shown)


src/brevitas/config.py
```diff
@@ -11,6 +10,15 @@
...
+    val = val.lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return True
+    if val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    raise ValueError(f"invalid truth value {val!r}")
+
+
 def env_to_bool(name: str, default: bool) -> bool:
     return bool(strtobool(os.environ.get(name, "{}".format(default))))
```
nickfraser: The cast doesn't seem to be needed anymore

pablomlago (approved):
LGTM

PR #1565, opened 2026-07-28
Fix (nn/graph/test): Do not rename tensors if it's not supported (e.g., `PyTorch>=2.13`)

nickfraser:
# Support PyTorch >= 2.13 by replacing named tensors

[Example failure](https://github.com/Xilinx/brevitas/actions/runs/30342832276/job/90222122095?pr=1563)

## Summary

PyTorch 2.13 removed the named-tensors API (`Tensor.rename`, `Tensor.rename_`,
`Tensor.names`), which crashed nearly every quant-layer forward with
`AttributeError: 'Tensor' object has no attribute 'rename'` and broke the
batch-dimension signalling that PTQ algorithms relied on for
`QuantMultiheadAttention`. This PR makes the remaining named-tensor calls safe on
all versions, replaces named-tensor batch-dimension detection with a
module-level mechanism plus shared helpers, and fixes a latent
`LayerwiseActivationEqualization` bug in the process.

### Named-tensor compatibility shim
- Added `NAMED_TENSORS_SUPPORTED` and `rename_tensor` / `rename_tensor_`
  helpers in `src/brevitas/utils/torch_utils.py`. They delegate to the native
  methods when named tensors exist and become no-ops on PyTorch >= 2.13.
- Routed the producer sites through the shim: `unpack_input` in
  `src/brevitas/nn/mixin/base.py` (the path hit by most quant layers and the
  original crash site) and all `L`/`N`/`E` mark/clear calls in
  `src/brevitas/nn/quant_mha.py` (consolidated behind a local `_rename_` helper
  that also unwraps `QuantTensor`).

### Module-level batch dimension
Named tensors were previously used to tell PTQ hooks that a
`QuantMultiheadAttention` internal projection receives inputs in `(L, N, E)`
layout (batch at index 1). That signal disappears on PyTorch >= 2.13.
- `QuantMultiheadAttention` now sets a constant `batch_dim = 1` on each internal
  projection (`in_proj`/`q_proj`/`k_proj`/`v_proj`/`out_proj`). The value is
  invariant because the module normalizes inputs to `(L, N, E)` before the
  projections regardless of `batch_first`.
- `_process_input` (`src/brevitas/graph/base.py`) now also strips any residual
  dimension names before returning, so downstream `permute`/`reshape` work. On
  PyTorch < 2.13 the `'N'` name previously leaked into the collected activation
  statistics and made activation equalization on MHA projections fail with
  `permute is not yet supported with named tensors`; this makes that path work
  as intended on all versions. The strip is guarded (`isinstance`/`hasattr`) so
  it never touches a `QuantTensor` or the removed `Tensor.names` attribute.

### Batch-dimension resolution helpers
- Added `get_batch_dim(module, inp, default)` and
  `resolve_region_batch_dim(modules, default)` to `src/brevitas/graph/utils.py`.
  `get_batch_dim` resolves a single module's batch dimension from an explicit
  `batch_dim` attribute, then `batch_first`, then a legacy named-tensor
  fallback. `resolve_region_batch_dim` aggregates across a region and raises on
  genuinely conflicting declared dimensions instead of silently reconciling
  them (it uses a `None` sentinel to distinguish "unspecified" from an explicit
  dim 0).
- Consumers (`graph/base.py`, `graph/gpxq.py`, `graph/permute.py`,
  `graph/equalize.py`) were refactored to use these helpers, removing duplicated
  `hasattr(..., 'batch_first')` logic.
- Fixed a latent bug in `LayerwiseActivationEqualization.setup`: it checked
  `hasattr(region, 'batch_first')` on a frozen `Region` dataclass that never has
  that attribute, so `batch_dim` was always `0`. It now derives the batch
  dimension from the module. This was inert under the default `maxabs` statistic
  but produced wrong scales under `range` for `batch_first=False`.

### Type annotations
- Added signatures to the new helpers (`get_batch_dim`,
  `resolve_region_batch_dim`, `rename_tensor`, `rename_tensor_`, `_rename_`).

### Tests

Most of the diff lives in tests; this section describes what each one checks and,
where relevant, the failure it pins down.

**Fixtures** (`tests/brevitas/graph/equalization_fixtures.py`): `mha_model`
(bare `nn.MultiheadAttention`), `mha_with_source_model` (`Linear -> ReLU -> MHA`,
so graph mode forms a region around the MHA sink), `quant_mha_gpxq_model`
(`QuantMultiheadAttention` with weight quantization), and `mha_input` (lays data
out as `(N, L, E)` / `(L, N, E)` per `batch_first`). Sequence length and batch
size differ so that reducing or splitting over the wrong axis produces a
detectable difference.

**Regression tests (red before the fix, green after)**

- `test_act_equalization_mha_layerwise` — Verifies layerwise activation
  equalization on a plain `nn.MultiheadAttention` (the realistic pre-quantization
  target). It asserts exactly one module is hooked and that it is the MHA itself
  (equalization treats MHA as a leaf and does not hook internal projections such
  as `out_proj`), that the detected `batch_dim` is `0` for `batch_first=True`
  `(N, L, E)` and `1` for `batch_first=False` `(L, N, E)` (since `batch_first`
  only permutes the module's I/O), and that the output is preserved. *Before:*
  `LayerwiseActivationEqualization.setup` probed `hasattr(region, 'batch_first')`
  on a frozen `Region` dataclass that never carries that attribute, so
  `batch_dim` silently defaulted to `0` for every region; the `batch_first=False`
  case expected `1` and got `0` (AssertionError). *After:* `setup` derives
  `batch_dim` from the hooked module, producing the correct value.

- `test_act_equalization_mha_layout_equivalence` (parametrized `layerwise`) — A
  behavioural, black-box check: two models with identical weights, one
  `batch_first=True` and one `batch_first=False`, are fed the same data in their
  respective layouts, and their equalization scaling factors must match. It uses
  `scale_computation_type='range'` deliberately: the default `maxabs` reduces to
  the same per-channel result regardless of axis order (max is order-invariant)
  and so *hides* a wrong `batch_dim`, whereas `range` (`max - min`) does not.
  *Before (`layerwise=True`):* the `(L, N, E)` model reduced statistics over the
  sequence axis instead of the batch axis, so its scales diverged from the
  `(N, L, E)` model (allclose failed). *After:* the corrected `batch_dim` makes
  both layouts yield identical scales. The `layerwise=False` case is a control
  (graph mode was already correct) and passes throughout.

- `test_cache_vision_store_plain_tensors`
  (`tests/brevitas_examples/test_learned_round_utils.py`) — Constructs
  `CacheVision` and calls `store_inputs`/`store_output` with plain (unnamed)
  tensors, asserting no exception and the expected cached shapes. This exercises
  the one named-tensor call site the rest of the PR had missed. *Before
  (PyTorch >= 2.13):* `store_output` ran `output.rename_(None)` unconditionally —
  its `if self.batch_dim is not None` guard is always true because `batch_dim` is
  initialised to `0`, never `None` — raising `AttributeError: 'Tensor' object has
  no attribute 'rename_'` on any learned-round vision run. *After:* the call goes
  through the no-op-safe `rename_tensor_` and the dead guard is removed.
  (`store_inputs` name detection degrades to `batch_dim=0` on >= 2.13 where names
  are gone; this is accepted best-effort behaviour, not a regression versus
  `master`, which crashed outright.)

**Controls and coverage (green by construction; guard against regressions)**

- `test_act_equalization_mha_graph` — Graph-mode counterpart of the layerwise
  test using `mha_with_source_model`. Graph mode already derived `batch_dim` from
  `module.batch_first`, so this confirms the graph path stays correct across
  `batch_first` values.
- `test_gpxq_quant_mha` — Runs GPTQ and GPFQ over a `QuantMultiheadAttention` and
  asserts finite output, exercising the migrated QuantMHA forward and GPxQ on its
  internal projections both with and without named-tensor support. GPxQ
  preprocessing (transpose + reshape to `[tokens, features]`) is
  permutation-invariant w.r.t. the batch dimension, so this is end-to-end smoke
  coverage rather than a `batch_dim` assertion.
- `test_mha_projections_expose_batch_dim` (`tests/brevitas/nn/test_attention.py`)
  — Pins the producer contract: every internal projection exposes
  `batch_dim == 1`, for both packed and unpacked in-projection configurations.
- `test_process_input_reads_module_batch_dim` /
  `test_process_input_defaults_without_batch_dim`
  (`tests/brevitas/graph/test_transforms.py`) — Pin the consumer contract:
  `_process_input` reads `batch_dim` from `module.batch_dim` (parametrized
  `0/1/2`) for an unnamed tensor and returns the same tensor object (verifying the
  name-strip path leaves unnamed tensors untouched), and falls back to the
  provided default when the attribute is absent.

## Verification

- Exercised locally on both `torch==2.12.1+cpu` (named tensors present) and
  `torch==2.13.0+cpu` (named tensors removed).
- The three regression tests above were confirmed red before / green after the
  fix. `test_equalization.py` (excluding the download-heavy torchvision cases,
  left to CI), `test_transforms.py`, `test_permute.py`, `test_attention.py`,
  `test_gpxq.py::test_gpxq_quant_mha`, and `test_learned_round_utils.py` pass on
  both versions.
- `isort` (5.11.5) and `yapf` (0.32.0) report no changes on the modified files.

### DIFF
--- a/src/brevitas/graph/base.py
+++ b/src/brevitas/graph/base.py
@@ -22,6 +22,7 @@ from brevitas.fx import immutable_dict
 from brevitas.fx import Node
 from brevitas.graph.utils import *
 from brevitas.utils.python_utils import islambda
+from brevitas.utils.torch_utils import rename_tensor
 
 INPUT_NAMES = ('input', 'inp', 'query', 'hidden_states', 'x')
 
@@ -95,9 +96,18 @@ class GraphTransform(Transform):
             assert len(inp) == 1, "Expected single element tuple"
             inp = inp[0]
 
-        # Extra check for batch_dim using named tensors
-        if hasattr(inp, 'names') and 'N' in inp.names:
+        # Refine batch_dim: prefer an explicit batch_dim exposed by the module (works on all
+        # PyTorch versions), otherwise fall back to named tensors (PyTorch < 2.13).
+        if hasattr(module, 'batch_dim'):
+            batch_dim = module.batch_dim
+        elif hasattr(inp, 'names') and 'N' in inp.names:
             batch_dim = inp.names.index('N')
+
+        # Drop any dimension names before returning so that downstream ops that do not support
+        # named tensors (e.g. permute/reshape on PyTorch < 2.13) keep working.
+        if isinstance(inp, torch.Tensor) and hasattr(inp, 'names') and any(
+                name is not None for name in inp.names):
+            inp = rename_tensor(inp, None)
 
         return inp, batch_dim
 
--- a/src/brevitas/graph/base.py
+++ b/src/brevitas/graph/base.py
@@ -97,10 +97,7 @@ class GraphTransform(Transform):
             inp = inp[0]
 
         # Prefer batch_dim/batch_first exposed by the module; fall back to named tensors (< 2.13).
-        if hasattr(module, 'batch_dim') or hasattr(module, 'batch_first'):
-            batch_dim = get_batch_dim(module, batch_dim)
-        elif hasattr(inp, 'names') and 'N' in inp.names:
-            batch_dim = inp.names.index('N')
+        batch_dim = get_batch_dim(module, inp, default=batch_dim)
 
         # Drop names so downstream ops unsupported for named tensors (e.g. permute/reshape) work.
         if isinstance(inp, torch.Tensor) and hasattr(inp, 'names') and any(
--- a/src/brevitas/graph/base.py
+++ b/src/brevitas/graph/base.py
@@ -96,15 +96,13 @@ class GraphTransform(Transform):
             assert len(inp) == 1, "Expected single element tuple"
             inp = inp[0]
 
-        # Refine batch_dim: prefer an explicit batch_dim exposed by the module (works on all
-        # PyTorch versions), otherwise fall back to named tensors (PyTorch < 2.13).
-        if hasattr(module, 'batch_dim'):
-            batch_dim = module.batch_dim
+        # Prefer batch_dim/batch_first exposed by the module; fall back to named tensors (< 2.13).
+        if hasattr(module, 'batch_dim') or hasattr(module, 'batch_first'):
+            batch_dim = get_batch_dim(module, batch_dim)
         elif hasattr(inp, 'names') and 'N' in inp.names:
             batch_dim = inp.names.index('N')
 
-        # Drop any dimension names before returning so that downstream ops that do not support
-        # named tensors (e.g. permute/reshape on PyTorch < 2.13) keep working.
+        # Drop names so downstream ops unsupported for named tensors (e.g. permute/reshape) work.
         if isinstance(inp, torch.Tensor) and hasattr(inp, 'names') and any(
                 name is not None for name in inp.names):
             inp = rename_tensor(inp, None)
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -1368,8 +1368,8 @@ class LayerwiseActivationEqualization(ActivationEqualization):
             name = list(region.sinks.keys())[0]
             module = region.get_module_from_name(name)
             batch_dim = 0
-            if hasattr(region, 'batch_first'):
-                batch_dim = 0 if region.batch_first else 1
+            if hasattr(module, 'batch_first') and not module.batch_first:
+                batch_dim = 1
 
             hook_fn = partial(
                 self.forward_stats_hook, name=module, batch_dim=batch_dim, use_inp=True)
--- a/src/brevitas/graph/equalize.py
+++ b/src/brevitas/graph/equalize.py
@@ -42,6 +42,7 @@ from brevitas.graph.hadamard import get_hadK
 from brevitas.graph.hadamard import matmul_hadU
 from brevitas.graph.hadamard import matmul_hadU_cuda
 from brevitas.graph.hadamard import random_hadamard_matrix
+from brevitas.graph.utils import get_batch_dim
 from brevitas.graph.utils import get_module
 from brevitas.graph.utils import get_node
 from brevitas.nn import ScaledDotProductAttention
@@ -1367,9 +1368,7 @@ class LayerwiseActivationEqualization(ActivationEqualization):
         for region in self.regions:
             name = list(region.sinks.keys())[0]
             module = region.get_module_from_name(name)
-            batch_dim = 0
-            if hasattr(module, 'batch_first'):
-                batch_dim = 0 if module.batch_first else 1
+            batch_dim = get_batch_dim(module)
 
             hook_fn = partial(
                 self.forward_stats_hook, name=module, batch_dim=batch_dim, use_inp=True)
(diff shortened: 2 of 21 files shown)


src/brevitas/graph/base.py
```diff
@@ -95,10 +96,19 @@ def _process_input(
             assert len(inp) == 1, "Expected single element tuple"
             inp = inp[0]
 
-        # Extra check for batch_dim using named tensors
-        if hasattr(inp, 'names') and 'N' in inp.names:
+        # Refine batch_dim: prefer an explicit batch_dim exposed by the module (works on all
+        # PyTorch versions), otherwise fall back to named tensors (PyTorch < 2.13).
+        if hasattr(module, 'batch_dim'):
```
Giuseppe5: Might be worth creating a utility function to handle this (including the `elif` part of the branch, although its scope changes slightly from case to case), but I'm sure we can clean-up to separate concernes

src/brevitas/graph/equalize.py
```diff
@@ -1458,14 +1457,8 @@ def setup(self):
...
-            for name in region.srcs:
-                module = region.get_module_from_name(name)
-                if hasattr(module, 'batch_first') and not module.batch_first:
-                    batch_dim = 1
-            for name in region.sinks:
-                module = region.get_module_from_name(name)
-                if hasattr(module, 'batch_first') and not module.batch_first:
-                    batch_dim = 1
+            for name in list(region.srcs) + list(region.sinks):
+                batch_dim = max(batch_dim, get_batch_dim(region.get_module_from_name(name)))
```
nickfraser: Add a comment on why `max` makes sense here.

src/brevitas/graph/base.py
```diff
@@ -95,9 +96,13 @@ def _process_input(
...
             inp = inp[0]
 
-        # Extra check for batch_dim using named tensors
-        if hasattr(inp, 'names') and 'N' in inp.names:
-            batch_dim = inp.names.index('N')
+        # Prefer batch_dim/batch_first exposed by the module; fall back to named tensors (< 2.13).
+        batch_dim = get_batch_dim(module, inp, default=batch_dim)
+
+        # Drop names so downstream ops unsupported for named tensors (e.g. permute/reshape) work.
+        if isinstance(inp, torch.Tensor) and hasattr(inp, 'names') and any(
```
nickfraser, line 103: Does this cause an actual failure today? Or is this to prevent some future failure?

src/brevitas/graph/permute.py
```diff
@@ -207,14 +208,8 @@ def _setup_hooks(self):
...
-            for name in region.srcs:
-                module = region.get_module_from_name(name)
-                if hasattr(module, 'batch_first') and not module.batch_first:
-                    batch_dim = 1
-            for name in region.sinks:
-                module = region.get_module_from_name(name)
-                if hasattr(module, 'batch_first') and not module.batch_first:
-                    batch_dim = 1
+            for name in list(region.srcs) + list(region.sinks):
+                batch_dim = max(batch_dim, get_batch_dim(region.get_module_from_name(name)))
```
nickfraser: Again, a comment here about why `max` is a good choice?

nickfraser (commented):
A few issues to address

src/brevitas/graph/base.py
nickfraser, line 103: Is this guard necessary? Can we just always call `rename_tensor`?

nickfraser (commented):
One more question

src/brevitas/graph/base.py
nickfraser: The guard is necessary, because this might be a `QuantTensor`.

PR #1568, opened 2026-07-29
feat (test/ort): reduce the time for ONNXruntime tests with random sampling

nickfraser:
# Sample WBIOL ORT integration tests with Hypothesis

Example runs:
 - before this [PR](https://github.com/Xilinx/brevitas/actions/runs/31189959886/), avg=2154s / job
 - with this PR (1000 samples) [PR](https://github.com/Xilinx/brevitas/actions/runs/33525055749/), avg=289s / job
 - with this PR (10k samples) [PR](https://github.com/Xilinx/brevitas/actions/runs/33533165228/), avg=1376s / job

## Summary

The ORT integration suite enumerated the full cross-product of every WBIOL axis via
`pytest_cases` (~170k collected nodes, the vast majority `pytest.skip`ped), which made a run
take ~36 minutes per job (2154 s; see above). This PR replaces the WBIOL enumeration with a
Hypothesis strategy that draws a random, reproducible subset of valid configurations, while
leaving the much smaller AvgPool
and LSTM suites on their existing full enumeration. It also adds a repo-wide Hypothesis seed,
hardens ONNX artifact cleanup, and — using CI as the oracle — removes two configuration
constraints that turned out to be obsolete while keeping the two that are genuinely required.

### WBIOL: Hypothesis sampling instead of exhaustive enumeration
`QuantWBIOLCases` (a stacked `pytest_cases` parametrization) is replaced by
`wbiol_config_st`, an `@st.composite` strategy that draws each axis (quantizer, impl, rounding,
bit-widths, export type) so that every draw is valid by construction, plus a `build_wbiol_model`
helper and a `WBIOLConfig` dataclass. `test_ort_wbiol` becomes a single `@given` test running
`WBIOL_MAX_EXAMPLES` (10000) examples. Drawing per-axis keeps rare quantizers/export types
well represented without special handling. AvgPool and LSTM tests/cases are unchanged.

Hypothesis' shrinking phase is disabled (`phases=` without `Phase.shrink`) because each example
is a full ONNX export plus ORT inference, making shrink-on-failure prohibitively slow; the
failing example is still reported and is reproducible from the global seed.

### Repo-wide Hypothesis seed (`tests/conftest.py`)
`get_hypothesis_seed()` derives a seed from the environment (Python/torch/platform) via
`zlib.crc32`, and `pytest_configure` applies it through `core.global_force_seed` unless the user
passes `--hypothesis-seed`. Using `crc32` rather than the builtin `hash()` (which is
per-process randomised) keeps xdist workers in agreement on generated examples, while different
matrix jobs still explore different subsets. This applies to all Hypothesis tests in the repo,
not just the ORT suite.

### Configuration-constraint audit
The strategy encodes only the constraints shown to be genuinely required:
- **Kept:** FP8 forces all-8 bit-widths (OCP e4m3 is a fixed 1+4+3 layout that the QCDQ exporter
  rejects otherwise) and dynamic-activation quant forces 8-bit input/output (it maps to ONNX
  `DynamicQuantizeLinear`); weight bit-width stays free for the dynamic case.
- **Removed:** the previous `floor`-rounding all-8 restriction and the `QuantLinear` +
  asymmetric exclusion, both of which now pass across the CI matrix at all bit-widths. This
  widens coverage relative to the previous enumeration.

### ONNX cleanup hardening (`tests/export_fixture.py`)
`rm_onnx` now removes the primary `.onnx` file best-effort (guarding `os.remove` against
`OSError`). Previously, an export that raised before writing the file caused cleanup to throw
`FileNotFoundError`, masking the real error. Relatedly, the all-zero-output case raised inside
`is_brevitas_ort_close` is translated to `assume(False)` in `test_ort_wbiol` (a bare skip would
abort the entire `@given` node), and cleanup runs in a `finally` so artifacts are not leaked
across examples.

## Coverage

Figures are for the CI ORT matrix (36 jobs, all PyTorch <2.8). The strategy's widened valid
WBIOL space is ~67,400 configs; because each matrix job derives a different seed
(`get_hypothesis_seed`), seeded runs accumulate coverage across the matrix. Across-run figures
are approximate (independent seeds; per-axis sampling saturates the small fp8/dynamic branches
and samples the large branch uniformly).

| Option | Runtime/job | Configs per run | Distinct across 36 seeded runs |
|---|---|---|---|
| Before PR (exhaustive) | ~36 min (2154 s) | 32,412 (100% of the old, narrower space) | 32,412 — identical every job (deterministic) |
| This PR, 1k samples | ~5 min (289 s) | 1,000 (~1.5%) | ~25,000 (~37%) |
| This PR, 10k samples | ~23 min (1376 s) | 10,000 (~15%) | ~67,000 (~99%) |

Percentages are relative to the widened ~67,400-config space. The pre-PR suite was exhaustive
but over a smaller, more-constrained space, and repeated identically each job; the sampled suite
trades per-run completeness for broad, accumulating coverage of the wider space across the matrix.

## Verification

Not recorded as a local full-suite run (the 10000-example WBIOL test is intended to run under
CI). Per conversation context, the constraint decisions were validated on the CI matrix: the
retained FP8 and dynamic-activation constraints were confirmed by deterministic, matrix-wide
export failures when relaxed, and the removed `floor` and `QuantLinear`+asymmetric constraints
were confirmed obsolete by clean matrix-wide runs. Strategy behavior (valid-by-construction
draws, seed determinism) was checked locally by sampling configurations without executing
exports.


nickfraser:
Note: for the periodic tests, we should have much larger coverage than the randomly-sampled tests.

tests/brevitas_ort/quant_module_cases.py
```diff
@@ -12,73 +15,121 @@
...
+    if 'asymmetric' in quantizer_name:
+        # QuantLinear + asymmetric fails unreliably in ORT execution.
+        impls = [i for i in impls if i is not QuantLinear]
+    impl = draw(st.sampled_from(impls))
+
+    rounding_type = draw(st.sampled_from(['round', 'floor']))
+
+    # fp8 export and floor rounding require all-8 bit-widths; dynamic act quant requires 8-bit i/o.
+    if is_fp8 or rounding_type == 'floor':
+        o = w = i = 8
```
nickfraser, line 63: Check that these are real constraints

nickfraser: Resolved - removed unnecessary constraints through testing.

Issue #1573, opened 2026-07-31
ci: Illegal instruction errors on windows


nickfraser:
# Windows CPU low-precision MHA crashes

Could be related to forcing the [`+cpu` suffix to specify the PyTorch version](#1563 ), causing different builds to be used in the CI on some platforms. In any case, the tests cases are well handled by the other OSes, so if these continue to be an issue, then we can skip these tests on windows.

## Details

The native failures form a systematic platform and operation-family pattern rather than a single Python or PyTorch version regression:

- They occur only in Windows logs and span Python 3.10, 3.11, and 3.13 with PyTorch 2.5.1, 2.6.0, 2.7.1, 2.8.0, and 2.9.1.
- Every affected parametrization uses `torch.bfloat16` in an MHA-based toy model (`linearmha_model`, `mhalinear_model`, or `layernormmha_model`). No downloaded Linux or macOS log contains a bfloat16 failure, and the corresponding float32 parametrizations do not fail.
- The usual signature is Windows status `0xc000001d` (illegal instruction) followed by an abruptly terminated xdist worker. One trace in the [3.11 / 2.8.0 / Windows](https://github.com/Xilinx/brevitas/actions/runs/30551228857/job/90900122207) job reaches `torch.nn.functional._in_projection_packed` before termination.
- The [3.12 / 2.13.0 / Windows](https://github.com/Xilinx/brevitas/actions/runs/30551228857/job/90900122514) job reports a related but non-fatal backend error, `RuntimeError: could not create a primitive`, for four bfloat16 and two float16 `linearmha_model` parametrizations.
- Some bfloat16 MHA parametrizations pass in the same jobs. This does not make the failures non-systematic: native worker termination marks whichever parametrization is active, so the exact failed cases can vary with xdist scheduling.
- The [3.11 / 2.10.0 / Windows](https://github.com/Xilinx/brevitas/actions/runs/30551228857/job/90900122281) job did not show this signature, but it failed earlier on the unrelated `merge_quant_weights` `KeyError`; it is not sufficient evidence that the combination is unaffected.

[...]

PR #1575, opened 2026-08-07
Fix (ci): skip fp16/bf16 cpu tests on windows+cpu

nickfraser:
# Skip fp16/bf16 activation-equalization tests on Windows CPU

[Example failing run](https://github.com/Xilinx/brevitas/actions/runs/30661657513/job/91259069447), addresses #1573.

## Summary

CI's periodic Pytest runs were failing on Windows CPU jobs (e.g. torch 2.12.1, Python 3.10) with `RuntimeError: could not create a primitive` inside `test_act_equalization_models[linearmha_model-...]`. The failure is an upstream PyTorch/oneDNN limitation: the Windows CPU backend cannot build an mkldnn matmul primitive for `float16`/`bfloat16` inputs to `torch.nn.Linear`, which the `linearmha_model` toy model exercises. This PR skips the affected test combinations instead of letting them fail.

## Changes

- `tests/brevitas/graph/test_equalization.py`: in `test_act_equalization_models`, added a skip guard for `dtype in [torch.float16, torch.bfloat16]` when `device == 'cpu'` and `platform.system() == 'Windows'`, alongside the existing `torch < 2.3.0` skip for the same dtypes. The condition is platform/dtype-based rather than tied to a specific torch version, so it also covers other Windows CPU bf16/fp16 failures independent of the exact PyTorch release.

## Verification

Not run against the CI's Windows matrix (not reproducible locally). Locally verified: `ast.parse` syntax check, `yapf`/`isort` formatting checks against the repo's configured style produced no diff, and `pytest --collect-only -k test_act_equalization_models` collected all 186 parametrized cases without error on Linux (where the new skip does not trigger).

### DIFF
--- a/tests/brevitas/graph/test_equalization.py
+++ b/tests/brevitas/graph/test_equalization.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import copy
+import platform
 
 from packaging.version import parse
 import pytest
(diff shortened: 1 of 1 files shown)


nickfraser:
Failing tests are either flaky, or are addressed in #1560.

PR #1579, opened 2026-08-14
Giuseppe/quant tensor subclass

Giuseppe5:
### Summary

This PR converts QuantTensor from a NamedTuple into a torch.Tensor subclass. 

- Reimplement integer, floating-point, and groupwise quant tensors as tensor subclasses. The subclass tensor represents the dequantized value; quantization metadata is retained as attributes.
- Preserve quantization-aware behavior across set(), detach, contiguity, device/dtype transfers, arithmetic, torch-function handler dispatch, and fallback unpacking to plain tensors.
- Add indexing support that slices leading-dimension metadata together with the value. For groupwise tensors, indexing also maintains the dequantization shape and adjusts the group dimension, including correct handling of negative dimensions. **This is relevant for subsequent PRs and MoE quantization support.**
- Remove the legacy .tensor alias in favor of .value, and update affected tests, export code, and the minifloat MX tutorial.
- Replace None payloads in metadata-only cached quant tensors with empty tensors, which is valid for the tensor-subclass representation and remains compatible with cached-IO flows.
- Tighten decoupled weight-quantization input handling so only IntQuantTensor is treated as a quantized activation input; other tensor subclasses no longer bypass cached-input handling.
- Remove the old tracing workaround that reconstructed a QuantTensor after its NamedTuple representation decayed into a plain tuple. As a tensor subclass, QuantTensor retains its tensor identity during tracing, so type inference from tuple length and field count is no longer needed. ONNX export now also treats genuine tuples consistently as multi-argument inputs, without a special QuantTensor tuple exception.
- Harden ONNX Dynamo export for tensor-subclass QuantTensors by temporarily clearing cache_class on modules that reconstruct cached quantized inputs or outputs. During torch.export, cached values may be represented as Dynamo FakeTensors, from which constructing a QuantTensor subclass is unsupported and can break tracing. The export manager saves the original cache class, clears it before export, restores it after export mode is disabled, and restores it if export setup fails. Normal cached-IO behavior is therefore preserved outside the export context.

### Rationale

Using a real PyTorch tensor subclass makes quantized values more interoperable with standard PyTorch APIs and removes the fragile tuple-decay/reconstruction path used by tracing. The associated export safeguards ensure the new representation remains compatible with Dynamo and ONNX export, while preserving existing quantization behavior and metadata semantics.

### Validation

- Added comprehensive coverage for integer, floating-point, and groupwise tensor-subclass behavior.
- Covers type identity, plain-tensor .value access, gradient preservation, metadata replacement, arithmetic, handler dispatch and fallback behavior, indexing, construction from non-tensor values, unpacking, and device/dtype operations.
- Adds regression coverage verifying that Dynamo export clears and restores cached quant-tensor reconstruction classes.


notebooks/minifloat_mx_tutorial.ipynb
```diff
@@ -137,7 +137,7 @@
    "source": [
     "Groupwise quantization is built on top of integer/minifloat quantization, with special considerations to accomodate for the groupwise scaling.\n",
     "\n",
-    "Compared to Int/Float QuantTensor, the main difference of their groupwise equivalent is that value, scale, and zero_point are not direct attributes anymore but properties. The new attributes are value_, scale_, and zero_point_.\n",
+     "Compared to Int/Float QuantTensor, the main difference of their groupwise equivalent is that value, scale, and zero_point are not direct attributes anymore but properties. The raw (grouped) attributes are `_value_`, `scale_`, and `zero_point_`.\n",
```
nickfraser: Sorry, why the change to `_value_`?

src/brevitas/quant_tensor/base_quant_tensor.py
```diff
@@ -15,38 +14,130 @@
...
 
-# Base class for all QuantTensor.
-# Only assumptions made by these methods are:
-# - `self` is a NamedTuple with a `_fields` attribute
-# - `self` has a `value` attribute
-class QuantTensor:
+# Base class for all QuantTensor types.
+# Subclasses torch.Tensor, where the underlying tensor data represents the dequantized `value`.
+# Quantization metadata (scale, zero_point, bit_width, etc.) are stored as regular attributes.
+class QuantTensor(Tensor):
```
nickfraser, line 22: A few top-level comments:
 - Should this be an ABC? I.e., do we ever want someone to instantiate a `QuantTensor`?
 - Some methods seem to be unnecessarily complicated so they can handle child classes (e.g., lots of `if hasattr(self, '_is_groupwise')`) - shouldn't we delegate this logic to a `GroupwiseMixin` or similar?

src/brevitas/quant_tensor/base_quant_tensor.py
```diff
@@ -15,38 +14,130 @@
...
 
     def detach_(self):
+        super().detach_()
         for field in self._fields:
-            getattr(self, field).detach_()
+            val = getattr(self, field)
+            if isinstance(val, Tensor):
+                val.detach_()
 
     def detach(self):
```
nickfraser, line 118: `def detach` and `def contiguous` (and maybe others? `to`, `cpu`, `cuda`?) seem identical except `detach` -> `contiguous`. If these blocks are to be repeated several times in the code (in the spirit of don't-repeat-yourself), is it worth having a generic private helper to do this?

src/brevitas/utils/quant_utils.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, quant_tensor: IntQuantTensor, metadata_only: bool):
         self.shape = quant_tensor.value.shape
         if metadata_only:
             self.value = None
-            self.quant_tensor = quant_tensor.set(value=None)
+            self.quant_tensor = quant_tensor.set(value=torch.empty(0))
```
nickfraser: Any side-effects of this? Worth auditing anywhere we check for `qt.value is None` in the code or similar?

nickfraser (changes requested):
My feeling is that we want to change the design pattern here. Current `QuantTensor`'s methods are relatively complicated so that they may be used unmodified in the all the derived classes. My personal preference is to keep `QuantTensor` simple / readable and to override these methods in the derived classes through a Mixin or similar - unless there is a good reason to avoid this that I'm not seeing.

src/brevitas/quant_tensor/base_quant_tensor.py
Giuseppe5: Agreed

notebooks/minifloat_mx_tutorial.ipynb
Giuseppe5: There was a bit of overloading between `value`, `value_` and `_value_`. 

Now we only have private methods with leading underscore, indicate the raw storage formats (e.g., for groupwise tensors it's their groupwise representation), while the non-underscore versions are the public-facing values.

src/brevitas/quant_tensor/base_quant_tensor.py
```diff
@@ -13,40 +12,136 @@
...
+            ctor_kwargs = self._get_constructor_kwargs()
+        else:
+            ctor_kwargs = dict(ctor_kwargs)
+        if metadata_transform is not None:
+            for parameter, metadata in ctor_kwargs.items():
+                if isinstance(metadata, Tensor):
+                    ctor_kwargs[parameter] = metadata_transform(parameter, metadata)
+        return type(self)(value, **ctor_kwargs)
+
+    def _apply_and_reconstruct(self, tensor_op, *args, **kwargs):
```
Giuseppe5: If this only calls _reconstruct, maybe we don't need it

src/brevitas/export/onnx/manager.py
```diff
@@ -229,10 +230,20 @@ class ONNXDynamoExportMixin:
...
+            model.apply(lambda module: _override_cache_class(module, enabled=True))
+        try:
+            # NOTE: this QuantTensor-disabling logic is not ONNX-specific and could move to a generic
+            # (backend-agnostic) DynamoExportMixin; deferred to avoid an import cycle (it needs
+            # QuantizationStatusManager and _override_create_quant_tensor).
+            super().set_export_mode(model=model, enabled=enabled)
+        except Exception:
+            if enabled:
+                model.apply(lambda module: _override_cache_class(module, enabled=False))
+            raise
```
nickfraser: Does a bare `raise` like this still produce a human-readable error?

src/brevitas/export/onnx/manager.py
```diff
@@ -229,10 +230,20 @@ class ONNXDynamoExportMixin:
...
+        finally:
+            if not enabled:
+                model.apply(lambda module: _override_cache_class(module, enabled=False))
```
nickfraser: It's not clear to be if there is a change in behaviour here or if we're just guarding against future failures. Can you elaborate?

src/brevitas/proxy/float_runtime_quant.py
```diff
@@ -103,8 +103,8 @@ def create_quant_tensor(
                 qt_args,
                 x.scale,
                 x.zero_point,
-                x.mantissa_bit_width,
                 x.exponent_bit_width,
+                x.mantissa_bit_width,
```
nickfraser: Was this a bug before?

src/brevitas/quant_tensor/base_quant_tensor.py
```diff
@@ -108,56 +178,134 @@ def is_zero_zero_point(tensor):
...
-    value_: Tensor
-    scale_: Tensor
-    zero_point_: Tensor
-    group_size: Tensor
-    group_dim: Tensor
-    bit_width: Tensor
-    signed_t: Tensor
-    training_t: Tensor
-    dequant_shape: Optional[Tuple] = None
+class GroupwiseQuantTensorMixin:
```
nickfraser: Why are there so many method defined here that are unrelated to Groupwise-specific logic. My hope was to avoid lots of duplicated logic my introducing the Mixins. Is there a reason why these should live in the Groupwise Mixin?

Methods that don't seem to be related to groupsize-specific logic:
 - `signed`
 - `training`
 - `__torch_function__`
 - `device`
 - `view`
 - `reshape`
 - `flatten`
 - `transpose`
 - `permute`
 - `__add__`
 - `__mul__`
 - `__truediv__`

A follow-up question is: _should_ `view`, `reshape`, `flatten`, `transpose`, `permute` have groupwise-specific implementations, but we can leave that for a follow-up PR.

src/brevitas/quant_tensor/float_quant_tensor.py
```diff
@@ -56,20 +56,94 @@ def __new__(
...
+        return self._nan_values
+
+    @nan_values.setter
+    def nan_values(self, value):
+        self._nan_values = value
 
     @property
     def signed(self):
-        return self.signed_t.item()
+        return self._signed.item()
```
nickfraser: Other renaming of attributes have just been related to _private_ values. Does the old version no longer exist? Do we want a breaking change like this with this PR?

nickfraser (changes requested):
Quite a few comments - we need to be a bit cautious here. This PR is growing legs from what originally was a sideways movement from `NamedTuple` -> `QuantTensor`, with some cleanup included. We seem to be also changing:
 - some behaviour
 - some public APIs

The more we change, the more it feels like we should "do everything" in this PR, which would make it unwieldy and difficult to review. We might want to talk about scope here.

Essentially, we've deduped some code and left others duped, which leaves this PR is some "halfway" point between a proper refactor or not.

src/brevitas/export/onnx/manager.py
Giuseppe5: No, this must be changed.

src/brevitas/quant_tensor/float_quant_tensor.py
Giuseppe5: I think so; The point is that we have two distinct ways to indicate a "private" attribute currently in master: value_ (NamedTuple doesn't like leading underscore), and training_t (since we store a tensor but we always retrieve a bool).

The idea is to have leading underscore variables as storage variables in all cases; I can also revert training_t/signed_t and leave them as an exception, but since we're changing stuff here (and the fact that this will be a breaking change is almost unavoidable anyway), I'd rather do it here.

src/brevitas/quant_tensor/base_quant_tensor.py
Giuseppe5: I've cleaned up a lot of them, but all the views have to be Groupwise specific.

For non-groupwise quanttensor we made the decision of allowing unsafe operations, i.e. you can flatten a tensor but the quant metadata will not receive the same view, even though a new QuantTensor is generated.

```python
# Regular IntQuantTensor: unsafe reshape is allowed.
qt = IntQuantTensor(
    value=torch.ones(2, 4),
    scale=torch.ones(1, 4),  # Per-channel metadata.
    zero_point=torch.zeros(1, 4),
    bit_width=torch.tensor(8.),
    signed=True,
    training=False,
)

flat = qt.view(-1)

type(flat)       # IntQuantTensor
flat.value.shape # torch.Size([8])
flat.scale.shape # torch.Size([1, 4])  <- unchanged, no longer aligned with value

```


For groupwise quant tensor, generating a new QT with an unsafe view will cause issues everytime .value is called (or `.scale`, `.zero_point`), which is why we override and explicitly demote to normal Tensor.
If we preserved QT structure:
```python

# GroupwiseIntQuantTensor: raw storage is compressed.
qt = GroupwiseIntQuantTensor(
    value=compressed_value,     # e.g. shape [2, 4, 8]
    scale=compressed_scale,
    zero_point=compressed_zp,
    group_size=8,
    group_dim=1,
    bit_width=torch.tensor(8.),
    signed=True,
    training=False,
    dequant_shape=(2, 32),      # Logical expanded shape.
)

flat = qt.view(-1)
bad_qt.value       # Expansion uses invalid storage/group geometry.
bad_qt.scale       # Same issue.
bad_qt.zero_point  # Same issue.
```

src/brevitas/proxy/parameter_quant.py
```diff
@@ -279,7 +279,7 @@ def forward(
             self._cached_act = cached_inp
 
         if self.is_quant_enabled:
-            if quant_input is None or isinstance(quant_input, Tensor):
+            if quant_input is None or not isinstance(quant_input, IntQuantTensor):
```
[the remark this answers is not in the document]
Giuseppe5: This is only compatible with IntQuantTensor inputs (and it only produces IntQuantTensor weights, since this inherits from `DecoupledWeightQuantWithInputProxyFromInjector` which inherits from `WeightQuantProxyFromInjector`, which is the Int-specific implementation of weight proxy).


I'll add a specific assert to reject other QT

src/brevitas/proxy/float_runtime_quant.py
Giuseppe5: Yes

src/brevitas/export/onnx/manager.py
Giuseppe5: This is a new issues introduced with Tensor subclass instead of NamedTuple; it is a bit lengthy to explain here but the main issue is what happens when:
1. caching is enabled
2. and we try to perform operations like quant_tensor.set(value=new_value)

With NamedTuple this was not an issue.

PR #1592, opened 2026-08-25
Feat (utils): compile-friendly utils

Giuseppe5:
## Reason for this PR
Currently we can't compile the groupwise quantizers with `fullgraph=True` due to some utils being not compile-friendly.

This PR addresses this issue.

During quant_inference_mode, we need `group_size` and `group_dim` to be integer to avoid data-conditional failures during compile (they are registered as constant). However, for some export paths (e.g., vLLM), we need them as torch buffers so that they are part of the state dict.
For this reason, we add a thin mixin just for the vLLM export flow so that both behaviours co-exist.

### DIFF
--- a/src/brevitas/export/inference/handler.py
+++ b/src/brevitas/export/inference/handler.py
@@ -184,8 +184,8 @@ class GroupwiseMixin(torch.nn.Module):
             **kwargs) -> None:
         super().__init__(**kwargs)
         self.skip_create_quant_tensor = True
-        self._group_dim = 0
-        self._group_size = 1
+        self._group_dim = None
+        self._group_size = None
 
     @property
     def group_dim(self) -> int:
--- a/src/brevitas/export/inference/handler.py
+++ b/src/brevitas/export/inference/handler.py
@@ -184,23 +184,24 @@ class GroupwiseMixin(torch.nn.Module):
             **kwargs) -> None:
         super().__init__(**kwargs)
         self.skip_create_quant_tensor = True
-        self.register_buffer('group_dim_t', torch.ones((), dtype=torch.int, device=device))
-        self.register_buffer('group_size_t', torch.ones((), dtype=torch.int, device=device))
+        self._group_dim = 0
+        self._group_size = 1
 
     @property
-    def group_dim(self) -> torch.Tensor:
-        return self.group_dim_t.int()
+    def group_dim(self) -> int:
+        return self._group_dim
 
     @property
-    def group_size(self) -> torch.Tensor:
-        return self.group_size_t.int()
+    def group_size(self) -> int:
+        return self._group_size
 
     def prepare_for_export(self, module: nn.Module) -> None:
         if hasattr(super(), 'prepare_for_export'):
             super().prepare_for_export(module)
         if module.is_quant_enabled:
-            self.group_dim_t = torch.tensor(module.group_dim)
-            self.group_size_t = torch.tensor(module.group_size)
+            # Keep grouping metadata in Python so Dynamo treats it as static control flow.
+            self._group_dim = int(module.group_dim)
+            self._group_size = int(module.group_size)
 
     def reshape(self, x: Tensor, group_dim: int, group_size: int) -> Tensor:
         init_shape = list(x.shape)
--- a/src/brevitas/export/inference/handler.py
+++ b/src/brevitas/export/inference/handler.py
@@ -184,24 +184,16 @@ class GroupwiseMixin(torch.nn.Module):
             **kwargs) -> None:
         super().__init__(**kwargs)
         self.skip_create_quant_tensor = True
-        self._group_dim = None
-        self._group_size = None
-
-    @property
-    def group_dim(self) -> int:
-        return self._group_dim
-
-    @property
-    def group_size(self) -> int:
-        return self._group_size
+        self.group_dim = None
+        self.group_size = None
 
     def prepare_for_export(self, module: nn.Module) -> None:
         if hasattr(super(), 'prepare_for_export'):
             super().prepare_for_export(module)
         if module.is_quant_enabled:
             # Keep grouping metadata in Python so Dynamo treats it as static control flow.
-            self._group_dim = int(module.group_dim)
-            self._group_size = int(module.group_size)
+            self.group_dim = int(module.group_dim)
+            self.group_size = int(module.group_size)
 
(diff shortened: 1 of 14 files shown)


src/brevitas/export/inference/handler.py
```diff
@@ -184,23 +184,24 @@ def __init__(
             **kwargs) -> None:
         super().__init__(**kwargs)
         self.skip_create_quant_tensor = True
-        self.register_buffer('group_dim_t', torch.ones((), dtype=torch.int, device=device))
-        self.register_buffer('group_size_t', torch.ones((), dtype=torch.int, device=device))
+        self._group_dim = 0
```
pablomlago: In the original code, this value was initialized to 1, is the change of the initialization value to 0 intended? (`self.register_buffer('group_dim_t', torch.ones((), dtype=torch.int, device=device))`). Should this value be initialized to `None`? I imagine they get their relevant values in `prepare_for_export`.

src/brevitas/export/inference/vLLM/handler.py
```diff
@@ -49,7 +51,25 @@ def compute_scale(self, x: Tensor, group_dim: Optional[int] = None) -> Tensor:
         return scale
 
 
-class vLLMGroupwiseIntInferenceHandler(GroupwiseIntInferenceHandler, StandaloneGroupwiseQuantMixin):
+class vLLMGroupwiseMetadataMixin:
+
+    def __init__(self, **kwargs):
+        device = kwargs.get('device')
```
pablomlago: I'd rather have `super().__init__` as the first line of the function.

src/brevitas/utils/quant_utils.py
```diff
@@ -219,23 +218,14 @@ def float_to_int_impl_to_enum(module):
         return None
 
 
-@brevitas_compiler.disable
 def groupwise_dequant_expand(
```
pablomlago, line 221: Should there be a unit test for this function?

src/brevitas/utils/torch_utils.py
```diff
@@ -148,13 +147,11 @@ def float_internal_scale(
 
 
 @brevitas.jit.ignore
-@brevitas_compiler.disable
 def padding_to_multiple(x: torch.Tensor, dim_to_expand: int, dim_multiple: int) -> torch.Tensor:
```
pablomlago, line 150: Should there be a unit test for this function?

pablomlago (changes requested):
Some minor comments.

src/brevitas/export/inference/handler.py
Giuseppe5: Init value is irrelevant, I'll change this to None, but I believe a buffer can't be None (maybe it can be an empty tensor though)

src/brevitas/export/inference/handler.py
```diff
@@ -184,23 +184,24 @@ def __init__(
...
         self.skip_create_quant_tensor = True
-        self.register_buffer('group_dim_t', torch.ones((), dtype=torch.int, device=device))
-        self.register_buffer('group_size_t', torch.ones((), dtype=torch.int, device=device))
+        self._group_dim = None
+        self._group_size = None
 
     @property
-    def group_dim(self) -> torch.Tensor:
-        return self.group_dim_t.int()
+    def group_dim(self) -> int:
```
pablomlago, line 191: Are the properties `group_dim` and `group_size` needed now? The relevant attributes could just be made public,

tests/brevitas/proxy/test_compile.py
```diff
@@ -113,3 +113,41 @@ def test_compile_act(inp, act_quantizer):
...
+
+    expected = linear(inp)
+    with quant_inference_mode(linear, compile=True):
+        linear(inp)
+        actual = linear(inp)
+
+    assert torch.allclose(expected, actual)
+
+
+@requires_torch_compile()
```
pablomlago, line 137: I would try to unify the markers related to compile into a single one, as it seems that now for every `torch.compile` tests, it is required to add both `@requires_torch_compile` and `@jit_disabled_for_compile`, as well as the skip
```
if platform.system() == "Windows":
        pytest.skip("Skip compile + Windows because the CI runner has no C++ compiler")
```
which could be included into the new marker.

tests/brevitas/utils/test_compile.py
```diff
@@ -0,0 +1,73 @@
...
+@pytest.mark.parametrize(
+    'shape, dim, multiple, expected_shape',
+    [((2, 5), 1, 4, (2, 8)), ((2, 8), 1, 4, (2, 8)), ((2, 3, 5), -1, 4, (2, 3, 8)),
+     ((5, 3), 0, 4, (8, 3))])
+@requires_pt_ge('2.2')
+@requires_torch_compile()
+@jit_disabled_for_compile()
+def test_compile_padding_to_multiple(shape, dim, multiple, expected_shape):
+    x = torch.arange(math.prod(shape), dtype=torch.float32).reshape(shape)
+    compiled_fn = torch.compile(padding_to_multiple, backend='eager', fullgraph=True)
```
pablomlago, line 25: Why is backend set to `eager`? Wouldn't other backends also be possible, e.g. `'inductor'`?

pablomlago (changes requested):
Minor comments:

tests/brevitas/utils/test_compile.py
Giuseppe5: inductor tests fail easily on cpu workers. This allows us to test that dynamo capture works and that fullgraph=True works.

tests/brevitas/proxy/test_compile.py
Giuseppe5: Yes I was thinking about that. I will make this change

Giuseppe5: `pytest.skip("Skip compile + Windows because the CI runner has no C++ compiler")`
I will leave this out, since torch.compile with eager backend still works on windows.

Out of scope for this PR, but we need to make this more robust.
In the PR that will enable GPU testing, this will be one of the main focus.

tests/brevitas/utils/test_compile.py
pablomlago: What about having a `parametrize` over backends, and `xfail` on certain combinations, e.g. `cpu` + `inductor`?

Giuseppe5: I am planning to address that in the GPU tests PR (coming up soon)

pablomlago (approved):

PR #1595, opened 2026-08-27
Fix (generative/quant): test and fix preset quantizers

pablomlago:
## Reason for this PR

Several activation/weight quantizers reachable through `WEIGHT_QUANT_MAP` / `INPUT_QUANT_MAP` were silently broken and never exercised by tests. Adding a systematic sweep over every quantizer in both maps surfaced four latent defects: duplicate `proxy_class` assignments, missing MSE zero-point mixins on asymmetric group quantizers, a broken `no_scale` input path, and activation MSE-asymmetric quantizers failing to resolve `dtype`/`device`.

This PR also reorganizes the LLM test suite: the growing set of `test_llm*.py` files is grouped under a dedicated `tests/brevitas_examples/llm/` package with simplified names (the redundant `llm` prefix is dropped now that the folder conveys it).

## Changes Made in this PR

**Bug fixes (core + examples):**
- **Duplicate `proxy_class`** in `Fp8e4m3DynamicActPerGroupFloat` and `Fp8e4m3OCPDynamicActPerGroupFloat` (`quantizers.py`): removed the second assignment that silently overrode the intended `Groupwise*` proxy with a `Dynamic*` proxy.
- **Missing `MSEWeightZeroPoint`** on asymmetric per-group MSE weight quantizers: added it to `ShiftedUint8WeightGroupQuantFloatMSE` (`quantizers.py`) and `ShiftedMXUInt8WeightMSE` (`mx_quant_ocp.py`), matching the working per-channel sibling. Fixes `IndexError` from the non-MSE zero-point's `stats_reduce_dim` mismatch on grouped shapes.
- **`dtype`/`device` resolution for activation quantizers** (`quant/base.py`, `quant/solver/act.py`): fixed a `type`->`dtype` typo in `MSESymmetricScaleSubInjector`, restored `dtype` in `MSEAsymmetricScaleSubInjector`, and added `SolveActDtypeDevice` (`dtype = None`, `device = None`) to `ActQuantSolver`. Activations have no `tracked_parameter_list` (unlike weights), so they default to `None`; any explicitly provided value takes precedence. Fixes `DependencyError: can not resolve attribute 'dtype'/'device'` for asymmetric MSE activation quantizers.
- **`no_scale` input path** in `generate_quantizers` (`common/generative/quantize.py`): initialize attention quantizers to `None` and apply `bit_width`/`quantize_zero_point`, and skip the scale-specific block for `no_scale`. Fixes `UnboundLocalError: k_transposed_quant`.

**Tests:**
- New `tests/brevitas_examples/test_generative_quant.py`: parametrized sweep over every `WEIGHT_QUANT_MAP` and `INPUT_QUANT_MAP` leaf, using the same flow as the generative entrypoints (`generate_quantizers -> generate_quant_maps -> layerwise_quantize`) plus a forward pass. The unit under test is the common generative quantizer set, so it is applied to a plain `nn.Linear` layer (no HF model / `transformers` dependency), keeping the test fast and self-contained. Each combination runs under both `weight_scaling_impl_type` values (`parameter_from_stats`, `stats`) - 128 cases total. The previously-failing combinations now serve as regression coverage.

**Test reorganization:**
- Grouped all LLM tests under a new `tests/brevitas_examples/llm/` package and dropped the redundant `llm` prefix from the file names:
  - `test_llm_benchmark.py` -> `llm/test_benchmark.py`
  - `test_llm_cases.py` -> `llm/test_cases.py`
  - `test_llm_data.py` -> `llm/test_data.py`
  - `test_llm_gguf_quants.py` -> `llm/test_gguf_quants.py`
  - `test_llm.py` -> `llm/test_llm.py` (kept as-is; a bare `test_.py` is not a valid name)
  - support files: `llm_test_plugin.py` -> `llm/test_plugin.py`, `llm_test_template.yml` -> `llm/test_template.yml`, `benchmark_test_{grid,rand}.yaml` -> `llm/benchmark_test_{grid,rand}.yaml`
- Updated cross-imports (`tests.brevitas_examples.llm.test_cases`), in-string plugin/template paths, and the benchmark test's `_SRC_BENCHMARK_DIR` relative path for the deeper nesting.
- Consolidated the `llm` marker: files where every test is LLM-only now use a module-level `pytestmark = pytest.mark.llm` (`test_benchmark.py`, `test_data.py`, `test_gguf_quants.py`) instead of per-test decorators. `test_llm.py` keeps per-test markers because it also holds `onnx_export`/`few_shot` tests that must not be marked `llm`.
- Simplified `noxfile.py`: the `tests_brevitas_examples_llm` session now runs the whole folder in one call (`pytest -m llm tests/brevitas_examples/llm/`) instead of listing each file. GitHub Actions workflows are unchanged: they invoke nox by session name (`tests_brevitas_examples_llm*`), which is unchanged, and the CPU session's `--ignore-glob 'tests/brevitas_examples/*llm*'` still excludes the whole `llm/` folder while picking up the new `test_generative_quant.py`.
- Shared, non-LLM files (`common.py`, `conftest.py`, `test_benchmark_utils.py`) remain in `tests/brevitas_examples/`.

## Testing Summary

- New parametrized test module added (`tests/brevitas_examples/test_generative_quant.py`); the previously-failing combinations now serve as regression coverage.
- Run: `pytest tests/brevitas_examples/test_generative_quant.py -q` (also covered by the CPU examples nox session), or the full LLM suite via `nox -s tests_brevitas_examples_llm`.

## Risk Highlight

- [x] There are coverage gaps not covered by tests. - attention (`q/k/v/output`) quantization under `no_scale` is set to `None` (unquantized) rather than exercised.

Note: `SolveActDtypeDevice` adds resolvable `dtype`/`device` (defaulting to `None`) to all activation quantizers; behavior is unchanged for any quantizer that doesn't request them, and provided values still take precedence.

Note: the LLM test files are moved (and some also edited), so a few may render as delete+add rather than rename in the diff depending on similarity thresholds.

## Checklist

- [ ] Initial CI/CD passing.


noxfile.py
```diff
@@ -150,12 +150,14 @@ def tests_brevitas_examples_llm(session, pytorch, jit_status):
     cmd += install_pytorch_cmd(pytorch)
     cmd += install_torchvision_cmd(pytorch)  # Optimum seems to require torchvision
     session.install('-e', '.[test, llm, export]', *cmd)
-    session.run('pytest', '-n', 'logical', '-m', 'llm', 'tests/brevitas_examples/test_llm.py')
-    session.run('pytest', '-n', 'logical', 'tests/brevitas_examples/test_llm_data.py')
+    session.run('pytest', '-n', 'logical', '-m', 'llm', 'tests/brevitas_examples/llm/test_llm.py')
```
Giuseppe5:
```suggestion
    session.run('pytest', '-n', 'logical', '-m', 'llm', 'tests/brevitas_examples/llm/')
```

Would this work? So we don't have to call each test one by one.
If it doesn't, it might be worth spending 2 min to fix the markers.

tests/brevitas_examples/llm/test_quants.py
```diff
@@ -0,0 +1,254 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Apply every quantizer in WEIGHT_QUANT_MAP / INPUT_QUANT_MAP and run a forward pass.
```
Giuseppe5: Here and in the other docstrings, can we have a new line after `"""`?

src/brevitas/quant/solver/act.py
```diff
@@ -174,6 +174,13 @@ def input_view_impl(scaling_per_output):
             return Identity
 
 
+class SolveActDtypeDevice(ExtendedInjector):
```
Giuseppe5, line 177: Was the absence of this class causing any issues?

tests/brevitas_examples/llm/test_quants.py
```diff
@@ -0,0 +1,254 @@
...
+WEIGHT_COMBINATIONS: List[Combination] = _weight_combinations()
+INPUT_COMBINATIONS: List[Combination] = _input_combinations()
+
+# Weight scaling impl types used by the LLM entrypoint (see brevitas_examples/llm/main.py).
+# Exercise both for the weight quantizer.
+WEIGHT_SCALING_IMPL_TYPES: List[str] = ['parameter_from_stats', 'stats']
+
+
+@pytest.fixture(scope="module", params=MODEL_IDS)
+def model(request: pytest.FixtureRequest) -> Tuple[Module, Tensor]:
```
Giuseppe5: If this is test_quants, do we need to instantiate a HF model to test them? Can't we simply have a linear layer?

tests/brevitas_examples/llm/test_quants.py
```diff
@@ -0,0 +1,254 @@
...
+    model.eval()
+    with torch.no_grad():
+        out = model(input_ids)
+    logits = out.logits if hasattr(out, "logits") else out
+    assert torch.isfinite(logits).all(), "Non-finite values in output logits"
+    return logits
+
+
+@pytest.mark.llm
+@requires_pt_ge('2.4')
```
Giuseppe5: Here and below, why do we need this restriction?

src/brevitas/quant/mx_quant_ocp.py
```diff
@@ -149,7 +150,7 @@ class MXInt8WeightMSE(MSESymmetricScale, MXInt8Weight):
     pass
 
 
-class ShiftedMXUInt8WeightMSE(MSEAsymmetricScale, ShiftedMXUInt8Weight):
+class ShiftedMXUInt8WeightMSE(MSEAsymmetricScale, MSEWeightZeroPoint, ShiftedMXUInt8Weight):
     """
     MX Int signed weight quantizer with per-channel MSE-based scaling.
```
nickfraser: Update the docstring to match the quantizer behaviour, .e.g., asymmetric int with MSE scale / zero_point

tests/brevitas_examples/llm/test_quants.py
```diff
@@ -0,0 +1,254 @@
...
+
+# Weight scaling impl types used by the LLM entrypoint (see brevitas_examples/llm/main.py).
+# Exercise both for the weight quantizer.
+WEIGHT_SCALING_IMPL_TYPES: List[str] = ['parameter_from_stats', 'stats']
+
+
+@pytest.fixture(scope="module", params=MODEL_IDS)
+def model(request: pytest.FixtureRequest) -> Tuple[Module, Tensor]:
+    """Load a model and a small input batch once per model id."""
+    transformers = pytest.importorskip("transformers")
```
nickfraser: Why skip if transformers cannot be imported? Don't we want this to fail _loudly_? I'm sure other tests would fail loudly too, so we'll still notice, but I'm not sure why we'd allow a silent failure (skip) here...

nickfraser (approved):
A few comments / questions, otherwise, LGTM!

noxfile.py
pablomlago: Good point, I think it should work.

src/brevitas/quant/solver/act.py
pablomlago: Yes, the failure was the following:
```
FAILED tests/brevitas_examples/test_llm_quants.py::test_input_quant_map[int-static-float_scale-mse-per_tensor-asym] - _dependencies.exceptions.DependencyError: 'ShiftedUint8ActPerTensorFloatMSE' can not resolve attribute 'dtype'
````
This was due to the MSE subinjector requesting `dtype = (this << 1).dtype`.

tests/brevitas_examples/llm/test_quants.py
pablomlago: I refactored the test and removed the restriction.

tests/brevitas_examples/llm/test_quants.py
pablomlago: Agreed. Following @Giuseppe5 comments, I ended refactoring the test, and it no longer needs to import `transformers` anyway.

PR #1598, opened 2026-08-27
Feat (eval): implementing EAR and KL-div evaluation metrics

i-colbert:
## Reason for this PR

Extend LLM evaluation with distribution-based metrics that quantify the effect of quantization beyond perplexity.

Perplexity (PPL) measures the quality of the quantized model on the observed target tokens. EAR and KLD provide additional information about how closely the quantized model preserves the float model's output distribution.

## Changes Made in this PR

- Add Expected Acceptance Rate (EAR) to quantized LLM evaluation*.
- Add KL divergence (KLD) to quantized LLM evaluation*.
- Compute both metrics over the float model’s top-K tokens, either normalized or unnormalized.
- Cache the float model’s top-K token IDs and probabilities for reuse during quantized evaluation.
- Keep `--eval` as the single switch for PPL, EAR, and KLD evaluation.
- Add EAR and KLD to the returned evaluation results and benchmark output.
- Separate float and quantized evaluation paths while sharing logits and evaluation-window handling.

*Both EAR and KLD are normalized by the reference top-K probability mass by default.

## Testing Summary

The following tests were added or updated:
- `tests/brevitas_examples/llm/test_eval.py`
  - Tests EAR and KLD calculations.
  - Tests normalized and unnormalized metric behavior.
  - Tests reference-cache validation and one-forward-per-chunk evaluation.
- `tests/brevitas_examples/llm/test_llm.py`
  - Adds small-model integration coverage for EAR and KLD.
- `tests/brevitas_examples/llm/test_benchmark.py`
  - Tests metric parsing and benchmark result propagation.

Local evaluation was run with the following config:

```yaml
bos_preprocessing: sequence
dtype: bfloat16
eval: true
model: meta-llama/Llama-3.2-1B-Instruct
weight_bit_width: 4
weight_quant_granularity: per_channel
weight_quant_type: asym
weight_scale_precision: float_scale
```

Environment versions used for the evaluation:

- Python: 3.12.13
- PyTorch: 2.6.0+rocm6.1
- Transformers: 5.16.1

### Baseline

Results from master at `6df6a69`:

```text
Float perplexity (wikitext2): 12.154
Quantized perplexity (wikitext2): 21.806
```

### This PR

```text
Float perplexity (wikitext2): 12.154
Quantized perplexity (wikitext2): 21.806
Quantized expected acceptance rate (wikitext2): 0.635022
Quantized KL divergence (wikitext2): 0.687447
```

Using the above configuration with `--weight-bit-width=8` we get:

```text
Float perplexity (wikitext2): 12.154
Quantized perplexity (wikitext2): 12.161
Quantized expected acceptance rate (wikitext2): 0.978353
Quantized KL divergence (wikitext2): 0.002100
```


src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +50,234 @@ def create_validation_dataloader(data, seqlen, device):
...
+        shift_logits = lm_logits[:, context_length - 1:-1]
+        shift_logits = shift_logits.to(dtype)
 
+        if compute_perplexity:
+            loss = cross_entropy_loss(
+                shift_logits.reshape(-1, shift_logits.shape[-1]), reference_labels.reshape(-1))
             nlls.append(loss)
 
-    ppl = torch.exp(torch.stack(nlls).mean())
+        if build_reference_probabilities:
```
Giuseppe5: I would probably split this function in two, one for float one for quant, and find a way to share everything that's shared and keep the rest separate, so we simplify the interface (quant version doesn't need `build_reference_probabilities`).

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +50,234 @@ def create_validation_dataloader(data, seqlen, device):
...
+
+    return EvaluationMetrics(
+        perplexity=perplexity,
+        expected_acceptance_rate=expected_acceptance_rate,
+        kld=kld,
+        reference_probabilities=reference_cache)
+
+
+@torch.no_grad()
+def compute_perplexity(
```
Giuseppe5: I think we have this the wrong way around.
I would rather have `compute_perplexity` function that takes as input the quant_logits and spits out ppl, and then `compute_evaluation_metrics` that looks something like:

```python
def compute_float_evaluation_metrics(....)
     logits = get_logits(...)
     ppl = compute_ppl(quant_logits,...)
```

Then the function `get_logits` should also be shared between the `float` and `quant` version of `compute_evaluation_metrics`.

Currently maybe it does not make a lot of difference, but if we want to add a new metrics, it would make it easier since the metric function would also concerns itself with the metric itself, and all the other required inputs (quant logits, float logits) are provided by other functions.

In the current interface, we would have always to add new stuff to `compute_evaluation_metrics`.

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +55,192 @@ def create_validation_dataloader(data, seqlen, device):
...
+        top_logits, top_ids = scored_logits.topk(top_k, dim=-1)
+        top_probabilities = torch.exp(top_logits - scored_logits.logsumexp(dim=-1, keepdim=True))
+        reference_chunks.append(
+            TopKReferenceChunk(
+                token_ids=top_ids.to(device="cpu", dtype=torch.int32),
+                probabilities=top_probabilities.to(device="cpu", dtype=torch.float32)))
+        num_positions += top_ids.numel() // top_k
+
+    return EvaluationMetrics(
+        ppl=torch.exp(torch.stack(nlls).mean()).item(),
```
Giuseppe5: Shouldn't this function call _compute_perplexity etc.?

i-colbert: If we change `_compute_perplexity`, then yes. I have implemented that now.

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
+        quantized_log_q = torch.log_softmax(
+            logits.to(dtype=torch.float32), dim=-1).gather(-1, token_ids)
+        quantized_q = quantized_log_q.exp()
+        # Use full-softmax probabilities on the reference model's top-K support.
+        # Normalize EAR and KLD by the reference top-K mass when requested.
+        if normalize:
+            reference_mass = reference_p.sum(dim=-1, keepdim=True)
+            reference_p = reference_p / reference_mass
+            quantized_q = quantized_q / reference_mass
+        ear.update(quantized_q, reference_p)
```
Giuseppe5, line 194: I would probably still move this as part of ear/kld update. Even if we have a bit of duplication, I think it's fine as long as it's not crazy expensive (which it shouldn't be)

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
-    model = model.eval()
 
-    cross_entropy_loss = nn.CrossEntropyLoss()
+def _move_subsample_to_model(model: torch.nn.Module, subsample: Dict[str, Any]) -> Dict[str, Any]:
+    use_accelerate = hasattr(model, "hf_device_map")
+    if not use_accelerate or not hasattr(model, "_hf_hook"):
+        device = next(model.parameters()).device
+    else:
+        device = model._hf_hook.execution_device
+    for name, value in subsample.items():
```
pablomlago, line 100: I would maybe expand `recursive_to_device` to also accept `Mapping`s, and have this `for` there, as I think it could be useful in other places. Then, this function would just call `recursive_to_device` (after the `accelerate` checks).

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
+
+def _set_eval_seed(seed: int) -> None:
     random.seed(seed)
     np.random.seed(seed)
     torch.random.manual_seed(seed)
 
-    model = model.eval()
 
-    cross_entropy_loss = nn.CrossEntropyLoss()
+def _move_subsample_to_model(model: torch.nn.Module, subsample: Dict[str, Any]) -> Dict[str, Any]:
```
pablomlago, line 94: Maybe this function should be `_move_input_to_model_device` (or some other explicit name) and its argument `subsample` be renamed to `sample`, as the fact that it is a slice of an input is not relevant for this function.

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
+    if not use_accelerate or not hasattr(model, "_hf_hook"):
+        device = next(model.parameters()).device
+    else:
+        device = model._hf_hook.execution_device
+    for name, value in subsample.items():
+        subsample[name] = recursive_to_device(value, device)
+    return subsample
+
+
+def _get_logits(
```
pablomlago, line 105: This is actually returning the shifted logits (`shift_logits` in the original code) so this should be reflected in the docstring/name of the function.

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
+                subsample["input_ids"][:, context_length:]
+
+
+class MetricBase(ABC):
+
+    def __init__(self, dtype: torch.dtype = torch.float32):
+        self.dtype = dtype
+
+    @abstractmethod
+    def update(self, output: torch.Tensor, target: torch.Tensor) -> None:
```
pablomlago: I would maybe rename these methods, e.g. `update` to `accumulate` and `finalize` to `aggregate`, or something along those lines.

src/brevitas_examples/llm/llm_quant/eval.py
```diff
@@ -47,60 +57,237 @@ def create_validation_dataloader(data, seqlen, device):
...
+        target = target.to(dtype=self.dtype)
+        self.kld_sum += (target * (target.log() - output.log())).sum().item()
+        self.num_positions += target.numel() // target.shape[-1]
+
+    def finalize(self) -> float:
+        return self.kld_sum / self.num_positions
+
+
+@torch.no_grad()
+def compute_float_evaluation_metrics(
```
pablomlago, line 134: Should the method name reflect that it is doing some caching apart on top of computing metrics?

src/brevitas_examples/llm/benchmark/llm_benchmark.py
```diff
@@ -31,6 +31,11 @@ def parse_log(job_log: str) -> Dict[str, Any]:
         # Find the line containing Quant PPL number
```
pablomlago, line 31: Maybe it would be worth extracting the common functionality, e.g.:
```
metric_logs = {"quant_ppl": r"Quantized perplexity \((.*?)\): (\d+\.\d+)", ...}
metrics = {metric_key: re.search(metric_value) for metric_key, metric_value in metric_logs.items()}
```

pablomlago (approved):
Minor suggestions:

PR #1606, opened 2026-09-01
Fix (brevitas_examples/llm): remove deprecated flag

Giuseppe5:
## Reason for this PR

The flag `save_safetensor` has been removed in transformers 5.x, and now by default all weights are saved in safetensor format.

Although we have a test with a yaml that exposed that flag, the yaml was never validated, so this error was never triggered.

Even though until now the yamls specified `save_safetensor=False`, this change should have no impact since we were never storing checkpoints during training anyway.


tests/brevitas_examples/llm/test_llm.py
```diff
@@ -566,8 +559,9 @@ def test_parse_yaml_trainer_arguments(caplog, kwargs):
 
     def quantize_llm_assert_args(args, extra_args=None):
         for key in extra_args_keys:
-            assert key not in args, f"Key {key} should not be known by the parser"
+            assert key not in vars(args), f"Key {key} should not be known by the parser"
         assert extra_args == expected_extra_args, f"Expected extra arguments {expected_extra_args} but got {extra_args}"
+        parse_rotation_optimization_args(extra_args=extra_args, trainer_cls=GeneralizedTrainer)
```
nickfraser, line 564: What is the reason for this change?

docsrc/source/user_guide/rotations.rst
```diff
@@ -182,4 +181,4 @@ Note that the training parameters used in the SpinQuant paper [3]_ can be found
 .. [2] Ashkboos, S., Mohtashami, A., Croci, M., Li, B., Cameron, P., Jaggi, M., ... & Hensman, J. (2025). Quarot: Outlier-free 4-bit inference in rotated llms. Advances in Neural Information Processing Systems, 37, 100213-100240.
 .. [3] Liu, Z., Zhao, C., Fedorov, I., Soran, B., Choudhary, D., Krishnamoorthi, R., ... & Blankevoort, T. (2024). Spinquant: Llm quantization with learned rotations. arXiv preprint arXiv:2405.16406.
 .. [4] Li, J., Fuxin, L., & Todorovic, S. (2020). Efficient riemannian optimization on the stiefel manifold via the cayley transform. arXiv preprint arXiv:2002.01113.
-.. [5] https://docs.pytorch.org/tutorials/intermediate/parametrizations.html
\ No newline at end of file
+.. [5] https://docs.pytorch.org/tutorials/intermediate/parametrizations.html
```
nickfraser: Same URL? Revert this for less churn?

nickfraser (approved):
A few comments, otherwise, LGTM!

tests/brevitas_examples/llm/test_llm.py
Giuseppe5: We were manually checking which extra keys we expected to find in the yaml, whether or not they were supported by transformers' Trainer.
This is done with:
```python
assert extra_args == expected_extra_args, f"Expected extra arguments {expected_extra_args} but got {extra_args}"
```

This line matches the yaml args against the ones supported in transformers.

PR #1613, opened 2026-09-09
Feat (fx): always import backport fx

Giuseppe5:
## Reason for this PR

We need to use our custom FX for value_tracer and other imports.

This remove the torch_version check, although potentially creating incompatibility with future versions.

Originally designed while waiting for https://github.com/pytorch/pytorch/pull/94461 to get merged, which will never happen.


nickfraser (approved):
LGTM!

PR #1619, opened 2026-09-18
Fix (export/inference): guard torch._dynamo usage on old torch versions

nickfraser:
# Guard `torch._dynamo` usage in `quant_inference_mode`

Failing CI run:
https://github.com/Xilinx/brevitas/actions/runs/34640106969
(for example job `build (3.10, 2.0.1, ubuntu-latest, jit_disabled)`:
https://github.com/Xilinx/brevitas/actions/runs/34640106969/job/103397447034)

## Summary

`quant_inference_mode.__enter__` called `torch._dynamo.reset()` while
`src/brevitas/export/inference/manager.py` only imported the top-level `torch`
package, so the call relied on some other module having attached the
`torch._dynamo` submodule earlier in the same process. On torch 1.13.1 and
2.0.1 that side effect does not reliably occur and entering the context manager
raised `AttributeError: module 'torch' has no attribute '_dynamo'`. This PR
imports the submodule explicitly and skips the reset on torch versions that do
not ship it.

## Changes

`src/brevitas/export/inference/manager.py` now performs an explicit
`import torch._dynamo` at module scope, wrapped in `try`/`except ImportError`
that records availability in a module-level `_DYNAMO_AVAILABLE` flag. The
`torch._dynamo.reset()` call in `quant_inference_mode.__enter__` is guarded by
that flag.

The guard is required rather than a bare import: `torch._dynamo` does not exist
at all before torch 2.0, so an unconditional top-level import would turn a
runtime `AttributeError` into an import-time `ModuleNotFoundError` for the whole
`brevitas.export.inference` module on torch 1.13.x. Torch 2.0.x has the
submodule but only imports it lazily from `torch.compile`; torch 2.1+ attaches
it eagerly in `torch/__init__.py`, which is why only the two older versions
failed. The approach mirrors existing defensive patterns in
`src/brevitas/__init__.py` and `src/brevitas/compiler.py`.

The second `torch._dynamo` reference in the same file
(`torch._dynamo.config.force_parameter_static_shapes`) is left unguarded: it is
only reachable with `compile=True`, which already requires `torch.compile` and
therefore torch >= 2.0, where the new explicit import guarantees the attribute
is present.

## Verification

- `python -m py_compile` on the modified file succeeds.
- `pre-commit run --files src/brevitas/export/inference/manager.py`: isort and
  the other hooks pass. The local yapf hook aborts with
  `ModuleNotFoundError: lib2to3`, a pre-existing environment issue on Python
  3.13; running `yapf==0.32.0 --diff` under Python 3.11 against the repository
  config reports no changes.
- Scoped re-grep of `src/brevitas/` for `torch._dynamo` confirms no other
  unguarded usage: remaining occurrences are version-guarded, already wrapped in
  `try`/`except`, function-local explicit imports under `backport/fx/`, or
  comments and string literals.
- The affected tests were not executed. No torch build is installed in the
  development environment used here, and the reporting torch versions (1.13.1,
  2.0.1) have no wheels for its Python 3.14 interpreter, so the behavior
  difference across torch versions was confirmed from upstream PyTorch sources
  rather than by running the suite.


Giuseppe5 (approved):
LGTM

PR #1620, opened 2026-09-18
Fix (tests): skip groupwise dequant compile test on torch 2.4.1 Windows

nickfraser:
# Skip groupwise dequant compile test on torch 2.4.1 / Windows

## Summary

`test_compile_groupwise_dequant_expand` fails on Windows with torch 2.4.1 only,
because dynamo refuses to inline its own `list_cmp` polyfill
(`torch._dynamo.exc.Unsupported: 'inline in skipfiles: list_cmp ... skipped
according trace_rules.lookup SKIP_DIRS'`) when tracing the `torch.Size` vs
tuple comparisons in the expand path of `groupwise_dequant_expand`. The defect
is upstream, so this PR skips the test for that one torch release on Windows
instead of changing Brevitas code. Failure evidence comes from the periodic
Pytest run
[34640106969](https://github.com/Xilinx/brevitas/actions/runs/34640106969).

## Changes

`tests/brevitas/utils/test_compile.py` guards the body of
`test_compile_groupwise_dequant_expand` with
`if torch_version == version.parse('2.4.1') and platform.system() == 'Windows':
pytest.skip(...)`, following the in-body skip pattern already used for
version- and platform-specific `torch.compile` quirks in
`tests/brevitas/proxy/test_compile.py` and
`tests/brevitas_end_to_end/test_torchvision_models.py`. The existing
`requires_pt_ge('2.2')` and `requires_torch_compile()` decorators are
unchanged.

The version match is exact rather than a range, since neighbouring releases
(2.3.1, 2.5.1, 2.7.1, ...) trace this function fine. The Windows condition is
equally deliberate: all three torch 2.4.1 Windows cells (Python 3.10/3.11/3.12)
failed in that run while all six Linux and macOS 2.4.1 cells passed, so
coverage is retained everywhere else. No production code, no other test, and no
CI configuration is touched.

## Verification

Not run against torch 2.4.1 locally; torch is not installed in this
environment. Validation performed outside the repository:

- A pytest run over a stand-in replica of the guard (with `torch_version` and
  `platform.system()` injected) across `{2.3.1, 2.4.0, 2.4.1, 2.5.1, 2.7.1}` x
  `{Windows, Linux, Darwin}` and all eight test parametrizations: the affected
  cells are reported as skipped with the intended reason, every other cell
  still runs.
- `yapf` 0.32.0 and `isort` 5.11.5, the versions pinned by the repository's
  pre-commit hooks, report no changes for the edited file.

Known limitation: `brevitas.torch_version` strips only a `+cpu` local version
tag, so an install such as `2.4.1+cu121` would not match the guard. CI installs
`torch==2.4.1+cpu` on Windows, so the skip applies there today; if that ever
changes, the test simply stops being skipped rather than silently passing.


tests/brevitas/utils/test_compile.py
```diff
@@ -40,6 +41,11 @@ def test_compile_padding_to_multiple(shape, dim, multiple, expected_shape):
 @pytest.mark.parametrize('expand_metadata', [False, True])
 @requires_pt_ge('2.2')
 @requires_torch_compile()
+@excludes_pt_version(
```
Giuseppe5: We do pytest.skip inside the test body usually.

I don't mind this decorator, but then we might want to extend it to all the other use-cases with pytest.skip?

nickfraser: Moved `pytest.skip` to test body in [813f8b9](https://github.com/Xilinx/brevitas/pull/1620/commits/813f8b9007066ae39924b0d3f2b02d7de7f7ed50).

PR #1623, opened 2026-09-22
Fix (examples/llm): update qronos configs for lighteval 0.13 task names

averosim:
requirements/requirements-lighteval.txt pins lighteval[math]==0.13, but the qronos example configs still use the suite|task|num_fewshot|truncate task syntax from lighteval 0.9.2, which was removed in 0.10. Every config under src/brevitas_examples/papers/qronos/ therefore fails with:

ValueError: Cannot find task lighteval|piqa|0|0 in task list or in custom task registry

This updates all 7 configs to the 0.13 naming.

| Old | New |
|---|---|
| lighteval\|piqa\|0\|0 | piqa_lm_eval |
| lighteval\|arc:easy\|0\|0 | arc:easy |
| leaderboard\|arc:challenge\|0\|0 | arc:challenge |
| leaderboard\|winogrande\|0\|0 | winogrande |
| leaderboard\|hellaswag\|0\|0 | hellaswag_lm_eval |

PIQA and HellaSwag map to the custom tasks already registered in eval_lighteval.py (TASKS_TABLE), not the native lighteval ones. This is deliberate: as the comments there note, recent lighteval changed those task definitions (native piqa now evaluates validation + test rather than validation only), so using the native names would silently change results.

Note: this alone is not sufficient to run the configs end-to-end — PIQA additionally fails under datasets>=4.0. See the follow-up PR: " Fix (examples/llm): load PIQA from parquet revision for datasets>=4.0"


Giuseppe5:
Does this impact only Qronos? What about the other papers?

averosim:
> Does this impact only Qronos? What about the other papers?

I just checked and it impacts all the papers .yaml configs that have few_shot_tasks. Do you want me to prepare another PR to fix all of them in the same way as qronos?

Giuseppe5:
We should do everything in this PR, just change the title

averosim:
Other paper configs fixed. Title and description updated.

